PR #14, opened 2024-12-11
chore(ci): refactor install mdbook code

fgimenez:
Also some other nits, like aligning `on` sections format of all workflows and `checkout` action versions


syjn99 (approved):
Thanks!

PR #16, opened 2024-12-12
feat: Init Network Types

varun-doshi:
This PR marks the beginning of implementing the Network types according to the official CL specs.
Uses `alloy-primitives` and  will add `ghiliweld-sszb` as and when required.


Cargo.toml
```diff
@@ -24,3 +24,4 @@ version = "0.1.0"
 
 [workspace.dependencies]
 clap = "4"
+alloy-primitives = "0.8.15"
```
KolbyML, line 27: please add a new line to remove this 
image

crates/common/Cargo.toml
```diff
@@ -10,3 +10,4 @@ rust-version.workspace = true
 version.workspace = true
 
 [dependencies]
+alloy-primitives = {workspace = true}
```
KolbyML, line 13:
```suggestion
alloy-primitives.workspace = true
```

crates/common/src/phase0/primitives.rs
```diff
@@ -0,0 +1,4 @@
+use alloy_primitives::{U256,U64};
+
+pub type NodeID = U256;         //Node Idnetifier
+pub type SubnetID=U64;          //Subnet Identifier
```
KolbyML, line 4: please add a new line 
image

KolbyML:
It appears lint is failing for the CI, I made a PR so we can do `make lint` going forward https://github.com/ReamLabs/ream/pull/17

crates/common/Cargo.toml
KolbyML, line 13: Add a new line to this file too

crates/common/src/phase0/constants.rs
```diff
@@ -0,0 +1,21 @@
+use alloy_primitives::{aliases::B32,FixedBytes};
```
KolbyML, line 1:
```suggestion
use alloy_primitives::{B32,FixedBytes};
```

you should just be able to import B32 directly, let me know if that works or not

crates/common/src/phase0/constants.rs
syjn99, line 1: In my perspective, I'm not sure all the constants are needed now in this phase. Some constants might not be used, or for prototyping we might not have to inherit all the constants from phase0. There are few changes as forks go on, and I think we can start from the latest on which is Deneb.

KolbyML: I think we should only support deneb to be honest

syjn99: Besides, it would be better to add those constants when they are needed.

varun-doshi: Should I continue building out the primitives then(non-constant types) and add the constants if required?

KolbyML: It is fine to continue building out the the primitive types as long as they are for deneb, 

here is an example PR with tests I made https://github.com/ReamLabs/ream/pull/20 I got the test data from https://github.com/ethereum/consensus-spec-tests

generally we should be adding tests as we add these primatives

varun-doshi:
@KolbyML it might be better to use custom types like `Version` since they may have required impl functions later

KolbyML:
> @KolbyML it might be better to use custom types like `Version` since they may have required impl functions later

I would like to avoid using custom primitives unless we have a good reason to use them, as they abstract the primitives and make the code harder to read.

Instead of looking at a struct and knowing what makes it up, I have to juggle 10-20 custom custom types or wrappers around containers in my head.

if a custom type is just a single primitive like `type Slot = u64`, I don't think we should have a custom type. As we already know it is a slot by it's variable name.

having `slot: Slot` isn't descriptive, `slot: u64` is much more descriptive as you know what it is, and it's actual type.

KolbyML:
When  you want to put your changes on top of master, please use this going forward

```rust
git fetch upstream
git rebase upstream/master
```

varun-doshi:
> > @KolbyML it might be better to use custom types like `Version` since they may have required impl functions later
> 
> I would like to avoid using custom primitives unless we have a good reason to use them, as they abstract the primitives and make the code harder to read.
> 
> Instead of looking at a struct and knowing what makes it up, I have to juggle 10-20 custom custom types or wrappers around containers in my head.
> 
> if a custom type is just a single primitive like `type Slot = u64`, I don't think we should have a custom type. As we already know it is a slot by it's variable name.
> 
> having `slot: Slot` isn't descriptive, `slot: u64` is much more descriptive as you know what it is, and it's actual type.

Understood

KolbyML:
Do you still want this PR to be reviewed or is it fine to close?

varun-doshi:
> Do you still want this PR to be reviewed or is it fine to close?

I think its best to close it now
Will circle back to it if needed later

KolbyML:
Sounds good

PR #19, opened 2024-12-20
feat: add tracing

syjn99:
Add basic structure for `tracing`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "aho-corasick"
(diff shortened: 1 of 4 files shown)


KolbyML:
In rust you can just do RUST_LOG=verbosity_level before the executable. So I am not that big of a fan of this direction as it doesn't seem rusty

syjn99:
@KolbyML Just fixed to use `EnvFilter`. Note that `verbosity` option is just an example for argument.

KolbyML (approved):
:shipit: looks good

PR #20, opened 2025-01-01
feat: add initial Beacon Chain types

KolbyML:
We want to implement the Beacon Chain types for deneb, here is the start

the test data is from https://github.com/ethereum/consensus-spec-tests


Kayden-ML (approved):

Kayden-ML (approved):
read through it all looks good

PR #24, opened 2025-01-01
feat: implement IndexedAttestation Beacon Chain type

KolbyML:
**Review this PR first** https://github.com/ReamLabs/ream/pull/23

implements https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#indexedattestation

### DIFF
--- a/crates/common/consensus/src/indexed_attestation.rs
+++ b/crates/common/consensus/src/indexed_attestation.rs
@@ -1,0 +1,111 @@
+use serde::{Deserialize, Serialize};
+use ssz_derive::{Decode, Encode};
+use ssz_types::{typenum, VariableList};
+use tree_hash_derive::TreeHash;
+
+use crate::{attestation_data::AttestationData, signature::BlsSignature};
+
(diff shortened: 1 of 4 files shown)


crates/common/consensus/src/indexed_attestation.rs
```diff
@@ -0,0 +1,111 @@
+use serde::{Deserialize, Serialize};
+use ssz_derive::{Decode, Encode};
+use ssz_types::{typenum, VariableList};
+use tree_hash_derive::TreeHash;
+
+use crate::{attestation_data::AttestationData, signature::BlsSignature};
+
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
+pub struct IndexedAttestation {
+    pub attesting_indices: VariableList<u64, typenum::U2048>,
```
syjn99, line 10: Sorry for late review: [EIP-7549](https://eips.ethereum.org/EIPS/eip-7549) modifies `Attestation` containers, as well as `IndexedAttestation`. I'm not sure when Pectra will be enabled, but would be better handling for future upgrades.

PR #33, opened 2025-01-08
feat: implemented compute_signing_root function

Kayden-ML:
I implemented the compute_signing_root function


KolbyML (approved):
:shipit: looks good

PR #34, opened 2025-01-11
feat: Implemented more beacon functions

Kayden-ML:
I implemented quite a few beacon implementations such as get_previous_epoch, get_block_root, get_block_root_at_slot, get_randao_mix and get_validator_churn_limit


crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -78,7 +88,58 @@ pub struct BeaconState {
...
+    /// Return the block root at the start of a recent ``epoch``.
+    pub fn get_block_root(&self, epoch: u64) -> B256 {
+        self.get_block_root_at_slot(compute_start_slot_at_epoch(epoch))
+    }
+
+    /// Return the block root at a recent ``slot``.
+    pub fn get_block_root_at_slot(&self, slot: u64) -> B256 {
+        assert!(slot < self.slot && self.slot <= slot + SLOTS_PER_HISTORICAL_ROOT);
+        self.block_roots[(slot % SLOTS_PER_HISTORICAL_ROOT) as usize]
+    }
```
KolbyML, line 115:
```suggestion
    /// Return the block root at the start of a recent ``epoch``.
    pub fn get_block_root(&self, epoch: u64) -> anyhow::Result<B256> {
        self.get_block_root_at_slot(compute_start_slot_at_epoch(epoch))
    }

    /// Return the block root at a recent ``slot``.
    pub fn get_block_root_at_slot(&self, slot: u64) -> anyhow::Result<B256> {
        bail!(slot < self.slot && self.slot <= slot + SLOTS_PER_HISTORICAL_ROOT, "slot given was outside of block_roots range");
        Ok(self.block_roots[(slot % SLOTS_PER_HISTORICAL_ROOT) as usize])
    }
```

use bail instead of an assert, so the program doesn't crash and we can peacefully handle the error

crates/common/consensus/src/fork_choice/helpers/constants.rs
```diff
@@ -12,3 +12,7 @@ pub const REORG_PARENT_WEIGHT_THRESHOLD: u64 = 160;
 pub const SECONDS_PER_SLOT: u64 = 12;
 pub const SLOTS_PER_EPOCH: u64 = 32;
 pub const SHUFFLE_ROUND_COUNT: u8 = 90;
+pub const SLOTS_PER_HISTORICAL_ROOT: u64 = 8192;
+pub const EPOCHS_PER_HISTORICAL_VECTOR: u64 = 65536;
+pub const MIN_PER_EPOCH_CHURN_LIMIT: u64 = 4;
+pub const CHURN_LIMIT_QUOTIENT: u64 = 65536;
```
KolbyML, line 18: please sort these alphabetically

KolbyML (approved):
:shipit: looks good

PR #38, opened 2025-01-13
feat: add bootnode support and basic discv5 support

KolbyML:
fixes https://github.com/ReamLabs/ream/issues/91

- Add support for initialing libp2p with bootnodes
- handle finding peers using discv5 and dialing them once found.

## Followup PR's
In my next PR I am going to be implementing basic Req/Resp for the Req/Resp domain of the consensus p2p-interface spec https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#the-reqresp-domain

There is still work to do like finding subnets, and setting a key-value in the enr to specify the node is an eth2 node


crates/networking/discv5/src/discovery.rs
```diff
@@ -64,16 +74,18 @@ impl Discovery {
             }
 
             let _ = discv5.add_enr(bootnode_enr).map_err(|e| {
-                println!("Discv5 service failed. Error: {:?}", e);
+                info!("Discv5 service failed. Error: {:?}", e);
```
hopinheimer, line 77: error here

hopinheimer (commented):
small nit, but I got peer responses from teku and nimbus, let's get this in we'll work on keeping the connection alive

jihoonsong (approved):
Just some nits. It looks work as expected within its scope. Once we have bootstrapping network done, we probably want to refactor it.

KolbyML:
>  Once we have bootstrapping network done, we probably want to refactor it.

what do you mean refactor it?

jihoonsong:
I see some room for improvements. Could be just trimming or refactor depending on rest of the implementation.

PR #40, opened 2025-01-22
chore:remove patch version from deps

supernovahs:
This PR removes patch version from deps allowing users to specify any patch version instead of being forced to use the same patch version as ream


unnawut (approved):
lgtm thank you! ❤️

PR #42, opened 2025-01-30
feat: add macro for testing operations

syjn99:
Added macro for operations which integrates codebase with EF tests([consensus-specs](https://github.com/ethereum/consensus-specs)).

`operations` are the first building block for snarkifying the current beacon chain STF. `state_transition` looks like following:

```python
def state_transition(state: BeaconState, signed_block: SignedBeaconBlock, validate_result: bool=True) -> None:
    block = signed_block.message
    # Process slots (including those with no blocks) since block
    process_slots(state, block.slot)
    # Verify signature
    if validate_result:
        assert verify_block_signature(state, signed_block)
    # Process block
    process_block(state, block)
    # Verify state root
    if validate_result:
        assert block.state_root == hash_tree_root(state)
```

`process_operations` is a part of `process_block`, and it would be a good starting point for implementing the whole STF. You could also refer to this [link](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/operations) at `consensus-specs`, which defines the spec tests for operations.

_NOTE_: We have to add few more `operations` for post-Electra. All those operations are based on deneb.
_NOTE 2_: For each operation, we can also compile it into RISC-V-compatible program to run on zkVM.
_NOTE 3_: To integrate with implementation, you may refer to this [gist](https://gist.github.com/syjn99/75c04a74639d72cc3e952df684afb858).


unnawut:
Curious how should we go about doing this?

> NOTE 2: For each operation, we can also compile it into RISC-V-compatible program to run on zkVM.

My experience with [consenzero](https://github.com/unnawut/consenzero) that ream's implementation so far is incompatible with the target zkVM so I still need to do some hacking on the ream codebase to get it to work with risczero.

If we can guarantee at least RISC-V compatibility in our builds before splitting the work implementing `process_*` that'd save a lot of fixing time later I think.

But still lgtm! We can do above in another PR.

syjn99:
@unnawut Yes, what I meant was that we could split the entire STF into multiple parts. But as you pointed out, lifting BeaconState onto zkVMs should come first. I’ll also dig deeper into that.

KolbyML (changes requested):
shouldn't we just use anyhow instead of defining our own error type especially for the time being.

I don't get why we would need a skeleton for people to implement these functions, I am also not a fan of merging commented code.

I think it is fine if we merge the testing macro, but I don't see a reason to merge the rest, the rest can be done in follow up PR's

KolbyML:
also the way this `impl BlockState` is done and located within the file structure of the project will cause problems if these functions are used, so we would need to move them regardless, so I don't we should be the impl there, if we will have to move it regardless.

KolbyML (approved):
:shipit: looks good

syjn99:
@KolbyML Thank you for your comment! Based on your feedback, I removed the commented and unimplemented code.

PR #54, opened 2025-01-30
feat: implement process_block_header

### LINKED ISSUE #46: Implement process_block_header()
Requires #42

Implement `consensus::state_transition::operations::process_block_header()` according to its latest specification (Phase0).

https://ethereum.github.io/consensus-specs/specs/phase0/beacon-chain/#block-header

```py
def process_block_header(state: BeaconState, block: BeaconBlock) -> None:
    # Verify that the slots match
    assert block.slot == state.slot
    # Verify that the block is newer than latest block header
    assert block.slot > state.latest_block_header.slot
    # Verify that proposer index is the correct index
    assert block.proposer_index == get_beacon_proposer_index(state)
    # Verify that the parent matches
    assert block.parent_root == hash_tree_root(state.latest_block_header)
    # Cache current block as the new latest block
    state.latest_block_header = BeaconBlockHeader(
        slot=block.slot,
        proposer_index=block.proposer_index,
        parent_root=block.parent_root,
        state_root=Bytes32(),  # Overwritten in the next process_slot call
        body_root=hash_tree_root(block.body),
    )

    # Verify proposer is not slashed
    proposer = state.validators[block.proposer_index]
    assert not proposer.slashed
```

varun-doshi:
Closes #46

### DIFF
--- a/crates/common/consensus/src/deneb/beacon_state.rs
+++ b/crates/common/consensus/src/deneb/beacon_state.rs
@@ -14,9 +14,10 @@ use ssz_types::{
     typenum::{U1099511627776, U16777216, U2048, U4, U65536, U8192},
     BitVector, FixedVector, VariableList,
 };
+use tree_hash::TreeHash;
 use tree_hash_derive::TreeHash;
 
-use super::execution_payload_header::ExecutionPayloadHeader;
+use super::{beacon_block::BeaconBlock, execution_payload_header::ExecutionPayloadHeader};
 use crate::{
     attestation::Attestation,
     attestation_data::AttestationData,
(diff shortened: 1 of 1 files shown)


unnawut (approved):
perfect!

unnawut:
btw any chance you could have at least 1 unit test covering this function?

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -581,4 +582,37 @@ impl BeaconState {
         }
         Ok((rewards, penalties))
     }
+
+    pub fn process_block_header(mut self, block: BeaconBlock) -> anyhow::Result<()> {
```
KolbyML, line 586:
```suggestion
    pub fn process_block_header(&mut self, block: BeaconBlock) -> anyhow::Result<()> {
```

we shouldn't be consuming self

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -581,4 +582,37 @@ impl BeaconState {
...
+        ensure!(
+            self.slot == block.slot,
+            "State slot must be equal to block slot"
+        );
+        ensure!(
+            block.slot > self.latest_block_header.slot,
+            "Block slot must be greater than latest block header slot of state"
+        );
+        ensure!(
+            block.proposer_index == self.get_beacon_proposer_index().unwrap(),
```
KolbyML, line 596:
```suggestion
            block.proposer_index == self.get_beacon_proposer_index()?,
```

no using unwrap

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -581,4 +582,37 @@ impl BeaconState {
...
+            "Block proposer index must be equal to beacon proposer index"
+        );
+        ensure!(
+            block.parent_root == self.latest_block_header.tree_hash_root(),
+            "Block Parent Root must be equal root of latest block header"
+        );
+
+        self.latest_block_header = BeaconBlockHeader {
```
KolbyML, line 604:
```suggestion
        // Cache current block as the new latest block
        self.latest_block_header = BeaconBlockHeader {
```

It would be nice to carry over the comments from the pseudo code

KolbyML (changes requested):

Issue #45, opened 2025-01-30
Implement process_attester_slashing()


unnawut:
Requires #42

Implement `consensus::state_transition::operations::process_attester_slashing()` according to its latest specification (Phase0).

https://ethereum.github.io/consensus-specs/specs/phase0/beacon-chain/#attester-slashings

```py
def process_attester_slashing(state: BeaconState, attester_slashing: AttesterSlashing) -> None:
    attestation_1 = attester_slashing.attestation_1
    attestation_2 = attester_slashing.attestation_2
    assert is_slashable_attestation_data(attestation_1.data, attestation_2.data)
    assert is_valid_indexed_attestation(state, attestation_1)
    assert is_valid_indexed_attestation(state, attestation_2)

    slashed_any = False
    indices = set(attestation_1.attesting_indices).intersection(attestation_2.attesting_indices)
    for index in sorted(indices):
        if is_slashable_validator(state.validators[index], get_current_epoch(state)):
            slash_validator(state, index)
            slashed_any = True
    assert slashed_any
```

varun-doshi:
Completed as per #57

PR #57, opened 2025-01-31
feat: implement process_attester_slashing

varun-doshi:
Ref #45

### DIFF
--- a/crates/common/consensus/src/deneb/beacon_state.rs
+++ b/crates/common/consensus/src/deneb/beacon_state.rs
@@ -26,6 +26,7 @@ use super::{
 use crate::{
     attestation::Attestation,
     attestation_data::AttestationData,
+    attester_slashing::AttesterSlashing,
     beacon_block_header::BeaconBlockHeader,
     bls_to_execution_change::SignedBLSToExecutionChange,
     checkpoint::Checkpoint,
@@ -56,7 +57,9 @@ use crate::{
     misc::{
         compute_activation_exit_epoch, compute_committee, compute_domain, compute_epoch_at_slot,
         compute_shuffled_index, compute_signing_root, compute_start_slot_at_epoch,
+        is_sorted_and_unique,
     },
+    predicates::is_slashable_attestation_data,
     proposer_slashing::ProposerSlashing,
     pubkey::PubKey,
     signature::BlsSignature,
@@ -284,6 +287,7 @@ impl BeaconState {
             Some(fork_version),
             Some(self.genesis_validators_root),
         )
+        .map_err(|err| anyhow!("Domain computation failed: {:?}", err))
     }
 
     /// Return the beacon committee at ``slot`` for ``index``.
@@ -296,6 +300,56 @@ impl BeaconState {
             (slot % SLOTS_PER_EPOCH) * committees_per_slot + index,
             committees_per_slot * SLOTS_PER_EPOCH,
         )
+    }
+
+    /// Check if ``indexed_attestation`` is not empty, has sorted and unique indices and has a valid
+    /// aggregate signature.
+    pub fn is_valid_indexed_attestation(
+        &self,
+        indexed_attestation: &IndexedAttestation,
+    ) -> anyhow::Result<bool> {
+        let indices: Vec<usize> = indexed_attestation
+            .attesting_indices
+            .iter()
+            .map(|&i| i as usize)
+            .collect();
+        // Verify indices are sorted and unique
+        if indices.is_empty() || !is_sorted_and_unique(&indices) {
+            return Ok(false);
+        }
+
+        // Collect public keys of attesting validators
+        let pubkeys: Vec<_> = indices
+            .iter()
+            .filter_map(|&i| self.validators.get(i).map(|v| v.pubkey.clone()))
+            .collect();
+
+        // Compute domain and signing root
+        let domain = self.get_domain(
+            DOMAIN_BEACON_ATTESTER,
+            Some(indexed_attestation.data.target.epoch),
+        )?;
+
+        let sig = blst::min_pk::Signature::from_bytes(&indexed_attestation.signature.signature)
+            .map_err(|err| anyhow!("Signarure conversion failed:{:?}", err))?;
+        let signing_root = compute_signing_root(&indexed_attestation.data, domain);
+
+        let publickeys: Vec<blst::min_pk::PublicKey> = pubkeys
+            .iter()
+            .filter_map(|key| blst::min_pk::PublicKey::from_bytes(&key.inner).ok())
+            .collect();
+
+        let verification_result = sig.fast_aggregate_verify(
+            true,
+            signing_root.as_ref(),
+            DST,
+            publickeys.iter().collect::<Vec<_>>().as_slice(),
+        );
+
+        Ok(matches!(
+            verification_result,
+            blst::BLST_ERROR::BLST_SUCCESS
+        ))
     }
 
     /// Return the set of attesting indices corresponding to ``data`` and ``bits``.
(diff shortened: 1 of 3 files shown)


crates/common/consensus/src/predicates.rs
```diff
@@ -1,7 +1,74 @@
...
+// Check if ``validator`` is slashable.
+pub fn is_slashable_validator(validator: Validator, epoch: u64) -> bool {
+    if !validator.slashed
+        && validator.activation_epoch <= epoch
+        && epoch < validator.withdrawable_epoch
+    {
+        return true;
+    }
+    false
+}
```
KolbyML, line 74:
```suggestion
// Check if ``validator`` is slashable.
pub fn is_slashable_validator(validator: &Validator, epoch: u64) -> bool {
    !validator.slashed
        && validator.activation_epoch <= epoch
        && epoch < validator.withdrawable_epoch
}
```

as I said in the other PR, this function should be moved to `impl Validator` we don't need an if case here, and Validator should use a reference, but moving this into the Validator impl would fix that on its own assuming you use &self

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -620,4 +624,46 @@ impl BeaconState {
...
+        let current_epoch = self.get_current_epoch();
+        let indices_1: HashSet<_> = attestation_1.attesting_indices.iter().cloned().collect();
+        let indices_2: HashSet<_> = attestation_2.attesting_indices.iter().cloned().collect();
+
+        let mut slashed_any = false;
+
+        // Find common attesting indices and process slashing
+        for &index in indices_1.intersection(&indices_2).sorted() {
+            if is_slashable_validator(self.validators[index as usize].clone(), current_epoch) {
+                let _ = self.slash_validator(index, None);
```
KolbyML, line 660:
```suggestion
                let _ = self.slash_validator(index, None);
```

we need to handle the error here no?

KolbyML (commented):
Here is an initial look

crates/common/consensus/src/predicates.rs
```diff
@@ -1,7 +1,63 @@
...
+    }
+
+    // Collect public keys of attesting validators
+    let pubkeys: Vec<_> = indices
+        .iter()
+        .filter_map(|&i| state.validators.get(i).map(|v| v.pubkey.clone()))
+        .collect();
+
+    // Compute domain and signing root
+    let domain = match state.get_domain(
```
KolbyML, line 34: get_domain doesn't need to return a result, if you want to remove it in this PR I am cool with that, or I can, I think it is a little awkward to return false here

crates/common/consensus/src/predicates.rs
```diff
@@ -1,7 +1,63 @@
...
 
 /// Check if ``data_1`` and ``data_2`` are slashable according to Casper FFG rules.
-pub fn is_slashable_attestation_data(data_1: AttestationData, data_2: AttestationData) -> bool {
+pub fn is_slashable_attestation_data(data_1: &AttestationData, data_2: &AttestationData) -> bool {
     (data_1 != data_2 && data_1.target.epoch == data_2.target.epoch)
         || (data_1.source.epoch < data_2.source.epoch && data_2.target.epoch < data_1.target.epoch)
 }
+
+pub fn is_valid_indexed_attestation(
+    state: &BeaconState,
```
KolbyML, line 14: could you move this function to the BeaconState impl

KolbyML (approved):
:shipit: PR looks good

crates/common/consensus/src/predicates.rs
```diff
@@ -1,7 +1,63 @@
...
+        .iter()
+        .filter_map(|key| blst::min_pk::PublicKey::from_bytes(&key.inner).ok())
+        .collect();
+
+    let verification_result = sig.fast_aggregate_verify(
+        true,
+        signing_root.as_ref(),
+        domain.as_ref(),
+        publickeys.iter().collect::<Vec<_>>().as_slice(),
+    );
```
Kayden-ML, line 56:
```suggestion
pub const DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_ROPOP";
    let verification_result = sig.fast_aggregate_verify(
        true,
        signing_root.as_ref(),
        DST,
        publickeys.iter().collect::<Vec<_>>().as_slice(),
    );
```
Shouldn't it be this instead?

Kayden-ML (changes requested):
i think i found an issue here.

crates/common/consensus/src/predicates.rs
varun-doshi: I would assume the `get_domain` function returns the same thing?
cc: @KolbyML

KolbyML: I don't think `get_domain()` returns the same thing, do you have a source for that?

varun-doshi: so this is what `get_domain` returns: `Return the signature domain (fork version concatenated with domain type) of a message.`

varun-doshi: > I don't think `get_domain()` returns the same thing, do you have a source for that?

it doesn't but the return type from `get_domain` is what is used in the function...not generic BLS domain

KolbyML: The signing root utilizes domain. If you want you can add tests from the consensus-spec-tests repo to see if you or kayden is right then. It is always good to have more testing.

varun-doshi: resolved

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -284,6 +288,48 @@ impl BeaconState {
...
+            .collect();
+
+        // Compute domain and signing root
+        let domain = self.get_domain(
+            DOMAIN_BEACON_ATTESTER,
+            Some(indexed_attestation.data.target.epoch),
+        );
+
+        let sig =
+            blst::min_pk::Signature::from_bytes(&indexed_attestation.signature.signature).unwrap();
```
KolbyML, line 315: please do use unwrap

KolbyML (approved):
PR looks good

make sure to rebase using

```
git fetch upstream
git rebase upstream/master
```

crates/common/consensus/src/fork_choice/helpers/constants.rs
```diff
@@ -1,6 +1,7 @@
 use alloy_primitives::{aliases::B32, fixed_bytes};
 
 pub const BASE_REWARD_FACTOR: u64 = 64;
+pub const DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_ROPOP";
```
KolbyML, line 4: leave this const in beacon_state

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -292,6 +295,53 @@ impl BeaconState {
...
+        let domain = self.get_domain(
+            DOMAIN_BEACON_ATTESTER,
+            Some(indexed_attestation.data.target.epoch),
+        );
+
+        let sig =
+            match blst::min_pk::Signature::from_bytes(&indexed_attestation.signature.signature) {
+                Ok(sig) => sig,
+                Err(_) => return false,
+            };
```
KolbyML, line 327:
```suggestion
        let sig =blst::min_pk::Signature::from_bytes(&indexed_attestation.signature.signature).map_err(|err| anyhow!(..))?;
```

if this fails don't return false it is a error which should be returned

KolbyML (commented):
One more fix

Issue #48, opened 2025-01-30
Implement process_proposer_slashing()


unnawut:
Requires #42

Implement `consensus::state_transition::operations::process_proposer_slashing()` according to its latest specification (Phase0).

https://ethereum.github.io/consensus-specs/specs/phase0/beacon-chain/#proposer-slashings

```py
def process_proposer_slashing(state: BeaconState, proposer_slashing: ProposerSlashing) -> None:
    header_1 = proposer_slashing.signed_header_1.message
    header_2 = proposer_slashing.signed_header_2.message

    # Verify header slots match
    assert header_1.slot == header_2.slot
    # Verify header proposer indices match
    assert header_1.proposer_index == header_2.proposer_index
    # Verify the headers are different
    assert header_1 != header_2
    # Verify the proposer is slashable
    proposer = state.validators[header_1.proposer_index]
    assert is_slashable_validator(proposer, get_current_epoch(state))
    # Verify signatures
    for signed_header in (proposer_slashing.signed_header_1, proposer_slashing.signed_header_2):
        domain = get_domain(state, DOMAIN_BEACON_PROPOSER, compute_epoch_at_slot(signed_header.message.slot))
        signing_root = compute_signing_root(signed_header.message, domain)
        assert bls.Verify(proposer.pubkey, signing_root, signed_header.signature)

    slash_validator(state, header_1.proposer_index)
```

varun-doshi:
Closed as per #58

PR #58, opened 2025-01-31
feat: implement process_proposer_slashing

varun-doshi:
Ref #48

### DIFF
--- a/crates/common/consensus/src/deneb/beacon_state.rs
+++ b/crates/common/consensus/src/deneb/beacon_state.rs
@@ -57,6 +57,7 @@ use crate::{
         compute_activation_exit_epoch, compute_committee, compute_domain, compute_epoch_at_slot,
         compute_shuffled_index, compute_signing_root, compute_start_slot_at_epoch,
     },
+    proposer_slashing::ProposerSlashing,
     pubkey::PubKey,
     signature::BlsSignature,
     sync_committee::SyncCommittee,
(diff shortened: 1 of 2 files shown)


crates/common/consensus/src/predicates.rs
```diff
@@ -1,7 +1,17 @@
...
+
+pub fn is_slashable_validator(validator: Validator, epoch: u64) -> bool {
+    if !validator.slashed
+        && validator.activation_epoch <= epoch
+        && epoch < validator.withdrawable_epoch
+    {
+        return true;
+    }
+    false
+}
```
KolbyML, line 17:
```suggestion
pub fn is_slashable_validator(validator: &Validator, epoch: u64) -> bool {
    !validator.slashed
        && validator.activation_epoch <= epoch
        && epoch < validator.withdrawable_epoch
}
```

we don't need the if case for this, and Validator should probably be passed in as a reference, is_slashable_validator probably makes sense to be a method of the Validator Struct, so if you could move this function to `impl Validator` that would be great

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -620,4 +623,72 @@ impl BeaconState {
...
+
+        // Get the proposer and verify they are slashable
+        let proposer_index = header_1.proposer_index;
+        let proposer = self
+            .validators
+            .get(proposer_index as usize)
+            .ok_or_else(|| anyhow::anyhow!("Invalid proposer index"))?;
+
+        ensure!(
+            is_slashable_validator(proposer.clone(), self.get_current_epoch()),
```
KolbyML, line 654:
```suggestion
            proposer.is_slashable_validator(self.get_current_epoch()),
```

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -620,4 +623,72 @@ impl BeaconState {
...
+            ensure!(
+                verification_result == blst::BLST_ERROR::BLST_SUCCESS,
+                "BLS Signature verification failed!"
+            );
+        }
+
+        // Slash the validator
+        let _ = self.slash_validator(proposer_index, None);
+
+        Ok(())
```
KolbyML, line 692:
```suggestion
        self.slash_validator(proposer_index, None)
```

We aren't handling the error case from slash_validator, this will fix it

KolbyML (commented):
Here is some early feedback

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -620,4 +622,70 @@ impl BeaconState {
...
+                Some(compute_epoch_at_slot(signed_header.message.slot)),
+            )?;
+
+            let signing_root = compute_signing_root(&signed_header.message, domain);
+
+            let sig =
+                blst::min_pk::Signature::from_bytes(&signed_header.signature.signature).unwrap();
+
+            let public_key =
+                PublicKey::from_bytes(&proposer.pubkey.inner).expect("Cold not convert PublicKey");
```
KolbyML, line 673: don't use unwrap or expect do `.map_err(|err| anyhow!("xyz, {:?}", err)?;

KolbyML (commented):
here is some more feedback

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -620,4 +622,70 @@ impl BeaconState {
...
+
+            let verification_result = sig.fast_aggregate_verify(
+                true,
+                signing_root.as_ref(),
+                domain.as_ref(),
+                &[&public_key],
+            );
```
Kayden-ML, line 680:
```suggestion
pub const DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_ROPOP";
            let verification_result = sig.fast_aggregate_verify(
                true,
                signing_root.as_ref(),
                DST,
                &[&public_key],
            );
```

Shouldn't this use this instead?

Kayden-ML (dismissed):

crates/common/consensus/src/fork_choice/helpers/constants.rs
```diff
@@ -1,6 +1,7 @@
 use alloy_primitives::{aliases::B32, fixed_bytes};
 
 pub const BASE_REWARD_FACTOR: u64 = 64;
+pub const DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_ROPOP";
```
KolbyML, line 4: I don't think this constant belongs here, leave it in beacon_state

KolbyML (approved):
PR looks good make sure rebase

PR looks good

make sure to rebase using

```
git fetch upstream
git rebase upstream/master
```

PR #65, opened 2025-02-03
feat process_voluntary_exit

varun-doshi:
Ref #49 

https://ethereum.github.io/consensus-specs/specs/deneb/beacon-chain/#modified-process_voluntary_exit

PR #66, opened 2025-02-04
feat: implemented the process_voluntary_exit function

Kayden-ML:
I implemented the process_voluntary_exit function https://ethereum.github.io/consensus-specs/specs/deneb/beacon-chain/#modified-process_voluntary_exit


varun-doshi:
This is already in progress as per https://github.com/ReamLabs/ream/pull/65
Was assigned as well

PR #86, opened 2025-02-09
feat: implemented the process_operations function

Kayden-ML:
Implemented the process_operations function

https://ethereum.github.io/consensus-specs/specs/capella/beacon-chain/#modified-process_operations


KolbyML (approved):
:shipit: looks good

PR #94, opened 2025-02-11
feat: implemented the get_flag_index_deltas function

Kayden-ML:
implemented the get_flag_index_deltas function

https://ethereum.github.io/consensus-specs/specs/altair/beacon-chain/#get_flag_index_deltas

### DIFF
--- a/crates/common/consensus/src/deneb/beacon_state.rs
+++ b/crates/common/consensus/src/deneb/beacon_state.rs
@@ -1609,6 +1609,39 @@ impl BeaconState {
 
         Ok(())
     }
+
+    /// Return the deltas for a given ``flag_index`` by scanning through the participation flags.
+    pub fn get_flag_index_deltas(&self, flag_index: u8) -> anyhow::Result<(Vec<u64>, Vec<u64>)> {
+        let mut rewards = vec![0; self.validators.len()];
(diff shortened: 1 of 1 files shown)


crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -1609,6 +1609,68 @@ impl BeaconState {
...
+
+        for flag_index in 0..PARTICIPATION_FLAG_WEIGHTS.len() {
+            let delta = self.get_flag_index_deltas(flag_index as u8)?;
+            flag_deltas.push(delta);
+        }
+
+        let (penalties_rewards, penalties) = self.get_inactivity_penalty_deltas()?;
+
+        let mut deltas = flag_deltas;
+        deltas.push((penalties_rewards, penalties));
```
jihoonsong, line 1663:
```suggestion
        let mut deltas = vec![];

        for flag_index in 0..PARTICIPATION_FLAG_WEIGHTS.len() {
            deltas.push(self.get_flag_index_deltas(flag_index as u8)?);
        }

        deltas.push(self.get_inactivity_penalty_deltas()?);
```
How about this?

jihoonsong (commented):
It closely followed the spec making it easier to read, nice!

jihoonsong (approved):

PR #96, opened 2025-02-13
fix: nit block operations

syjn99:
1. `operation_objects` is read-only, so those could be just a reference.
2. `process_operations` has some misleading variable name, so fixed those.


KolbyML (approved):
:shipit: looks good

PR #98, opened 2025-02-13
fix: make DST correct

syjn99:
This could affect verification result, so replace the correct one.
Refer to [eth2book](https://eth2book.info/capella/part3/helper/crypto/#bls-signatures)


KolbyML (approved):
:shipit: looks good

PR #99, opened 2025-02-13
fix: check overflow at `process_voluntary_exit`

syjn99:
I found this issue while integrating with EF tests. `validator.activation_epoch` can be the max of `u64`.


KolbyML (commented):
I think this change is valid and correct, but if we are making a PR to fix this could be also include the test in this PR

KolbyML (approved):
I won't block this PR as I know it is right, although I think the next PR should add tests

PR #100, opened 2025-02-13
fix: make `decrease_balance` work

syjn99:
`saturating_sub` doesn't mutate the lhs value, so let it mutate


KolbyML (approved):
:shipit: looks good

PR #103, opened 2025-02-14
feat: update `apply_deposit` and `add_validator_to_registry` to altair changes

syjn99:
1. `apply_deposit` should ignore the corresponding deposit when pubkey is in wrong format.
2. `add_validator_to_registry` should append new entry for the corresponding validator to `previous_epoch_participation`, `current_epoch_participation`, and `inactivity_scores`. This was added at Altair fork. ([Link](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md#modified-add_validator_to_registry))

I added `test_operation!` for `process_deposit`, to check whether I correctly write the code.

### DIFF
--- a/crates/common/consensus/src/deneb/beacon_state.rs
+++ b/crates/common/consensus/src/deneb/beacon_state.rs
@@ -798,10 +798,10 @@ impl BeaconState {
             .push(amount)
             .map_err(|err| anyhow!("Couldn't push to balances {:?}", err))?;
         self.previous_epoch_participation
-            .push(0b00000000)
+            .push(0)
             .map_err(|err| anyhow!("Couldn't push to previous_epoch_participation {:?}", err))?;
         self.current_epoch_participation
-            .push(0b00000000)
+            .push(0)
             .map_err(|err| anyhow!("Couldn't push to current_epoch_participation {:?}", err))?;
         self.inactivity_scores
             .push(0)
--- a/crates/common/consensus/src/deneb/beacon_state.rs
+++ b/crates/common/consensus/src/deneb/beacon_state.rs
@@ -797,6 +797,15 @@ impl BeaconState {
         self.balances
             .push(amount)
             .map_err(|err| anyhow!("Couldn't push to balances {:?}", err))?;
+        self.previous_epoch_participation
+            .push(0b00000000)
+            .map_err(|err| anyhow!("Couldn't push to previous_epoch_participation {:?}", err))?;
+        self.current_epoch_participation
+            .push(0b00000000)
+            .map_err(|err| anyhow!("Couldn't push to current_epoch_participation {:?}", err))?;
+        self.inactivity_scores
+            .push(0)
+            .map_err(|err| anyhow!("Couldn't push to inactivity_scores {:?}", err))?;
         Ok(())
     }
 
@@ -823,8 +832,10 @@ impl BeaconState {
             let signing_root = compute_signing_root(deposit_message, domain);
             let sig = blst::min_pk::Signature::from_bytes(&signature.signature)
                 .map_err(|err| anyhow!("Failed to convert signiture type {err:?}"))?;
-            let public_key = PublicKey::from_bytes(&pubkey.inner)
-                .map_err(|err| anyhow!("Failed to convert pubkey type {err:?}"))?;
+            let public_key = match PublicKey::from_bytes(&pubkey.inner) {
+                Ok(pk) => pk,
+                Err(_) => return Ok(()) // Skip deposits with invalid public keys
+            };
             let verification_result =
                 sig.fast_aggregate_verify(true, signing_root.as_ref(), DST, &[&public_key]);
             if verification_result == blst::BLST_ERROR::BLST_SUCCESS {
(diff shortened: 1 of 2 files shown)


testing/ef-tests/tests/tests.rs
```diff
@@ -64,3 +64,6 @@ test_consensus_type!(VoluntaryExit);
 test_consensus_type!(Withdrawal);
 
 // Testing operations for block processing
+use ef_tests::{test_operation, utils};
```
jihoonsong, line 67: This can be added to L3.

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -823,8 +832,10 @@ impl BeaconState {
             let signing_root = compute_signing_root(deposit_message, domain);
             let sig = blst::min_pk::Signature::from_bytes(&signature.signature)
                 .map_err(|err| anyhow!("Failed to convert signiture type {err:?}"))?;
-            let public_key = PublicKey::from_bytes(&pubkey.inner)
-                .map_err(|err| anyhow!("Failed to convert pubkey type {err:?}"))?;
+            let public_key = match PublicKey::from_bytes(&pubkey.inner) {
+                Ok(pk) => pk,
+                Err(_) => return Ok(()), // Skip deposits with invalid public keys
```
jihoonsong, line 837: I don't think this is right. It already ignores the corresponding deposit when pubkey is in wrong format. This just changes it to return Ok instead of proper error.

jihoonsong (commented):
Nice catch on `add_validator_to_registry`!

crates/common/consensus/src/deneb/beacon_state.rs
syjn99: https://github.com/ReamLabs/ream/blob/5d5102886b22054a1a02455ceab9c84fc573d6c5/testing/ef-tests/src/macros.rs#L114-L129

Using `?` operator will propagate error, making this operation returns error. It made `key_validate_invalid_decompression` test case fail like: 

```
---- tests_process_deposit::test_operation stdout ----
Testing case: key_validate_invalid_decompression
thread 'tests_process_deposit::test_operation' panicked at testing/ef-tests/tests/tests.rs:69:1:
Test case key_validate_invalid_decompression should have succeeded but failed
```

I thought this should not be an error case.

Any feedback for this? Do I have to change the current macro(`test_operation`)?

KolbyML: Just check the expected post no? If they give an invalid key we want to propagate an error, we don't need to verify if the function ran successfully or not.

`key_validate_invalid_decompression` the test case is an error case, so it should error

syjn99: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/operations#postssz_snappy

If `post.ssz_snappy` is present, the operation should not be aborted: `key_validate_invalid_decompression` has it 
 while other test castee like `invalid_wrong_deposit_for_deposit_count` doesn't.

AFAIK deposit contract doesn't validate the pubkey itself(it only validates the length). So user can put some invalid form in his deposit data, and that action shouldn't abort the state transition.

jihoonsong: I've checked Prysm and Lighthouse implementations and I think @syjn99 is right. When `from_bytes` call to either of signature and public key fails, it should just return Ok(). Nice catch.

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -797,6 +797,15 @@ impl BeaconState {
         self.balances
             .push(amount)
             .map_err(|err| anyhow!("Couldn't push to balances {:?}", err))?;
+        self.previous_epoch_participation
+            .push(0b00000000)
+            .map_err(|err| anyhow!("Couldn't push to previous_epoch_participation {:?}", err))?;
+        self.current_epoch_participation
+            .push(0b00000000)
+            .map_err(|err| anyhow!("Couldn't push to current_epoch_participation {:?}", err))?;
```
KolbyML, line 805:
```suggestion
        self.previous_epoch_participation
            .push(0)
            .map_err(|err| anyhow!("Couldn't push to previous_epoch_participation {:?}", err))?;
        self.current_epoch_participation
            .push(0)
            .map_err(|err| anyhow!("Couldn't push to current_epoch_participation {:?}", err))?;
```

the extra 0's aren't need

crates/common/consensus/src/deneb/beacon_state.rs
KolbyML: Good find @syjn99 :saluting_face:

testing/ef-tests/tests/tests.rs
KolbyML: Just to confirm this import will be  moved to the top of the file right?

KolbyML (approved):
:shipit: great work

testing/ef-tests/tests/tests.rs
syjn99: Done at https://github.com/ReamLabs/ream/pull/103/commits/b1628f6ef5643231367ad65522a34d4bab06f3f6

crates/common/consensus/src/deneb/beacon_state.rs
syjn99: Done at https://github.com/ReamLabs/ream/pull/103/commits/b1628f6ef5643231367ad65522a34d4bab06f3f6

PR #104, opened 2025-02-16
fix: process_withdrawals and add corresponding ef-tests

KolbyML:
We are trying to support all the `consensus-spec-tests`, I looked at the work @syjn99 was doing and wanted to instead target testcases which are more minimal and slowly work my way up to the more complex test cases, as then it should be smoother to progressively support more test cases.

I got the operation `Withdrawals` tests working.

I found 2 bugs
- We weren't decoding the addresses properly
- We were doing `if X == 0`, when the spec asked for `if X != 0`

After I fixed these 2 problems, the `process_withdrawal` tests started passing :rocket: 

Lets keep knocking these tests out :fire:

### DIFF
--- a/testing/ef-tests/src/macros/mod.rs
+++ b/testing/ef-tests/src/macros/mod.rs
@@ -1,0 +1,2 @@
+pub mod operations;
+pub mod ssz_static;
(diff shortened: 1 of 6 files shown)


testing/ef-tests/src/macros/mod.rs
syjn99, line 1: Nice separation! 👍

syjn99 (approved):
LGTM

PR #107, opened 2025-02-18
test: add operation tests for `process_voluntary_exit` & `process_bls_to_execution_change`

syjn99:
This PR adds some operation tests that are currently passing the cases.
- `process_voluntary_exit`
- `process_bls_to_execution_change`

### DIFF
--- a/testing/ef-tests/tests/tests.rs
+++ b/testing/ef-tests/tests/tests.rs
@@ -66,8 +66,20 @@ test_consensus_type!(Withdrawal);
 // Testing operations for block processing
 test_operation!(deposit, Deposit, "deposit", process_deposit);
 test_operation!(
+    voluntary_exit,
+    SignedVoluntaryExit,
+    "voluntary_exit",
+    process_voluntary_exit
+);
+test_operation!(
     withdrawals,
     ExecutionPayload,
     "execution_payload",
     process_withdrawals
+);
+test_operation!(
+    bls_to_execution_change,
+    SignedBLSToExecutionChange,
+    "address_change",
+    process_bls_to_execution_change
 );


testing/ef-tests/tests/tests.rs
```diff
@@ -65,9 +65,21 @@ test_consensus_type!(Withdrawal);
...
     ExecutionPayload,
     "execution_payload",
     process_withdrawals
 );
+test_operation!(
+    bls_to_execution_change,
+    SignedBLSToExecutionChange,
+    "address_change",
+    process_bls_to_execution_change
+);
```
jihoonsong, line 85: It's a nit but I guess we want to sort this alphabetically by `operation_name` or `operation_object`. `operation_object` seems better imho.

syjn99: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/operations#condition

Here's the rationale: I wanted to sort the test in the same way as `consensus-specs` does.

KolbyML: I think we should sort it alphabetically instead of chronologically

syjn99: Alright, then I will sort it alphabetically. I think sorting by `operation_name` would be better than `operation_object` as `operation_name` clearly shows which operation to test.

KolbyML: > Alright, then I will sort it alphabetically. I think sorting by `operation_name` would be better than `operation_object` as `operation_name` clearly shows which operation to test.

Did you forget to sort it?

syjn99: Yes... already merged this

Issue #101, opened 2025-02-13
`compute_shuffled_index` doesn't return the correct value


syjn99:
I'm currently integrating the operations with [consensus-specs](https://github.com/ethereum/consensus-specs) test at #97, and found that `compute_proposer_index` doesn't return the correct value.

# Reproduction

1. Go to #97 branch.
2. Run test via `make test` at `testing/ef-tests` folder.

```bash
$ make test &> test.log
```

# Log

```plaintext
---- tests_process_attester_slashing::test_operation stdout ----
Testing case: invalid_all_empty_indices
Skipping test case invalid_all_empty_indices because bls_setting is 1
Testing case: invalid_no_double_or_surround
Testing case: invalid_unsorted_att_2
Testing case: invalid_att1_bad_extra_index
Skipping test case invalid_att1_bad_extra_index because bls_setting is 1
Testing case: invalid_att1_bad_replaced_index
Skipping test case invalid_att1_bad_replaced_index because bls_setting is 1
Testing case: invalid_att1_duplicate_index_normal_signed
Skipping test case invalid_att1_duplicate_index_normal_signed because bls_setting is 1
Testing case: basic_surround
Balance mismatch at index 94: actual=31999999302, expected=32499999302
Balance mismatch at index 120: actual=32499999302, expected=31999999302
thread 'tests_process_attester_slashing::test_operation' panicked at testing/ef-tests/tests/tests.rs:71:1:
assertion `left == right` failed: Post state mismatch in case basic_surround
```

## Analysis

The log says that `compute_proposer_index` returns `120`, not `94`.

KolbyML:
I am assuming #97 needs to be rebased or something? As I can't replicate your log output running the PR without changes

syjn99:
@KolbyML I rebased the branch. Could you check?

KolbyML:
From debugging even if you fixed compute_proposer_index, the same indexing error occurs

KolbyML:
I am removing myself as a assignee as there are more problems then just mentioned in this problem, and I don't think `compute_proposer_index` is necessarily the crux of the problem

PR #108, opened 2025-02-18
test: add shuffling test

syjn99:
Add shuffling test for `compute_shuffled_index`.

PR #109, opened 2025-02-18
fix: compute_shuffled_index computing wrong index

KolbyML:
fixes https://github.com/ReamLabs/ream/issues/101 , I tested the PR with the test macro @syjn99 added here https://github.com/ReamLabs/ream/pull/108


jihoonsong (approved):
Looks good. It's not about this PR but I think we can remove L30 and change L25 as ```mut index: usize,```.

hopinheimer (approved):
LGTM nice catch!

PR #114, opened 2025-02-21
test: add more opertation tests

syjn99:
We now need two operation tests:
- `process_sync_aggregate`
- `process_execution_payload`: Not implemented yet

It seems like `process_sync_aggregate` has some problem. I would open an issue for this.


syjn99:
#115 describes the current problem at `process_sync_aggregate`.

Kayden-ML (approved):
Just went over this PR and it looks good to me 👍

Issue #51, opened 2025-01-30
Implement process_execution_payload()


unnawut:
Requires #42

Implement `consensus::state_transition::operations::process_execution_payload()` according to its latest specification (Deneb).

https://ethereum.github.io/consensus-specs/specs/deneb/beacon-chain/#modified-process_execution_payload

```py
def process_execution_payload(state: BeaconState, body: BeaconBlockBody, execution_engine: ExecutionEngine) -> None:
    payload = body.execution_payload

    # Verify consistency of the parent hash with respect to the previous execution payload header
    assert payload.parent_hash == state.latest_execution_payload_header.block_hash
    # Verify prev_randao
    assert payload.prev_randao == get_randao_mix(state, get_current_epoch(state))
    # Verify timestamp
    assert payload.timestamp == compute_timestamp_at_slot(state, state.slot)

    # [New in Deneb:EIP4844] Verify commitments are under limit
    assert len(body.blob_kzg_commitments) <= MAX_BLOBS_PER_BLOCK

    # Verify the execution payload is valid
    # [Modified in Deneb:EIP4844] Pass `versioned_hashes` to Execution Engine
    # [Modified in Deneb:EIP4788] Pass `parent_beacon_block_root` to Execution Engine
    versioned_hashes = [kzg_commitment_to_versioned_hash(commitment) for commitment in body.blob_kzg_commitments]
    assert execution_engine.verify_and_notify_new_payload(
        NewPayloadRequest(
            execution_payload=payload,
            versioned_hashes=versioned_hashes,
            parent_beacon_block_root=state.latest_block_header.parent_root,
        )
    )

    # Cache execution payload header
    state.latest_execution_payload_header = ExecutionPayloadHeader(
        parent_hash=payload.parent_hash,
        fee_recipient=payload.fee_recipient,
        state_root=payload.state_root,
        receipts_root=payload.receipts_root,
        logs_bloom=payload.logs_bloom,
        prev_randao=payload.prev_randao,
        block_number=payload.block_number,
        gas_limit=payload.gas_limit,
        gas_used=payload.gas_used,
        timestamp=payload.timestamp,
        extra_data=payload.extra_data,
        base_fee_per_gas=payload.base_fee_per_gas,
        block_hash=payload.block_hash,
        transactions_root=hash_tree_root(payload.transactions),
        withdrawals_root=hash_tree_root(payload.withdrawals),
        blob_gas_used=payload.blob_gas_used,  # [New in Deneb:EIP4844]
[...]

PR #116, opened 2025-02-21
feat: implement eth syncing json engine api rpc call

Kayden-ML:
This is required to implement process_execution_payload() https://github.com/ReamLabs/ream/issues/51

implementing json engine api rpc calls


crates/common/consensus/src/execution_engine/utils.rs
```diff
@@ -0,0 +1,33 @@
+use serde::{Deserialize, Serialize};
+
+pub fn strip_prefix(s: &str) -> &str {
+    if let Some(stripped) = s.strip_prefix("0x") {
+        stripped
+    } else {
+        s
```
KolbyML, line 7:
```suggestion

pub fn strip_prefix(string: &str) -> &str {
    if let Some(stripped) = string.strip_prefix("0x") {
        stripped
    } else {
        string
```

KolbyML (approved):
:shipit: looks good

PR #117, opened 2025-02-21
feat: add `bls` crate for modularizing & integrate `zkcrypto/bls12_381`

### LINKED ISSUE #70: Replace blst BLS library with zkcrypto/bls12_381 for RISC-V compatibility
Replace blst BLS library with https://github.com/zkcrypto/bls12_381 which is RISC-V compatible which is important for Snarcification prototyping

syjn99:
Resolves #70 

Add `bls` crate. This crate has two features rn: `supranational` and `zkcrypto`. For each feature, it has a corresponding module(`crate/crypto/bls/src/{supranational,zkcrypto}`) that implements the basic aggregation and verification. As we don't have any signing feature right now, we don't have a struct like `SecretKey` or something.

## Rationale

```toml
# ream dependencies
ream-bls = { path = "crates/crypto/bls", features = ["zkcrypto"] } # Default feature is zkcrypto
```
- As default feature makes the linter unhappy, I chose to remove default feature and use it explicitly like `Cargo.toml` in the project root.
- `PubKey` and `Signature` is just a `FixedVector` of bytes. For each module, it implements traits(`Aggregatable` and `Verifiable`) for the functionality.


KolbyML (changes requested):
When this is ready for review I want to review it

syjn99:
@KolbyML I would open this up ASAP

KolbyML:
@syjn99 so are we
- supporting both blst and zkcrypto/bls12_381 or only one?

I have a few more questions but we can start with that

crates/crypto/bls/src/supranational/signature.rs
```diff
@@ -0,0 +1,145 @@
...
+        self.to_blst_signature()
+            .and_then(|sig| {
+                let public_keys = pubkeys
+                    .as_ref()
+                    .iter()
+                    .map(|key| key.to_blst_pubkey())
+                    .collect::<Result<Vec<_>, _>>()?;
+                Ok((sig, public_keys))
+            })
+            .is_ok_and(|(sig, public_keys)| {
```
KolbyML, line 136: We shouldn't hide the error using an `is_ok_and` we should return it

crates/crypto/bls/src/supranational/signature.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// - If the public key bytes are invalid or malformed
+    /// - If the actual signature verification fails
+    ///
+    /// # Arguments
+    /// * `pubkey` - The public key to verify against
+    /// * `message` - The message that was signed
+    ///
+    /// # Returns
+    /// `bool` - true if the signature is valid, false otherwise
+    pub fn verify(&self, pubkey: &PubKey, message: &[u8]) -> bool {
```
KolbyML, line 106: We shouldn't hide the error using an `is_ok_and` we should return it

crates/crypto/bls/src/supranational/signature.rs
syjn99: Yes, I would like to discuss with this also. It's quite similar as we discussed at #103.

I think we should hide error in this case. All of the errors here are `BLST_ERROR`s, which is fully dependent on `blst`. This would not be available on other libraries like `zkcrypto/bls12_381` or else. Here is the `BLST_ERROR`.

```rust
#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
pub enum BLST_ERROR {
    BLST_SUCCESS = 0,
    BLST_BAD_ENCODING = 1,
    BLST_POINT_NOT_ON_CURVE = 2,
    BLST_POINT_NOT_IN_GROUP = 3,
    BLST_AGGR_TYPE_MISMATCH = 4,
    BLST_VERIFY_FAIL = 5,
    BLST_PK_IS_INFINITY = 6,
    BLST_BAD_SCALAR = 7,
}
```

Other reason for this is that returning `bool` type is more natural imho. The function signatures for `bls` module is provided like [this](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#bls-signatures). Developers don't have to think of what's going underneath this module.
Returning bool type is also useful to deal with the functions like `apply_deposit`. [`apply_deposit`](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#deposits) doesn't throw an error if anything gone wrong. [`is_valid_indexed_attestation`](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#is_valid_indexed_attestation) also returns bool type. We could check by `ensure!` whether the verification result is true or not at the high level.

syjn99:
> supporting both blst and zkcrypto/bls12_381 or only one?

I thought supporting both (or more if there's any better alternative) and choosing with `features`. This framework could also be useful for other cryptographies(e.g. Winternitz-OTS).

crates/crypto/bls/src/supranational/signature.rs
KolbyML: We can define our own error enum using this error https://docs.rs/thiserror/latest/thiserror/

> [apply_deposit](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/beacon-chain.md#deposits) doesn't throw an error if anything gone wrong

That is fine we can handle the error in apply_deposit and choose to return ok if an error occurs, but we shouldn't hide the error, we should let the developer calling the functions on how to handle it, instead of choosing how to handle it for them

KolbyML:
> > supporting both blst and zkcrypto/bls12_381 or only one?
> 
> I thought supporting both (or more if there's any better alternative) 

I don't think we need to support both unless there is a good reason as it adds development overhead, unless there is a clear reason and value proposition to support both, we can always add blst support back in the future if we need it. To my knowledge `zkcrypto/bls12_381` is unaudited, but it can be compiled to riscv5 which is required for our ZKvm PoC's.

We are building a minimal Beacon Chain client right now for various reasons
- making PoC for ZKvm STF
- Helping in Beacon to Beam chain testnets (and mainnet?)
- Getting a head start so we can reuse code in the overlap of Beacon -> Beam

I don't think the minimal Beacon Chain client we are implementing is a waste or that throwaway, I am pretty sure it will be useful in some way or another, and it already is for the reasons stated above. Unless we try to compete with production Beacon chain clients, which we might 5 years is a long time then maybe it makes sense to add blst back, but unless we are trying to bring a Beacon Chain client into production to help onboard users and to secure funding I don't see a reason to support blst for the time being.

> The main goal of this PR is to add an interface level for modularizing the BLS functions and structs.
Could you add `zkcrypto/bls12_381` in this PR as well so I can see what you mean by `interface level for modularizing`


I am open to feedback, I like how we are simplifying the code, but I don't think we should be supporting both libraries for the time being, if we need blst in the future it should be super easy to add back, but currently it is adding tech debt we don't directly benefit from. If you have any good arguments for why we should support both `blst` and `zkcrypto/bls12_381` right now, I encourage you to fight me on it, and we can see where the discussion goes.

:saluting_face:

crates/crypto/bls/src/supranational/signature.rs
syjn99: Understood. Then I'm gonna think of which errors should be presented, and then handle those errors at the caller frame.

syjn99:
First of all, thanks @KolbyML for your genuine comment for my work! I really appreciate for this discussion and your interest.

We could separate one agenda into two:

- Agenda 1: Add `bls` module and maintain it in a separate way. <- This is what this PR wants to do. I thought we're getting aligned with this agenda.
- Agenda 2: Support multiple backends for BLS scheme. <- This is what you are feeling unnecessary with. But I think it is still meaningful to support more than one backend.

> as it adds development overhead.

I don't consider this could be an overhead. Verification feature using `blst` is already implemented, and I just migrated some files into `bls` module as you can see in this PR.

> we can always add blst support back in the future if we need it.

Also I think it could be hard to add `blst` support later on. Still, `blst` is a de facto standard in any other projects using BLS.

> Could you add zkcrypto/bls12_381 in this PR as well so I can see what you mean by interface level for modularizing

Sure, the discussion would be much clear if I show the whole picture of what I think of. I will include those works asap.

crates/crypto/bls/src/supranational/signature.rs
syjn99: https://github.com/ReamLabs/ream/pull/117/commits/84c9762a16350d5dd3222641841855bd25e8549b added a custom error.

syjn99:
@KolbyML I've been working on this, done of integrating `zkcrypto/bls12_381`(EF tests are passing). But I'm struggling on Rust features... It's really annoying than I expected at the first.

```
cargo clippy --all --all-targets --all-features --no-deps -- --deny warnings
```

As `make lint` runs clippy in `all-features` mode and those two features are mutually exclusive, linter is unhappy on this.  To make this run, I changed our `lint` command like:

```
cargo clippy --all --all-targets --features "$(FEATURES)" --no-deps -- --deny warnings
```

But I'm afraid this is not a desired option for us. Any idea on this?

KolbyML:
> As `make lint` runs clippy in `all-features` mode and those two features are mutually exclusive, linter is unhappy on this. To make this run, I changed our `lint` command like:

I would make `make lint` run 2 different clippy commands then, one line using blst and one line using the other one

KolbyML:
Can we call the features "bls_c" and "bls_rust", supernatural and zkcrypto aren't that clear on what the features are for

syjn99:
> Can we call the features "bls_c" and "bls_rust", supernatural and zkcrypto aren't that clear on what the features are for

It could be much confusing imho, so I selected the vendor name of the library. There are other library written in Rust: e.g. from [artworks](https://github.com/arkworks-rs/algebra/tree/v0.5.0).

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -327,41 +327,23 @@ impl BeaconState {
...
-            .filter_map(|&i| self.validators.get(i).map(|v| v.pubkey.clone()))
-            .collect();
-
-        // Compute domain and signing root
         let domain = self.get_domain(
             DOMAIN_BEACON_ATTESTER,
             Some(indexed_attestation.data.target.epoch),
         );
+        let signing_root: alloy_primitives::FixedBytes<32> =
+            compute_signing_root(&indexed_attestation.data, domain);
```
KolbyML, line 335:
```suggestion
        let signing_root = compute_signing_root(&indexed_attestation.data, domain);
```

crates/crypto/bls/src/constants.rs
```diff
@@ -0,0 +1 @@
+pub const DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_";
```
KolbyML, line 1: Can you remove the other DST const in the codebase if you are moving it here

crates/crypto/bls/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+pub mod traits;
+
+pub use aggregate_pubkey::AggregatePubKey;
+pub use pubkey::PubKey;
+pub use signature::BLSSignature;
+
+// The BLS (Boneh-Lynn-Shacham) cryptographic backend implementation is determined
+// at compile time via feature flags. Two implementations are supported:
+// - "supranational": Uses the supranational/blst library, optimized for performance
+// - "zkcrypto": Uses the zkcrypto/bls12_381 library implementation, optimized for zkVMs
```
KolbyML, line 15:
```suggestion
//! The BLS (Boneh-Lynn-Shacham) cryptographic backend implementation is determined
//! at compile time via feature flags. Two implementations are supported:
//! - "supranational": Uses the supranational/blst library, optimized for performance
//! - "zkcrypto": Uses the zkcrypto/bls12_381 library implementation, optimized for zkVMs

pub mod aggregate_pubkey;
pub mod constants;
pub mod errors;
pub mod pubkey;
pub mod signature;
pub mod traits;

pub use aggregate_pubkey::AggregatePubKey;
pub use pubkey::PubKey;
pub use signature::BLSSignature;

```

Most this to the top of the file  and make it inner doc comment's. Could you also make a README for the crate explaining how we support 2 different bls libraries which ones and examples and how to use the crate

KolbyML (commented):
Mostly looks good, I think we should improve the documentation a bit more then other then that it looks pretty good :fire: 

great stuff

crates/crypto/bls/src/lib.rs
syjn99: All reviews are resolved at https://github.com/ReamLabs/ream/pull/117/commits/9fa566ba68ac3909ca6a5f8252570951f7ccaa83! Thanks 👍

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -832,16 +811,13 @@ impl BeaconState {
...
-                sig.verify(true, signing_root.as_ref(), DST, &[], &public_key, false);
-            if verification_result == blst::BLST_ERROR::BLST_SUCCESS {
-                self.add_validator_to_registry(pubkey, withdrawal_credentials, amount)?;
+
+            match signature.verify(&pubkey, signing_root.as_ref()) {
+                Ok(true) => {
+                    self.add_validator_to_registry(pubkey, withdrawal_credentials, amount)?
+                }
+                _ => return Ok(()), /* Skip adding validator if either pubkey or signature is
+                                     * invalid */
```
KolbyML, line 820:
```suggestion
                // Skip adding validator if either pubkey or signature is invalid
                _ => return Ok(()),
```

crates/crypto/bls/src/supranational/signature.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+impl Verifiable for BLSSignature {
+    type Error = BLSError;
+
+    fn verify(&self, pubkey: &PubKey, message: &[u8]) -> Result<bool, BLSError> {
+        let sig = self.to_blst_signature()?;
+        let pk = pubkey.to_blst_pubkey()?;
+
+        Ok(sig.verify(true, message, DST, &[], &pk, false) == BLST_ERROR::BLST_SUCCESS)
```
KolbyML, line 24: can we use signature and public_key instead of the shorthands for readability

crates/crypto/bls/src/supranational/signature.rs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        let sig = self.to_blst_signature()?;
+        let public_keys = pubkeys
+            .as_ref()
+            .iter()
+            .map(|key| key.to_blst_pubkey())
+            .collect::<Result<Vec<_>, _>>()?;
+
+        Ok(
+            sig.fast_aggregate_verify(true, message, DST, &public_keys.iter().collect::<Vec<_>>())
```
KolbyML, line 39: can we use signature instead of the shorthands for readability

crates/crypto/bls/src/zkcrypto/signature.rs
```diff
@@ -0,0 +1,67 @@
...
+            DST,
+        );
+
+        let gt1 = pairing(
+            &G1Affine::try_from(&agg_pubkey.to_pubkey())?,
+            &G2Affine::from(h),
+        );
+        let gt2 = pairing(&G1Affine::generator(), &G2Affine::try_from(self)?);
+
+        Ok(gt1 == gt2)
```
KolbyML, line 63: can we not use shorthands like `agg_pubkey`, `h`, `gt1`, and `gt2`. Avoiding shorthands makes it easier for new contributors to read the code.

KolbyML (approved):
:fire: great work :rocket:

crates/crypto/bls/src/zkcrypto/signature.rs
syjn99: As terms like `gt` are widely used in BLS scheme, it's quite hard to think of good variable name for this... Applied for `agg_pubkey`.

syjn99:
I'd like to get one more approval for merging this!

Kayden-ML (approved):
Just went over this looks good to me 👍 
Looks really good nice job!

PR #118, opened 2025-02-21
fix: not to use `fast_aggregate_verify` for all BLS verification

syjn99:
I've been working on #70 and #117, and found that our implementation regarding with BLS verification relies on `fast_aggregate_verify`, which is a optimized method suggested by Vitalik. ([Link](https://ethresear.ch/t/fast-verification-of-multiple-bls-signatures/5407))

```python
def process_randao(state: BeaconState, body: BeaconBlockBody) -> None:
    epoch = get_current_epoch(state)
    # Verify RANDAO reveal
    proposer = state.validators[get_beacon_proposer_index(state)]
    signing_root = compute_signing_root(epoch, get_domain(state, DOMAIN_RANDAO))
    assert bls.Verify(proposer.pubkey, signing_root, body.randao_reveal)
    # Mix in RANDAO reveal
    mix = xor(get_randao_mix(state, epoch), hash(body.randao_reveal))
    state.randao_mixes[epoch % EPOCHS_PER_HISTORICAL_VECTOR] = mix
```

For the most of case including `process_randao`, it isn't necessary and spec also says to use `bls.Verify`.  This PR fixes to replace `fast_aggregate_verify` with simple `verify`.

NOTE: I'm currently working on abstracting BLS related code, so code is quite duplicative.
NOTE 2: This work could be done as a part of #117, but I separated this work for better code reviews. Diffs often make us confused...

### DIFF
--- a/crates/common/consensus/src/deneb/beacon_state.rs
+++ b/crates/common/consensus/src/deneb/beacon_state.rs
@@ -839,7 +839,7 @@ impl BeaconState {
                 Err(_) => return Ok(()), // Skip deposits with invalid public keys
             };
             let verification_result =
-                sig.fast_aggregate_verify(true, signing_root.as_ref(), DST, &[&public_key]);
+                sig.verify(true, signing_root.as_ref(), DST, &[], &public_key, false);
             if verification_result == blst::BLST_ERROR::BLST_SUCCESS {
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit: looks good

:smirk: diffs don't make me confused

Issue #85, opened 2025-02-09
Implement Attestation Delta functions


varun-doshi:
Implement Attestation Delta functions
Ref: https://ethereum.github.io/consensus-specs/specs/phase0/beacon-chain/#components-of-attestation-deltas
Ref: https://ethereum.github.io/consensus-specs/specs/phase0/beacon-chain/#get_attestation_deltas

PR #126, opened 2025-02-24
feat: implement get_attestation_component_delta

varun-doshi:
This is required to implement the other Attestation Delta functions

Part of #85 

Specs available here: https://ethereum.github.io/consensus-specs/specs/phase0/beacon-chain/#components-of-attestation-deltas


syjn99:
@varun-doshi I'm afraid that `get_attestation_component_delta` is a legacy function, and you should refer to the altair spec.

https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md#get_flag_index_deltas

Issue #128, opened 2025-02-24
Implement `process_participation_flag_updates()` for `process_epoch()`


syjn99:
https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md#participation-flags-updates

PR #129, opened 2025-02-24
feat: implement process_participation_flag_updates

varun-doshi:
Ref #128 

Spec: https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md#participation-flags-updates


syjn99 (approved):

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -1654,6 +1654,13 @@ impl BeaconState {
         }
         Ok(())
     }
+
+    pub fn process_participation_flag_updates(&mut self) -> anyhow::Result<()> {
+        self.previous_epoch_participation = self.current_epoch_participation.clone();
+        self.current_epoch_participation =
+            VariableList::from(vec![0b0000_0000; self.validators.len()]);
```
KolbyML, line 1661:
```suggestion
        self.current_epoch_participation = vec![0; self.validators.len()].into();
```

We should be able to do this then it will be 1 line instead of 2

KolbyML (approved):

crates/common/consensus/src/deneb/beacon_state.rs
KolbyML: I don't think there is value in matching the spec in specifying `0b0000_0000`, 0 is more then enough

jihoonsong (approved):

PR #138, opened 2025-02-24
feat: add network cli for specifying network specification

KolbyML:
Just the basic for adding multi network support. We will specify hardfork timestamps and such in the NetworkSpec. Any network specific configurations will belong here. As we build our client out the idea will be to for example 

```
if network.is_electra_enabled(timestamp) {
    // do xyz
}

```

by adding this now hopefully this gives everybody an idea and as we build the puzzle pieces will fall into place around this

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3792,6 +3792,7 @@ dependencies = [
  "discv5",
  "ream-discv5",
  "ream-executor",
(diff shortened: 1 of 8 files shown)


Kayden-ML (approved):
Just went over this looks good 👍

Issue #141, opened 2025-02-25
Add spec test for process_block()


Kayden-ML:
https://github.com/ReamLabs/ream/issues/127#issuecomment-2677561403

Kayden-ML:
this issue was fixed in this PR https://github.com/ReamLabs/ream/pull/168

Issue #127, opened 2025-02-24
Implement process_block() function


Kayden-ML:
Implement process_block() function

https://ethereum.github.io/consensus-specs/specs/capella/beacon-chain/#block-processing

syjn99:
After `process_block()` is implemented, we can add spec tests for it.

https://github.com/ethereum/consensus-specs/blob/dev/tests/formats/sanity/blocks.md

Kayden-ML:
Implemented in this PR https://github.com/ReamLabs/ream/pull/142

Issue #88, opened 2025-02-09
Implement state_transition() function


Kayden-ML:
Implement state_transition() function 

https://ethereum.github.io/consensus-specs/specs/phase0/beacon-chain/#beacon-chain-state-transition-function

Kayden-ML:
Working on this^

Kayden-ML:
Implemented in this PR https://github.com/ReamLabs/ream/pull/142

PR #142, opened 2025-02-25
feat: implement process_block(), state_transition() functions

Kayden-ML:
state_transition() Issue: https://github.com/ReamLabs/ream/issues/88
https://ethereum.github.io/consensus-specs/specs/phase0/beacon-chain/#beacon-chain-state-transition-function

Will add test in follow up PR https://github.com/ReamLabs/ream/issues/141
process_block() Issue: https://github.com/ReamLabs/ream/issues/127
https://ethereum.github.io/consensus-specs/specs/capella/beacon-chain/#block-processing

### DIFF
--- a/crates/common/consensus/src/deneb/beacon_state.rs
+++ b/crates/common/consensus/src/deneb/beacon_state.rs
@@ -1484,23 +1484,23 @@ impl BeaconState {
                 )
         );
 
-        for proposer_slashing in body.proposer_slashings.clone() {
-            self.process_proposer_slashing(&proposer_slashing)?;
+        for proposer_slashing in body.proposer_slashings.iter() {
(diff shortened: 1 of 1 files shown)


crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -1484,32 +1484,32 @@ impl BeaconState {
                 )
         );
 
-        for proposer_slashing in body.proposer_slashings {
+        for proposer_slashing in body.proposer_slashings.clone() {
```
KolbyML, line 1487: it there a way to avoid all these clones try an iter()

KolbyML (approved):
:shipit: looks good

PR #144, opened 2025-02-26
fix: `is_valid_indexed_attestation` and enable bls tests for operation tests

KolbyML:
The todo was removed in the last PR, but basically we weren't running operation tests which used bls, I fixed the bug so now the test suite works with BLS tests enabled


syjn99 (approved):
Thanks for fixing!

PR #149, opened 2025-02-28
fix: remove serde for VariableList/FixedVector from BeaconState

KolbyML:
@syjn99 said these don't work when he tries to use them in consensplus, I ran ef-tests and they passed without them


syjn99:
cc. @unnawut

syjn99 (approved):

unnawut:
Great thanks!

PR #151, opened 2025-02-28
feat: add blocking state transition function for PoC snarkification

KolbyML:
@unnawut said it would be ideal if we didn't use async functions when Proof of concepting the snarkification of the state transition function

### DIFF
--- a/crates/common/consensus/src/deneb/beacon_state.rs
+++ b/crates/common/consensus/src/deneb/beacon_state.rs
@@ -1790,6 +1790,99 @@ impl BeaconState {
         }
         Ok(())
     }
+
+    // START: This is for prototyping snarkifying the state transition function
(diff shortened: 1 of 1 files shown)


crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -1790,6 +1790,99 @@ impl BeaconState {
         }
         Ok(())
     }
+
+    // START: This is for prototyping snarkifying the state transition function
+    // TODO: remove this after we are done with the prototyping
```
unnawut, line 1795: Actually if we're expecting to remove this later, I'm happy to work from a branch and not having to pollute master?

KolbyML: Sure I will close the PR

PR #153, opened 2025-03-01
feat: initiating database for ream 

hopinheimer:
fancy some durability?

- [ ] I'm still looking into the use of `parking_lot::RwLock` [context](https://www.snoyman.com/blog/2024/01/best-worst-deadlock-rust/) 
- [ ] implement, key delete, key exists and other required functionalities 
- [ ] I've chosen to use CT-polymorphism to work around the overhead costs of RT, I may be wrong 
- [ ] refactoring the file structure

please free to edit/add other implementation requirements in the checklist, I'll try to include all of them in this PR


KolbyML:
@hopinheimer we are not using parking_lot::RwLock for the database as it is horrible https://github.com/ethereum/trin/pull/1700

KolbyML:
@hopinheimer if the PR is WIP it should be set to draft mode

hopinheimer:
yeah I suspected the issue. that's pretty comprehensive PR on trin thanks for the ref

crates/storage/src/lib.rs
```diff
@@ -1,14 +1,91 @@
...
+    Builder,
+    Database,
+    Durability,
+    TableDefinition
+};
+mod config;
+
+
+pub(crate) struct ReamDB {
+    db: RwLock<Database>,
```
KolbyML, line 13: We don't want to wrap a RwLock or Mutex around the database, maybe we should eventually have a queued write system but `RwLock` should probably not be here

crates/storage/src/lib.rs
KolbyML, line 1: I know this is a WIP, but we shouldn't use expect (I assume you already know this) For error's we won't use result here, instead use https://docs.rs/thiserror/latest/thiserror/

KolbyML, line 1: I know this is a WIP, but we shouldn't use expect (I assume you already know this) For error's we won't use result here, instead use https://docs.rs/thiserror/latest/thiserror/

KolbyML:
For getting the data directory path for Ream we will use the https://crates.io/crates/directories crate here is an example from Trin https://github.com/ethereum/trin/blob/master/crates/utils/src/dir.rs

crates/storage/src/errors.rs
```diff
@@ -0,0 +1,19 @@
...
+    TransactionError(#[from] redb::TransactionError),
+
+    #[error("Commit error")]
+    CommitError(#[from] redb::CommitError),
+
+    #[error("Storage error")]
+    StorageError(#[from] redb::StorageError),
+
+    #[error("Table error")]
+    TableError(#[from] redb::TableError),
```
KolbyML, line 18:
```suggestion
    #[error("Write transaction failure: {0}")]
    Database(#[from] redb::Error),

    #[error("Transaction error: {0}")]
    TransactionError(#[from] redb::TransactionError),

    #[error("Commit error: {0}")]
    CommitError(#[from] redb::CommitError),

    #[error("Storage error: {0}")]
    StorageError(#[from] redb::StorageError),

    #[error("Table error: {0}")]
    TableError(#[from] redb::TableError),
```

we want to display the underlying error

crates/storage/src/redb_interface.rs
```diff
@@ -0,0 +1,69 @@
+use anyhow::Result;
+use parking_lot::Mutex;
+use redb::{Builder, Database, Durability, ReadableTable, TableDefinition};
+
+use crate::{config, errors::StoreError};
+
+pub(crate) struct ReamDB {
+    db: Mutex<Database>,
```
KolbyML, line 8:
```suggestion
    db: Database,
```

no mutex

crates/storage/src/redb_interface.rs
```diff
@@ -0,0 +1,69 @@
...
+        let mut write_txn = self.db.lock().begin_write()?;
+        write_txn.set_durability(Durability::Immediate);
+        let mut table = write_txn.open_table(table_def)?;
+        table.insert(key, value)?;
+        drop(table);
+        write_txn.commit()?;
+        Ok(())
+    }
+
+    fn get_bytes(&self, name: &str, key: &[u8]) -> Result<Option<Vec<u8>>, StoreError> {
```
KolbyML, line 41: We don't need the _bytes suffix

KolbyML:
The PR isn't passing CI

crates/utils/src/dir.rs
KolbyML, line 1: Can we put this code into the storage crate

crates/storage/src/redb_interface.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, StoreError> {
+        let table_def: TableDefinition<'_, &[u8], &[u8]> = TableDefinition::new(&self.table);
+        let read_txn = self.backend.db.begin_read()?;
+        let table = read_txn.open_table(table_def)?;
+        let result = table.get(key)?;
+        Ok(result.map(|res| res.value().to_vec()))
+    }
+}
```
KolbyML, line 38: This code shouldn't be inbetween the struct and impl of ReamDB

crates/storage/src/redb_interface.rs
KolbyML, line 1: It feels a little weird that none of these functions for the most part are public when they should be

crates/storage/src/db.rs
```diff
@@ -0,0 +1,28 @@
+use crate::redb_interface::ReamDB;
+use crate::errors::StoreError;
+
+pub struct Database{
+    /// Database for the ream client 
+    /// 
+    /// Currently this implementation has a single reference to redb instance
+    /// but eventually we would like to add support for multiple backends and 
+    /// can be set behind feature flags
```
KolbyML, line 9: I don't think we should support multiple backends. For BLS fine that was fairly minimally, but for properly using a db it is a little more involved then a db expessially to get performance or to ensure crashes don't corrupt the database.

Cargo.toml
```diff
@@ -46,6 +48,8 @@ jsonwebtoken = "9.3.1"
 libp2p-identity = "0.2"
 libp2p-mplex = "0.42"
 libp2p = { version = "0.54", default-features = false, features = ["identify", "yamux", "noise", "dns", "tcp", "tokio", "plaintext", "secp256k1", "macros", "ecdsa", "metrics", "quic", "upnp", "gossipsub", "ping"] }
+parking_lot = { version = "0.12.3" }
```
KolbyML, line 51: This crate is no longer being used it should be removed from cargo.toml

crates/storage/src/config.rs
```diff
@@ -0,0 +1,2 @@
+pub const REDB_FILE: &str = "ream.durability.redb";
```
KolbyML, line 1:
```suggestion
pub const REDB_FILE: &str = "ream.redb";
```

Why do we need the word durability

crates/storage/src/db.rs
```diff
@@ -0,0 +1,28 @@
...
+    ///
+    /// `redb::Database` wrapper around with async security features
+    db: ReamDB,
```
KolbyML, line 12:
```suggestion
    pub db: ReamDB,
```

without pub this is unusable

crates/utils/src/dir.rs
```diff
@@ -0,0 +1,23 @@
...
+use directories::BaseDirs;
+
+/// Creates a `.ream` directory in the system's data directory if it doesn't exist.
+///
+/// # Returns
+///
+/// Returns the path to the `.ream` directory on success, or an `io::Error` if it fails.
+pub fn create_ream_dir() -> io::Result<PathBuf> {
+    if let Some(base_dirs) = BaseDirs::new() {
+        let ream_dir = base_dirs.data_dir().join(".ream");
```
KolbyML, line 12:
```suggestion
/// Creates a `ream` directory in the system's data directory if it doesn't exist.
///
/// # Returns
///
/// Returns the path to the `.ream` directory on success, or an `io::Error` if it fails.
pub fn create_ream_dir() -> io::Result<PathBuf> {
    if let Some(base_dirs) = BaseDirs::new() {
        let ream_dir = base_dirs.data_dir().join("ream");
```

Why are we trying to make the ream directory hidden? It should be a public folder

crates/storage/src/redb_interface.rs
```diff
@@ -0,0 +1,89 @@
...
+use redb::{Builder, Database, Durability, TableDefinition};
+
+use crate::{config, errors::StoreError};
+
+pub struct ReamDB {
+    db: Arc<Database>,
+}
+
+struct Connection<'a> {
+    backend: &'a ReamDB,
```
KolbyML, line 14: couldn't we just use an Arc<> here instead of passing the 'a lifetime

crates/storage/src/redb_interface.rs
```diff
@@ -0,0 +1,89 @@
...
+        Ok(())
+    }
+
+    fn acquire_connection<'a>(&'a mut self, table: String) -> Result<Connection<'a>, StoreError> {
+        Ok(Connection {
+            backend: self,
+            table,
+        })
+    }
+}
```
KolbyML, line 69:
```suggestion
    fn acquire_connection<'a>(&self, table: String) -> Result<Connection<'a>, StoreError> {
        Ok(Connection {
            backend: self.clone(),
            table,
        })
    }
}
```

If we are using an Arc around redb, we should just make the ReamDB struct clone-able

KolbyML (changes requested):
Overall the changes are fine, if we want to have robustness we are likely to expose a lot more then put and get, but for now that is fine.

I don't think we should attempt to support multiple databases, it is a developer resource drain, especially when to fully utilize a database and get the ACID benefits we would be using lower level calls then our `put` and `get` abstractions directly

crates/storage/src/redb_interface.rs
```diff
@@ -15,6 +15,7 @@ struct Connection<'a> {
     table: String,
 }
 
+#[allow(clippy::needless_lifetimes)] //TODO: (hopinheimer) - remove after further implementation
```
KolbyML, line 18: just remove the needless lifetimes

crates/storage/src/redb_interface.rs
```diff
@@ -0,0 +1,90 @@
...
+        let mut txn = self.db.begin_write()?;
+        txn.set_durability(Durability::Immediate);
+        txn.open_table(table)?;
+        txn.commit()?;
+        Ok(())
+    }
+
+    fn acquire_connection(&mut self, table: String) -> Result<Connection, StoreError> {
+        Ok(Connection {
+            backend: Arc::new(self.clone()),
```
KolbyML, line 39:
```suggestion
            backend: Arc::new(self),
```

we shouldn't need both new and clone here

bin/ream/src/main.rs
```diff
@@ -37,6 +38,9 @@ async fn main() {
         total_peers: 0,
     };
 
+    let ream_db = Database::new().unwrap();
```
KolbyML, line 41: no using unwrap

crates/utils/src/dir.rs
```diff
@@ -0,0 +1,23 @@
+use std::{fs, io, path::PathBuf};
+
+use directories::BaseDirs;
+
+/// Creates a `.ream` directory in the system's data directory if it doesn't exist.
```
KolbyML, line 5:
```suggestion
/// Creates a `ream` directory in the system's data directory if it doesn't exist.
```

this comment still has `.ream`, I would prefer if we didn't create this create and just had this code in the storage crate, but if you really want to keep it, the utils crate should go in the common folder, not in the base crates directory

crates/storage/src/config.rs
```diff
@@ -0,0 +1,2 @@
+pub const REDB_FILE: &str = "ream.redb";
+pub const REDB_CACHE_SIZE: usize = 1_024;
```
KolbyML, line 2: what unit is this const in? could you add a doc comment. Is this supposed to be mb, gb, kb etc

KolbyML (changes requested):
a bit more feedback

crates/storage/src/redb_interface.rs
```diff
@@ -0,0 +1,90 @@
...
+            table,
+        })
+    }
+}
+
+struct Connection {
```
jihoonsong, line 45: This file is named as interface but actually it's an impl. The original intention would've made sense if we have an interface for multiple backends and impls of each of them.

But I don't think we want to support multiple backend. I'm of the opinion that it won't be our priority for the time being.

crates/storage/src/redb_interface.rs
```diff
@@ -0,0 +1,90 @@
...
+        let table: TableDefinition<'_, &[u8], &[u8]> = TableDefinition::new(name);
+        let mut txn = self.db.begin_write()?;
+        txn.set_durability(Durability::Immediate);
+        txn.open_table(table)?;
+        txn.commit()?;
+        Ok(())
+    }
+
+    fn acquire_connection(&mut self, table: String) -> Result<Connection, StoreError> {
+        Ok(Connection {
```
jihoonsong, line 38: I think it should verify if the table exists and return error when it's not.

crates/storage/src/redb_interface.rs
```diff
@@ -0,0 +1,90 @@
...
+            backend: Arc::new(self.clone()),
+            table,
+        })
+    }
+}
+
+struct Connection {
+    backend: Arc<ReamDB>,
```
jihoonsong, line 46: This is self-referential. It should be Arc of abstraction of multiple backends. Also it's not accessible anyway as it's not public.

crates/storage/src/redb_interface.rs
```diff
@@ -0,0 +1,90 @@
...
+    table: String,
+}
+
+impl Connection {
+    fn put(&self, key: &[u8], value: &[u8]) -> Result<(), StoreError> {
+        let table_def: TableDefinition<'_, &[u8], &[u8]> = TableDefinition::new(&self.table);
```
jihoonsong, line 52: Do we really need to create this every time?

crates/storage/src/db.rs
```diff
@@ -0,0 +1,25 @@
...
+
+/// This implemenation houses all the client<>db interaction logic
+impl Database {
+    pub fn new() -> Result<Self, StoreError> {
+        let db_instance = ReamDB::new()?;
+        Ok(Self {
+            db: db_instance,
+            version: String::from("0.0.1"),
+        })
+    }
```
jihoonsong, line 24:
```suggestion
        Ok(Self {
            db: ReamDB::new()?,
            version: String::from("0.0.1"),
        })
    }
```
I'd prefer to use intermediate variables for both or not.

crates/storage/src/db.rs
jihoonsong: This is not resolved. Make sure you mark comments resolved only when it is.

jihoonsong (commented):
It seems it tried to support multiple backends but it's abstraction is not ready for it. I'm not sure if we want that support so I'd suggest to make it focus on only redb for now.

hopinheimer:
Screenshot 2025-03-05 at 2 08 40 PM

this describes the design I have in mind. I'm open to suggestions here or on tg

bin/ream/src/main.rs
```diff
@@ -21,9 +22,9 @@ async fn main() {
 
     let cli = Cli::parse();
 
-    let async_executor = ReamExecutor::new().unwrap();
+    let async_executor = ReamExecutor::new().expect("unable to create executor");
```
jihoonsong, line 25: Could you also change L53 as well?

Cargo.toml
```diff
@@ -9,7 +9,7 @@ members = [
     "crates/networking/p2p",
     "crates/rpc",
     "crates/runtime",
-    "crates/storage",
+    "crates/storage",
```
jihoonsong, line 12:
```suggestion
    "crates/storage",
```

crates/storage/src/redb_impl.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn create_table(&self, name: &str) -> Result<(), StoreError> {
+        let table: TableDefinition<'_, &[u8], &[u8]> = TableDefinition::new(name);
+        let mut txn = self.db.begin_write()?;
+        txn.set_durability(Durability::Immediate);
+        txn.open_table(table)?;
+        txn.commit()?;
+        Ok(())
+    }
+
+    fn acquire_connection(&self, table: String) -> Result<Connection, StoreError> {
```
jihoonsong, line 36: This has got to be public. Same with other functions.

jihoonsong (changes requested):
Please check if the storage crate can be used in other places.

https://doc.rust-lang.org/reference/visibility-and-privacy.html

crates/storage/src/config.rs
```diff
@@ -0,0 +1,2 @@
+pub const REDB_FILE: &str = "ream.redb";
+pub const REDB_CACHE_SIZE: usize = 1_024 * 1_024 * 1_024; // 1 GiB
```
KolbyML, line 2:
```suggestion

/// 1 GiB
pub const REDB_CACHE_SIZE: usize = 1_024 * 1_024 * 1_024;
```

can you use a doc comment

crates/storage/src/db.rs
```diff
@@ -0,0 +1,24 @@
+use crate::{errors::StoreError, redb_impl::ReamDB};
+
+pub struct Database {
+    /// Database for the ream client
+    ///
+    /// Currently this implementation has a single reference to redb instance
+    /// but eventually we would like to add support for multiple backends and
+    /// can be set behind feature flags
+    ///
```
KolbyML, line 9:
```suggestion
```

Can we remove this, as we don't plan to support multiple backends

crates/storage/src/dir.rs
```diff
@@ -0,0 +1,23 @@
...
+        if !ream_dir.exists() {
+            fs::create_dir_all(&ream_dir)?;
+        }
+        Ok(ream_dir)
+    } else {
+        Err(io::Error::new(
+            io::ErrorKind::NotFound,
+            "Base directories not found",
+        ))
+    }
```
KolbyML, line 22:
```suggestion
    let Some(base_dirs) = BaseDirs::new() else {
        Err(io::Error::new(
            io::ErrorKind::NotFound,
            "Base directories not found",
        ))
    }
    
        let ream_dir = base_dirs.data_dir().join("ream");
        if !ream_dir.exists() {
            fs::create_dir_all(&ream_dir)?;
        }
        Ok(ream_dir)    
```

we don't need this if/else indentation

crates/storage/src/lib.rs
```diff
@@ -1,14 +1,6 @@
...
-mod tests {
-    use super::*;
-
-    #[test]
-    fn it_works() {
-        let result = add(2, 2);
-        assert_eq!(result, 4);
-    }
-}
+#![allow(dead_code)]
```
KolbyML, line 1:
```suggestion
```

can we not add deadcode, if a function is public it isn't dead,

crates/storage/src/redb_impl.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        let db = Builder::new()
+            .set_cache_size(config::REDB_CACHE_SIZE)
+            .create(&ream_file)
+            .map_err(|err| StoreError::Database(err.into()))?;
+
+        Ok(Self { db: Arc::new(db) })
+    }
+
+    fn create_table(&self, name: &str) -> Result<(), StoreError> {
```
KolbyML, line 27:
```suggestion
    pub fn create_table(&self, name: &str) -> Result<(), StoreError> {
```

crates/storage/src/redb_impl.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    fn acquire_connection(&self, table: String) -> Result<Connection, StoreError> {
+        Ok(Connection {
+            backend: self.db.clone(),
+            table,
+        })
+    }
+}
+
+struct Connection {
```
KolbyML, line 44:
```suggestion
pub struct Connection {
```

crates/storage/src/redb_impl.rs
```diff
@@ -0,0 +1,89 @@
...
+    backend: Arc<Database>,
+    table: String,
+}
+
+impl Connection {
+    fn put(&self, key: &[u8], value: &[u8]) -> Result<(), StoreError> {
```
KolbyML, line 50:
```suggestion
    pub fn put(&self, key: &[u8], value: &[u8]) -> Result<(), StoreError> {
```

crates/storage/src/redb_impl.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        write_txn.set_durability(Durability::Immediate);
+        let mut table = write_txn.open_table(table_def)?;
+        table.insert(key, value)?;
+        drop(table);
+        write_txn.commit()?;
+        Ok(())
+    }
+
+    fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, StoreError> {
```
KolbyML, line 62:
```suggestion
   pub fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, StoreError> {
```

KolbyML (commented):
Here are some issues to resolve

hopinheimer:
switching ownership to @KolbyML for this PR

KolbyML:
Closing this as I am going to make a new PR

PR #165, opened 2025-03-16
feat: Implement final testing for process_registry_updates()

Kayden-ML:
Fixed the process_registry_updates() function to match the spec

Issue: https://github.com/ReamLabs/ream/issues/164

### DIFF
--- a/crates/common/consensus/src/deneb/beacon_state.rs
+++ b/crates/common/consensus/src/deneb/beacon_state.rs
@@ -401,21 +401,14 @@ impl BeaconState {
     }
 
     /// Initiate if validator already initiated exit.
-    pub fn initiate_validator_exit(&mut self, index: u64) {
+    pub fn initiate_validator_exit(&mut self, index: u64) -> anyhow::Result<()> {
+        // Return if validator already initiated exit
         if index as usize >= self.validators.len() {
-            return;
-        }
-        if self
-            .validators
-            .get(index as usize)
-            .expect("Can't find index in validators")
-            .exit_epoch
-            != FAR_FUTURE_EPOCH
-        {
-            return;
+            return Ok(());
         }
 
-        let mut exit_epochs: Vec<u64> = self
+        // Compute exit queue epoch
+        let exit_epochs: Vec<u64> = self
             .validators
             .iter()
             .filter_map(|v| {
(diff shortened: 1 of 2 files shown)


crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -440,12 +437,23 @@ impl BeaconState {
...
 
-        // Set validator exit epoch and withdrawable epoch
-        if let Some(validator) = self.validators.get_mut(index as usize) {
-            validator.exit_epoch = exit_queue_epoch;
-            validator.withdrawable_epoch =
-                validator.exit_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY;
+        let validator = self
+            .validators
+            .get_mut(index as usize)
+            .expect("could not get validator");
```
KolbyML, line 443:
```suggestion
            .expect("could not get validator");
```

We shouldn't use expect and instead use an anyhow error

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -1569,32 +1576,36 @@ impl BeaconState {
...
+        for &index in activation_queue
             .iter()
             .take(self.get_validator_activation_churn_limit() as usize)
         {
-            // Ensure no overflow when computing the activation epoch
-            let new_activation_epoch = compute_activation_exit_epoch(current_epoch);
-            let activation_epoch = new_activation_epoch
-                .checked_add(1)
-                .ok_or_else(|| anyhow::anyhow!("Overflow while calculating activation epoch"))?;
-            self.validators[*index].activation_epoch = activation_epoch;
```
KolbyML, line 1597: this should be added back

crates/common/consensus/src/deneb/beacon_state.rs
```diff
@@ -440,12 +437,23 @@ impl BeaconState {
...
-            validator.exit_epoch = exit_queue_epoch;
-            validator.withdrawable_epoch =
-                validator.exit_epoch + MIN_VALIDATOR_WITHDRAWABILITY_DELAY;
+        let validator = self
+            .validators
+            .get_mut(index as usize)
+            .expect("could not get validator");
+
+        if validator.exit_epoch != FAR_FUTURE_EPOCH {
+            return Ok(());
```
KolbyML, line 446: couldn't this if statement be moved up the function to match the pseduo code or is there a problem with doing that?

KolbyML (commented):
some feedback

crates/common/consensus/src/deneb/beacon_state.rs
Kayden-ML: The reason I did it like this is because self could not pass in a mutable reference twice

KolbyML (approved):
:shipit: looks good

PR #167, opened 2025-03-18
feat: Implemented macro for sanity_slots and added testing for process_slots()

Kayden-ML:
Implemented macro for sanity_slots and added testing for process_slots()

Spec: https://github.com/ethereum/consensus-specs/blob/8fa1aaae5eaca6d9bdcea99658e834963557c20d/tests/formats/sanity/slots.md


KolbyML (approved):
:shipit: assuming CI passes looks good

PR #174, opened 2025-03-25
chore: fix panic message formatting

mdqst:
Updated panic messages to use the correct Rust syntax for variable insertion (`{}`) instead of `{variable}`, ensuring the variables are properly displayed in error messages.


KolbyML (changes requested):
If we want to fix the formatting macros, then the PR should be thorough

KolbyML:
I am closing the PR, as this PR doesn't change any behavior this is a style change. And it is reversing the style we intentionally want and is set to do in our cargo fmt config, but it doesn't work in macro's for some reason

PR #249, opened 2025-03-26
feat: initialize db

KolbyML:
@Vid201 ping for review

I am initialing the storage crate, I also added a small demo for showing how we would create a table

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ anyhow = "1.0"
 async-trait = "0.1.86"
 kzg = { git = "https://github.com/grandinetech/rust-kzg" }
 clap = "4"
+directories = { version = "6.0.0" } 
 discv5 = { version = "0.9.0", features = ["libp2p"] }
 enr = "0.13.0"
 ethereum_hashing = { git = "https://github.com/ReamLabs/ethereum_hashing.git" }
@@ -48,6 +49,7 @@ jsonwebtoken = "9.3.1"
 libp2p-identity = "0.2"
 libp2p-mplex = "0.42"
 libp2p = { version = "0.54", default-features = false, features = ["identify", "yamux", "noise", "dns", "tcp", "tokio", "plaintext", "secp256k1", "macros", "ecdsa", "metrics", "quic", "upnp", "gossipsub", "ping"] }
+redb = { version = "2.4.0" }
 reqwest = { version = "0.12", features = ["json"] }
 rstest = "0.23"
 serde = { version = '1.0', features = ['derive', "rc"] }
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -48,13 +49,15 @@ jsonwebtoken = "9.3.1"
 libp2p-identity = "0.2"
 libp2p-mplex = "0.42"
 libp2p = { version = "0.54", default-features = false, features = ["identify", "yamux", "noise", "dns", "tcp", "tokio", "plaintext", "secp256k1", "macros", "ecdsa", "metrics", "quic", "upnp", "gossipsub", "ping"] }
+redb = { version = "2.4.0" }
```
Vid201, line 52: redb = "2.4.0"

Cargo.toml
```diff
@@ -71,6 +74,7 @@ ream-execution-engine = { path = "crates/common/execution_engine"}
 ream-executor = { path = "crates/common/executor" }
 ream-network-spec = { path = "crates/common/network_spec" }
 ream-p2p = { path = "crates/networking/p2p" }
+ream-storage = { path = "crates/storage"}
```
Vid201, line 77: { path = "crates/storage" }

missing whitespace

bin/ream/src/cli/mod.rs
```diff
@@ -32,6 +32,19 @@ pub struct NodeConfig {
         value_parser = network_parser
     )]
     pub network: Arc<NetworkSpec>,
+
+    #[arg(
+        long,
+        help = "The directory for storing application data. If used together with --ephemeral, new child directory will be created."
+    )]
+    pub data_dir: Option<PathBuf>,
```
Vid201, line 40: Would a default value be useful here? Like `~/.ream` ?

KolbyML: We don't need to set a default here as `setup_data_dir() will use the default application data directory  for the specific OS

crates/storage/src/tables/beacon_state.rs
```diff
@@ -0,0 +1,42 @@
...
+/// Key: Slot number
+/// Value: BeaconState
+pub const BEACON_STATE_TABLE: TableDefinition<u64, SSZEncoding<BeaconState>> =
+    TableDefinition::new("beacon_state");
+
+pub struct BeaconStateTable {
+    pub db: Arc<Database>,
+}
+
+impl Table for BeaconStateTable {
```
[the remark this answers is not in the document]
KolbyML: Maybe in the future, I think for the time being a macro would be a bit overkill. We still need to fully see how we would want to use the DB, this implementation isn't the final one I assume, but it is good enough to start.

Kayden-ML (approved):
Just went over this looks good 👍

PR #258, opened 2025-03-31
feat: add `/node/version` endpoint

### LINKED ISSUE #179: Implement /eth/v1/node/version
https://ethereum.github.io/beacon-APIs/#/Node/getNodeVersion

varun-doshi:
Closes #179 

Run node:
```
cargo run node
```

Make a `GET` request to:
```
/eth/v1/node/version
```

### DIFF
--- a/crates/rpc/src/handlers/version.rs
+++ b/crates/rpc/src/handlers/version.rs
@@ -1,0 +1,27 @@
+use ream_node::version::ream_node_version;
+use serde::{Deserialize, Serialize};
+use warp::{
+    http::status::StatusCode,
+    reject::Rejection,
+    reply::{Reply, with_status},
+};
+
+use super::Data;
+
+#[derive(Serialize, Deserialize, Default)]
+pub struct Version {
+    version: String,
+}
+
+impl Version {
+    pub fn new() -> Self {
+        Self {
+            version: ream_node_version(),
+        }
+    }
+}
+
+/// Called by `/version` to get the Node Version.
+pub async fn get_version() -> Result<impl Reply, Rejection> {
+    Ok(with_status(Data::json(Version::new()), StatusCode::OK))
+}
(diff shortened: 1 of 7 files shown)


crates/rpc/src/handlers/version.rs
```diff
@@ -0,0 +1,18 @@
...
+
+/// Called by `/version` to get the Node Version.
+pub async fn get_version() -> Result<impl Reply, Rejection> {
+    let version = format!("{}{}{}", APPLICATION_NAME, "/v", env!("CARGO_PKG_VERSION"));
+
+    Ok(with_status(
+        json(&json!({"data":{"version":version}})),
+        StatusCode::OK,
+    ))
+}
```
KolbyML, line 18: Hmm I don't like this way of getting the version string, I will make a PR tomorrow for generating it, then you can rebase your PR and use my code :+1:

KolbyML (changes requested):

crates/rpc/src/handlers/version.rs
KolbyML: I will just make a PR for it, as it is a little involved

crates/rpc/src/handlers/version.rs
```diff
@@ -0,0 +1,33 @@
...
+            version: ream_node_version(),
+        }
+    }
+}
+
+impl Default for Version {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
KolbyML, line 28:
```suggestion
#[derive(Serialize, Default)]
pub struct Version {
    version: String,
}

impl Version {
    pub fn new() -> Self {
        Self {
            version: ream_node_version(),
        }
    }
}
```

If you want Default you can derive it

KolbyML (approved):
:shipit: looks good you still have one thing to fix

PR #260, opened 2025-03-31
feat: add ream version info

KolbyML:
@varun-doshi once this is merged rebase your PR and do

in your route.rs file for the node rpc

```

pub struct Version {
    version: String
}

impl Version {
    pub fn new() -> Self {
         Self { version: ream_node_version() }
    }
}
```

Then you can do `Data::json(Version::new())`


varun-doshi (approved):
lgtm

PR #262, opened 2025-03-31
feat: add bootnodes cli param

KolbyML:
@Vid201 ping for review

@Vid201 wanted a way for the user to specify which bootnodes to connect to through the CLI. Initially he tried in this PR https://github.com/ReamLabs/ream/pull/253 to  extend the pre-existing list, but I would prefer fleshed out approach if we are going to add a bootnodes cli. I message him and he said it is fine if I make a PR for this


Vid201 (approved):
LGTM!

PR #263, opened 2025-04-01
chore: add cargo-machete, remove unused deps

Vid201:
Good tool to remove unused deps so it doesn't have to be checked manually.


KolbyML (changes requested):
I don't like this approach as it often has a lot of problems 

```
[package.metadata.cargo-machete]
ignored = ["vergen"]
```

I think we could instead add a CI which runs udeps for example https://github.com/sigp/discv5/pull/179

And I wouldn't want a command like this to be in `make lint`, I would want the CI to moreso act like an alert that the PR maintainer can then further look into themselves

Vid201:
> I don't like this approach as it often has a lot of problems like false reports
> 
> ```
> [package.metadata.cargo-machete]
> ignored = ["vergen"]
> ```
> 
> ^ I think this is kind of unidle, that we would need to manually ignore mistakes it makes.
> 
> I think we could instead add a CI which runs udeps for example [sigp/discv5#179](https://github.com/sigp/discv5/pull/179)
> 
> And I wouldn't want a command like this to be in `make lint`, I would want the CI to moreso act like an alert that the PR maintainer can then further look into themselves

Okay, it makes sense to have it only in CI.

Lately, I have preferred cargo-machete due to it being much faster, and I've also had problems with cargo-udeps where I had to manually ignore some dependencies. But I'm fine with cargo-udeps as well.

KolbyML:
> > I don't like this approach as it often has a lot of problems like false reports
> > ```
> > [package.metadata.cargo-machete]
> > ignored = ["vergen"]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ^ I think this is kind of unidle, that we would need to manually ignore mistakes it makes.
> > I think we could instead add a CI which runs udeps for example [sigp/discv5#179](https://github.com/sigp/discv5/pull/179)
> > And I wouldn't want a command like this to be in `make lint`, I would want the CI to moreso act like an alert that the PR maintainer can then further look into themselves
> 
> Okay, it makes sense to have it only in CI.
> 
> Lately, I have preferred cargo-machete due to it being much faster, and I've also had problems with cargo-udeps where I had to manually ignore some dependencies. But I'm fine with cargo-udeps as well.

As long as we don't have an `ignore section` in our Cargo.tomls I would be fine if you wanted to try `cargo-machete` in CI, assuming you can make it less strict so we don't need to add `ignore sections`

KolbyML:
I would also be fine with a solution where we didn't have a CI, but had a make command to run it just, not under lint

Vid201:
I just tried `cargo +nightly udeps` as well, but it doesn't find unused deps well (misses most of them). 

With machete, you can also specify ignore deps in the workspace Cargo.toml instead of having to specify in each package Cargo.toml separately.

I guess there are 3 options:
1. use udeps and miss some unused deps
2. use machete but we need `ignored` array in workspace Cargo.toml
3. leave this for now

KolbyML:
I am fine adding a page to the Ream Book, and an option called `make clean-deps`, which does a check and doesn't change the files automatically. Then a human can use the tool. I would only be fine with a CI if it was correct or under reported, if it overreported false positives then I would have a problem.

I am pretty against having `ignored array in workspace Cargo.toml`.

KolbyML (approved):
:shipit: looks good

PR #264, opened 2025-04-01
ci: add CI which runs cargo sort --check

KolbyML:
Adding a CI which makes sure our `Cargo.toml` files are sorted, we are using `cargo-sort` for this which was added to our makefile and Ream book recently. Since it is included in `make lint` we should have a ci for it


Kayden-ML (approved):
LGTM 👍

PR #269, opened 2025-04-03
feat: implement `/node/identity` endpoint

### LINKED ISSUE #177: Implement /eth/v1/node/identity
https://ethereum.github.io/beacon-APIs/#/Node/getNetworkIdentity

varun-doshi:
Closes #177 

To run, start a node:

```
cargo run node
```

Make `GET` request to:
```
/node/identity
```

Note: The `metadata` field gives dummy values now, this will be updated once we add `attnets` and other related networking fields

### DIFF
--- a/crates/common/network_spec/src/identity.rs
+++ b/crates/common/network_spec/src/identity.rs
@@ -1,0 +1,49 @@
+use enr::{CombinedKey, Enr};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Identity {
+    pub peer_id: String,
+    pub enr: String,
+    pub p2p_address: String,
+    pub discovery_address: String,
+    pub metadata: MetaData,
+}
+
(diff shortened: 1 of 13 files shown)


crates/common/network_spec/src/identity.rs
KolbyML, line 1: This doesn't belong in the network spec crate

KolbyML (changes requested):
I think this PR should be done when the pre-reqs are done, I am not a fan of shipping endpoints with dummy mock data.

KolbyML:
> Note: The `metadata` field gives dummy values now, this will be updated once we add `attnets` and other related networking fields

I marked the PR as a draft, as I don't think there is value in adding this endpoint until we can add it in a working state without mock data.

KolbyML:
Hi we recently migrated from using `warp` to `actix-web` https://github.com/ReamLabs/ream/pull/336 please update your PR accordingly

PR #270, opened 2025-04-03
feat: updated beacon state table to use block_root instead of slot number

Kayden-ML:
updated beacon state table to use block_root instead of slot number


KolbyML (approved):
:shipit: looks good

PR #271, opened 2025-04-03
feat: add beacon block table to the database

Kayden-ML:
added the beacon block table to the database


KolbyML (approved):
:shipit: looks good

PR #309, opened 2025-04-05
feat: implement /randao endpoint

### LINKED ISSUE #193: Implement /eth/v1/beacon/states/{state_id}/randao
https://ethereum.github.io/beacon-APIs/#/Beacon/getStateRandao

varun-doshi:
To run, start a node:
```
cargo run node
```
Make `GET` request to:
```
/states/<state_id>/randao
```
With optional epoch query:
```
/states/<state_id>/randao?epoch=1234
```

Closes #193

### DIFF
--- a/crates/rpc/src/handlers/randao.rs
+++ b/crates/rpc/src/handlers/randao.rs
@@ -1,4 +1,4 @@
-use alloy_primitives::FixedBytes;
+use alloy_primitives::B256;
 use ream_storage::db::ReamDB;
 use serde::{Deserialize, Serialize};
 use warp::{
@@ -8,11 +8,11 @@ use warp::{
 };
 
 use super::{BeaconResponse, state::get_state_from_id};
-use crate::types::id::ID;
+use crate::types::{id::ID, query::RandaoQuery};
 
 #[derive(Serialize, Deserialize)]
 struct RandaoResponse {
-    pub randao: FixedBytes<32>,
+    pub randao: B256,
 }
 
 /// Called by `/states/<state_id>/randao` to get the Randao mix of state.
@@ -20,11 +20,11 @@ struct RandaoResponse {
 /// else will fetch randao of the state epoch
 pub async fn get_randao_mix(
     state_id: ID,
-    epoch: Option<u64>,
+    query: RandaoQuery,
     db: ReamDB,
 ) -> Result<impl Reply, Rejection> {
     let state = get_state_from_id(state_id, &db).await?;
-    let randao_mix = match epoch {
+    let randao_mix = match query.epoch {
         Some(epoch) => state.get_randao_mix(epoch),
         None => state.get_randao_mix(state.get_current_epoch()),
     };
--- a/crates/rpc/src/handlers/randao.rs
+++ b/crates/rpc/src/handlers/randao.rs
@@ -1,0 +1,36 @@
+use alloy_primitives::FixedBytes;
+use ream_storage::db::ReamDB;
+use serde::{Deserialize, Serialize};
+use warp::{
+    http::status::StatusCode,
+    reject::Rejection,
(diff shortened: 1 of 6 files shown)


crates/rpc/src/handlers/randao.rs
```diff
@@ -0,0 +1,36 @@
...
+    reject::Rejection,
+    reply::{Reply, with_status},
+};
+
+use super::{BeaconResponse, state::get_state_from_id};
+use crate::types::id::ID;
+
+#[derive(Serialize, Deserialize)]
+struct RandaoResponse {
+    pub randao: FixedBytes<32>,
```
KolbyML, line 15:
```suggestion
    pub randao: B256,
```

crates/rpc/src/routes/beacon.rs
```diff
@@ -28,7 +34,23 @@ pub fn get_beacon_routes(
...
+        beacon_base
+            .and(path("states"))
+            .and(param::<ID>())
+            .and(path("randao"))
+            .and(query::<RandaoQuery>())
+            .and(end())
+            .and(get())
+            .and_then(move |state_id: ID, query: RandaoQuery| {
+                let epoch = query.epoch;
+                get_randao_mix(state_id, epoch, db.clone())
```
KolbyML, line 48:
```suggestion
        let db = db.clone();
        beacon_base
            .and(path("states"))
            .and(param::<ID>())
            .and(path("randao"))
            .and(query::<RandaoQuery>())
            .and(end())
            .and(get())
            .and_then(move |state_id: ID, query: RandaoQuery| {
                get_randao_mix(state_id, query, db)
```
we don't need the clone here because it is cloned above

also just pass `query` in

KolbyML (commented):
here is some feedback

crates/rpc/src/routes/beacon.rs
varun-doshi: So i did try removing the clone inside but it doesnt work, it says i need to clone both inside and outisde. i think its an issue with moving into the closure the FnOnce vs Fn

making a workaround using filters

crates/rpc/src/routes/beacon.rs
```diff
@@ -4,22 +4,35 @@ use ream_network_spec::networks::NetworkSpec;
...
+        id::{ID, ValidatorID},
+        query::RandaoQuery,
+    },
 };
 
+fn with_db(
+    db: ReamDB,
+) -> impl Filter<Extract = (ReamDB,), Error = std::convert::Infallible> + Clone {
+    warp::any().map(move || db.clone())
+}
```
KolbyML, line 27: this filter should probably be in mod.rs or a file called filters

crates/rpc/src/routes/beacon.rs
```diff
@@ -28,6 +41,21 @@ pub fn get_beacon_routes(
...
+            .and(path("randao"))
+            .and(query::<RandaoQuery>())
+            .and(end())
+            .and(get())
+            .and(db_filter.clone())
+            .and_then(move |state_id: ID, query: RandaoQuery, db: ReamDB| {
+                get_randao_mix(state_id, query, db)
+            })
+            .with(log("randao"))
+    };
```
KolbyML, line 57:
```suggestion
    let randao =
        beacon_base
            .and(path("states"))
            .and(param::<ID>())
            .and(path("randao"))
            .and(query::<RandaoQuery>())
            .and(end())
            .and(get())
            .and(db_filter.clone())
            .and_then(move |state_id: ID, query: RandaoQuery, db: ReamDB| {
                get_randao_mix(state_id, query, db)
            })
            .with(log("randao"));
```

now that we don't do `let db = db.clone()` I don't think there is a reason to keep these brackets for scoping it

KolbyML (approved):
:shipit: looks good

PR #310, opened 2025-04-06
feat: implement /eth/v2/debug/beacon/states/{state_id}

### LINKED ISSUE #176: Implement /eth/v2/debug/beacon/states/{state_id}
https://ethereum.github.io/beacon-APIs/#/Debug/getStateV2

aditya172926:
wip closes #176 

This PR introduces
- Allow `state_id` to be passed as strings `head`, `justified`, `finalized`, `genesis`
- Creates a new base url with `eth/v2/...`
- Return BeaconState from the api endpoint implemented

### DIFF
--- a/crates/rpc/src/handlers/state.rs
+++ b/crates/rpc/src/handlers/state.rs
@@ -1,7 +1,14 @@
 use ream_consensus::deneb::beacon_state::BeaconState;
 use ream_storage::{db::ReamDB, tables::Table};
+use warp::{
+    http::status::StatusCode, 
+    reject::Rejection, 
+    reply::{with_status, Reply}
+};
 
 use crate::types::{errors::ApiError, id::ID};
+
+use super::BeaconResponse;
 
 pub async fn get_state_from_id(state_id: ID, db: &ReamDB) -> Result<BeaconState, ApiError> {
     let block_root = match state_id {
@@ -19,4 +26,10 @@ pub async fn get_state_from_id(state_id: ID, db: &ReamDB) -> Result<BeaconState,
         .ok_or(ApiError::NotFound(format!(
             "Failed to find `beacon_state` from {block_root:?}"
         )))
+}
+
+pub async fn get_state(state_id: ID, db: ReamDB) -> Result<impl Reply, Rejection> {
+    let state = get_state_from_id(state_id, &db).await?;
+
+    Ok(with_status(BeaconResponse::json(state), StatusCode::OK))
 }
--- a/crates/rpc/src/handlers/state.rs
+++ b/crates/rpc/src/handlers/state.rs
@@ -1,5 +1,6 @@
+use alloy_primitives::FixedBytes;
 use ream_consensus::deneb::beacon_state::BeaconState;
-use ream_storage::{db::ReamDB, tables::Table};
+use ream_storage::{db::ReamDB, tables::{ Field, Table}};
 use warp::{
     http::status::StatusCode, 
     reject::Rejection, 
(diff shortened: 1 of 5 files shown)


crates/rpc/src/types/id.rs
```diff
@@ -8,6 +8,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum ID {
+    Named(String),
```
KolbyML, line 11:
```suggestion
    Head,
    Justified,
    Finalized,
    Genesis,
```

Instead of named it should be like this, except sorted in alphabetical order with the rest of the enums options

crates/rpc/src/handlers/state.rs
```diff
@@ -1,10 +1,35 @@
...
+};
 
 use crate::types::{errors::ApiError, id::ID};
 
+use super::BeaconResponse;
+
 pub async fn get_state_from_id(state_id: ID, db: &ReamDB) -> Result<BeaconState, ApiError> {
     let block_root = match state_id {
+        ID::Named(ref name) => match name.as_str() {
+            "head" => {Ok(Some(FixedBytes::new([0; 32])))},
```
aditya172926, line 17: Hey @KolbyML could you give some input here.
For `state_id` I am adding the variants `head`, `justified`, `finalized`, `genesis` to have this endpoint return the corresponding BeaconState.

I've just added some placeholders for the ones I am not sure how to get yet.
Here how should I be getting the `head` block_root?

crates/rpc/src/handlers/state.rs
```diff
@@ -1,10 +1,35 @@
...
+            },
+            "finalized" => {
+                let finalized_checkpoint = db.finalized_checkpoint_provider().get().map_err(|_| ApiError::InternalError)?
+                .ok_or_else(|| ApiError::NotFound(String::from("Finalized checkpoint not found")))?;
+                
+                Ok(Some(finalized_checkpoint.root))
+            },
+            "genesis" => db.slot_index_provider().get(0),
+            &_ => Ok(Some(FixedBytes::new([0; 32])))
+        },
```
KolbyML, line 32:
```suggestion
        ID::Head | ID::Genesis => return Err(ApiError::NotFound(format!(
        "This ID type is currently not supported: {state_id:?}"
    ))),
    ID:Justified => {
          let justified_checkpoint = db.justified_checkpoint_provider().get().map_err(|_| ApiError::InternalError)?
          .ok_or_else(|| ApiError::NotFound(String::from("Justified checkpoint not found")))?;
          
          Ok(Some(justified_checkpoint.root))
      },
    ID:Finalized => {
            let finalized_checkpoint = db.finalized_checkpoint_provider().get().map_err(|_| ApiError::InternalError)?
            .ok_or_else(|| ApiError::NotFound(String::from("Finalized checkpoint not found")))?;
            
            Ok(Some(finalized_checkpoint.root))
      },
```

I am not sure if we will end up supporting the genesis state, or we don't do currently so we should just throw an error

crates/rpc/src/routes/mod.rs
```diff
@@ -3,28 +3,35 @@ use std::sync::Arc;
...
     let node_routes = get_node_routes();
 
     let config_routes = get_config_routes(network_spec.clone());
 
-    eth_base.and(beacon_routes.or(node_routes).or(config_routes))
+    let debug_routes = get_debug_routes(db.clone());
+
+    let combined_routes = beacon_routes.or(node_routes).or(config_routes).or(debug_routes);
+    let versioned_base = eth_base_v1.or(eth_base_v2).and_then(|_| async { Ok::<_, Rejection>(()) });
+    versioned_base.and(combined_routes).map(|_, reply| reply)
```
KolbyML, line 34: v1 routes should only reply to v1 same with v2, wouldn't this current approach let you use either version for every RPC? which is unidle, maybe make a `debug_routes_v2` function, and consider ones without `_v1` v1

crates/rpc/src/handlers/state.rs
```diff
@@ -20,3 +45,9 @@ pub async fn get_state_from_id(state_id: ID, db: &ReamDB) -> Result<BeaconState,
             "Failed to find `beacon_state` from {block_root:?}"
         )))
 }
+
+pub async fn get_state(state_id: ID, db: ReamDB) -> Result<impl Reply, Rejection> {
+    let state = get_state_from_id(state_id, &db).await?;
+
+    Ok(with_status(BeaconResponse::json(state), StatusCode::OK))
+}
```
KolbyML, line 53:
```suggestion
}

```

missing new line 
image

crates/rpc/src/handlers/state.rs
KolbyML: > Here how should I be getting the head block_root?

I think it is fine if we return an error for now

KolbyML (commented):
Here is some quick feedback

crates/rpc/src/routes/mod.rs
aditya172926: Yeah, will fix this.

KolbyML (approved):
:shipit: looks good! Thank you for your contribution!

PR #311, opened 2025-04-06
feat: implement /<block>/root endpoint

### LINKED ISSUE #201: Implement /eth/v1/beacon/blocks/{block_id}/root
https://ethereum.github.io/beacon-APIs/#/Beacon/getBlockRoot

varun-doshi:
To run, start a node:
```
cargo run node
```
Make `GET` request to:
```
/blocks/<block_id>/root
```

Closes #201

### DIFF
--- a/crates/rpc/src/handlers/block.rs
+++ b/crates/rpc/src/handlers/block.rs
@@ -1,0 +1,64 @@
+use alloy_primitives::B256;
+use ream_storage::{
+    db::ReamDB,
+    tables::{Field, Table},
+};
+use serde::{Deserialize, Serialize};
+use warp::{
+    http::status::StatusCode,
+    reject::Rejection,
+    reply::{Reply, with_status},
+};
+
+use super::BeaconResponse;
+use crate::types::{errors::ApiError, id::ID};
+
+#[derive(Serialize, Deserialize)]
+struct BlockRootResponse {
+    pub root: B256,
+}
+
+/// Called by `/blocks/<block_id>/root` to get the Tree hash of the Block.
+pub async fn get_block_root(block_id: ID, db: ReamDB) -> Result<impl Reply, Rejection> {
+    let block_root = match block_id {
+        ID::Finalized => {
+            let finalized_checkpoint = db
+                .finalized_checkpoint_provider()
+                .get()
+                .map_err(|_| ApiError::InternalError)?
+                .ok_or_else(|| {
+                    ApiError::NotFound(String::from("Finalized checkpoint not found"))
+                })?;
+
+            Ok(Some(finalized_checkpoint.root))
+        }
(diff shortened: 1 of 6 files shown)


crates/rpc/src/handlers/block.rs
```diff
@@ -0,0 +1,48 @@
...
+        "Failed to find `block_root` from {block_id:?}"
+    )))?;
+
+    db.beacon_block_provider()
+        .get(block_id)
+        .map_err(|_| ApiError::InternalError)?
+        .ok_or(ApiError::NotFound(format!(
+            "Failed to find `beacon_state` from {block_id:?}"
+        )))
+}
```
KolbyML, line 48: we don't need to fetch the block from the database, because the indexes give it to us no? So fetching the block and recomputing the tree_hash_root is a little redundent in this example

crates/rpc/src/handlers/block.rs
```diff
@@ -0,0 +1,64 @@
...
+    reply::{Reply, with_status},
+};
+
+use super::BeaconResponse;
+use crate::types::{errors::ApiError, id::ID};
+
+#[derive(Serialize, Deserialize)]
+struct BlockRootResponse {
+    pub root: B256,
+}
```
KolbyML, line 19:
```suggestion
#[derive(Serialize, Deserialize)]
struct RootResponse {
    pub root: B256,
}
```

Could we just call this `RootResponse` as we are likely to use it in more then one place, so maybe put this in a different file as we will use it for the data as well

crates/rpc/src/handlers/block.rs
```diff
@@ -0,0 +1,64 @@
...
+            ))
+            .into());
+        }
+        ID::Slot(slot) => db.slot_index_provider().get(slot),
+        ID::Root(root) => Ok(Some(root)),
+    }
+    .map_err(|_| ApiError::InternalError)?
+    .ok_or(ApiError::NotFound(format!(
+        "Failed to find `block_root` from {block_id:?}"
+    )))?;
```
KolbyML, line 58: can this be in a function called `fn get_block_root`, as we can reuse it in the `get_block` function

crates/rpc/src/handlers/block.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    Ok(with_status(
+        BeaconResponse::json(BlockRootResponse { root: block_root }),
```
KolbyML, line 61:
```suggestion
        BeaconResponse::json(RootResponse::new(block_root)),
```

I think it would be cleaner to just use a constructor

KolbyML (commented):
here is some feedback

KolbyML (approved):
:shipit: looks good

PR #318, opened 2025-04-07
fix: make sure fork-choice functions follow the spec correctly and avoid unnecessary clones

jihoonsong:
This PR corrects some of fork-choice functions to follow the spec properly and uses reference to avoid unnecessary clones.


KolbyML (approved):
:shipit: looks good

PR #325, opened 2025-04-10
fix: api returns 500 internal error

varun-doshi:
Fixes an issue where the endpoints return `500 Internal Error` instead of `400 Not Found`

The issue: None of the tables were actually ever created since we never insert data(yet).
This fix will make it so that even if we dont insert data, it will atleast show StatusCode:400 instead of 500 which is a better output

### DIFF
--- a/crates/rpc/src/routes/beacon.rs
+++ b/crates/rpc/src/routes/beacon.rs
@@ -4,10 +4,7 @@ use ream_network_spec::networks::NetworkSpec;
 use ream_storage::db::ReamDB;
 use warp::{
     Filter, Rejection,
-    filters::{
-        path::{end, param},
-        query::query,
-    },
+    filters::{path::end, query::query},
     get, log, path,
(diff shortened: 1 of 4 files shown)


KolbyML (changes requested):
We are not handling 

```
	
The block ID supplied could not be parsed
{
  "code": 400,
  "message": "Invalid block ID: current"
}
```
get_block_root_from_id

and

```
Invalid state ID

{
  "code": 400,
  "message": "Invalid state ID: current"
}
```

in get_state_from_id

So currently the error `400` should be returned, but right now `404` is being returned

The fix in this PR makes this PR return `404`, not `400`

KolbyML (approved):
:shipit: looks good, I am assuming you tested this locally yourself

varun-doshi:
> :shipit: looks good, I am assuming you tested this locally yourself

I did, yes

PR #327, opened 2025-04-11
feat: store SignedBeaconBlock instead of BeaconBlock into the database

KolbyML:
We should store the `SignedBeaconBlock` instead as it is needed for
- the BeaconAPI
- if peers requests BeaconBlocks from us over P2P
- Required to backfill from a weak subjectivity checkpoint


varun-doshi (approved):
Lgtm

PR #328, opened 2025-04-11
implement /states/{state_id}/validator_balances route

Arch0125:
## Description
This PR implements the Beacon API for /states/{state_id}/validator_balances, supporting both GET and POST requests
- For GET, the optional validator id[] is passed as a query param
- For POST, the optional body id[] is passed as a json 

Current the max ID request for validator is capped as 1000, exceeding that request it returns error code `414`, although the max number can be discussed upon

## Running
- `cargo run node`

- Send a cURL request to `/states/{state_id}/validator_balances`


crates/rpc/src/handlers/validator.rs
```diff
@@ -35,6 +36,12 @@ impl ValidatorData {
     }
 }
 
+#[derive(Debug, Serialize)]
+struct ValidatorBalance {
+    index: String,
+    balance: String,
```
KolbyML, line 42: these should be u64, look at other structs to see how you would add the serialization to a string

crates/rpc/src/handlers/validator.rs
```diff
@@ -103,3 +110,54 @@ pub async fn validator_status(validator: &Validator, db: &ReamDB) -> Result<Stri
...
+        .collect();
+
+    let response = BeaconResponse {
+        execution_optimistic: false,
+        finalized: false,
+        data: validator_balances,
+    };
+
+    Ok(warp::reply::with_status(
+        warp::reply::json(&response),
```
KolbyML, line 160:
```suggestion
    Ok(with_status(
        warp::reply::json(BeaconResponse::new(validator_balances)),
```

crates/rpc/src/handlers/validator.rs
```diff
@@ -103,3 +110,54 @@ pub async fn validator_status(validator: &Validator, db: &ReamDB) -> Result<Stri
...
+        StatusCode::OK,
+    ))
+}
```
KolbyML, line 163: image

you are missing a new line

crates/rpc/src/types/query.rs
```diff
@@ -4,3 +4,8 @@ use serde::Deserialize;
 pub struct RandaoQuery {
     pub epoch: Option<u64>,
 }
+
+#[derive(Debug, Deserialize)]
+pub struct ValidatorBalanceQuery {
+    pub id: Option<Vec<String>>,
```
KolbyML, line 10: A enum should be used instead of a string here

crates/rpc/src/types/query.rs
```diff
@@ -4,3 +4,8 @@ use serde::Deserialize;
...
+}
```
KolbyML, line 11: image

KolbyML (commented):
I left some initial feedback

crates/rpc/src/handlers/validator.rs
```diff
@@ -116,48 +118,48 @@ pub async fn get_validator_balances_from_state(
...
         .filter_map(|(i, validator)| {
-            if let Some(ref ids) = filter {
-                if !ids.contains(&i.to_string()) && !ids.contains(&format!("{:?}", validator.pubkey)) {
+            if let Some(ref addresses) = filter {
+                if !addresses.contains(&validator.pubkey) {
                     return None;
                 }
             }
-            let balance = state.balances.get(i).unwrap_or(&0);
+            let balance = state.validators.get(i).map(|v| v.effective_balance).unwrap();
```
Arch0125, line 153: @KolbyML Let me know if this is the correct way to fetch validator balances, previously it was using `states.balances`

crates/rpc/src/types/query.rs
Arch0125: Fixed in the latest comment with `ValidatorID` type

crates/rpc/src/handlers/validator.rs
Arch0125: Fixed in the latest comment with `u64` as in other structs

KolbyML (commented):
I forgot to hit the submit review button

crates/rpc/src/handlers/validator.rs
```diff
@@ -103,3 +112,54 @@ pub async fn validator_status(validator: &Validator, db: &ReamDB) -> Result<Stri
...
+        .validators
+        .iter()
+        .enumerate()
+        .filter_map(|(i, validator)| {
+            if let Some(ref addresses) = filter {
+                if !addresses.contains(&validator.pubkey) {
+                    return None;
+                }
+            }
+            let balance = state.validators.get(i).map(|v| v.effective_balance).expect("Validator balance not found");
```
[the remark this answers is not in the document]
Arch0125: fixed in the latest commit

crates/rpc/src/handlers/validator.rs
```diff
@@ -192,3 +203,60 @@ pub async fn post_validators_from_state(
...
 }
+pub async fn get_validator_balances_from_state(
+    state_id: ID,
+    query: ValidatorBalanceQuery,
+    db: ReamDB,
+) -> Result<impl Reply, Rejection> {
+    let state = get_state_from_id(state_id, &db).await?;
+
+    if let Some(ref ids) = query.id {
+        if ids.len() > 1000 {
```
KolbyML, line 214: can 1000 be a const at the top of the file below imports

crates/rpc/src/handlers/validator.rs
```diff
@@ -192,3 +203,60 @@ pub async fn post_validators_from_state(
...
+                .iter()
+                .filter_map(|v| match v {
+                    ValidatorID::Address(addr) => Some(addr.clone()),
+                    _ => None,
+                })
+                .collect();
+            Some(addresses)
+        }
+        None => None,
+    };
```
KolbyML, line 232:
```suggestion
    let filter: Option<HashSet<PubKey>> = match &query.id {
        Some(ids) if !ids.is_empty() => {
            let addresses: HashSet<PubKey> = ids
                .iter()
                .filter_map(|v| match v {
                    ValidatorID::Address(addr) => Some(addr.clone()),
                    _ => None,
                })
                .collect();
            Some(addresses)
        }
        _ => None,
    };
```
This can be simplified to this

KolbyML, line 232: aren't we missing the `validator index` case?

KolbyML (commented):
left some feedback

crates/rpc/src/handlers/validator.rs
Arch0125: yeah got it fixed, mis-read the api requirements from the reference before

KolbyML:
Hi we recently migrated from using `warp` to `actix-web` https://github.com/ReamLabs/ream/pull/336 please update your PR accordingly

Arch0125:
Oh cool, let me update the pr accordingly then

Arch0125:
@KolbyML can you review the changes , thanks!

KolbyML:
image
could you rebase, I will try to review the PR tomorrow assuming you rebase it

Arch0125:
@KolbyML ive rebased and fixed the conflicts, up for review

crates/rpc/src/handlers/validator.rs
```diff
@@ -30,6 +32,8 @@ pub struct ValidatorData {
     validator: Validator,
 }
 
+pub const MAX_REQUEST_LENGTH: usize = 1000;
```
KolbyML, line 35: We already have a const `MAX_VALIDATOR_COUNT` could we consolidate these and only use 1?

KolbyML (commented):
Left some feedback

crates/rpc/src/handlers/validator.rs
```diff
@@ -304,3 +316,120 @@ pub async fn post_validator_identities_from_state(
...
+        }
+
+        let balance = validator
+            .effective_balance;
+
+        validator_balances.push(ValidatorBalance {
+            index: i as u64,
+            balance,
+        });
+    }
```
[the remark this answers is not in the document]
Arch0125: The original code was designed to handle duplications, also goes over the whole state.validators only once, as compared to the one suggested, ill try and combine the optimizations of both of the implmentations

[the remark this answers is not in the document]
KolbyML: Convert id's to a set then first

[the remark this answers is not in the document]
KolbyML: Or maybe make id's a set to begin with

crates/rpc/src/handlers/validator.rs
```diff
@@ -42,6 +44,56 @@ impl ValidatorData {
...
+        if include && seen.insert(idx) {
+            balances.push(ValidatorBalance {
+                index: idx,
+                balance: validator.effective_balance,
+            });
+        }
+    }
+
+    balances
+}
```
KolbyML, line 95:
```suggestion
fn build_validator_balances(
    validators: &[Validator],
    filter_ids: Option<&Vec<ValidatorID>>,
) -> Vec<ValidatorBalance> {
    let Some(filter_ids) = filter_ids else {
        return vec![];
    };

    let indexes = filter_ids
        .into_iter()
        .filter_map(|validator_id| match validator_id {
            ValidatorID::Index(index) => Some(*index),
            ValidatorID::Address(public_key) => validators
                .iter()
                .enumerate()
                .find(|(_, validator)| &validator.pubkey == public_key)
                .map(|(index, _)| index as u64),
        })
        .collect::<HashSet<_>>();

    let mut balances = vec![];
    for index in indexes {
        let Some(validator) = validators.get(index as usize) else {
            continue;
        };
        balances.push(ValidatorBalance {
            index,
            balance: validator.effective_balance,
        });
    }

    balances
}
```

We can simplify this function quite a bit as seen above

KolbyML (commented):
I left some feedback, after the feedback is addressed this PR should be good or almost good to merge

crates/rpc/src/handlers/validator.rs
Arch0125: In this case it isn't actually handling the case when the query.id is None, that is supposed to return all validators if i'm not wrong, which isn't the case with the code proposed.
Also this implementation runs at time complexity of O(n*m) whereas the previous one is O(n+m).
Let me know your thoughts about these pointers

KolbyML: > @KolbyML In this case it isn't actually handling the case when the query.id is None, that is supposed to return all validators if i'm not wrong, which isn't the case with the code proposed. Also this implementation runs at time complexity of O(n*m) whereas the previous one is O(n+m). Let me know your thoughts about these pointers

The previous implementation isn't O(n+m), I am not sure where you are getting that from.

```diff
diff --git a/crates/rpc/src/handlers/validator.rs b/crates/rpc/src/handlers/validator.rs
index 3cc8a47..dc990da 100644
--- a/crates/rpc/src/handlers/validator.rs
+++ b/crates/rpc/src/handlers/validator.rs
@@ -1,10 +1,10 @@
+use std::collections::HashSet;
+
 use actix_web::{
     HttpResponse, Responder, get, post,
     web::{Data, Json, Path, Query},
 };
-use std::collections::HashSet;
 use ream_bls::PubKey;
-
 use ream_consensus::validator::Validator;
 use ream_storage::db::ReamDB;
 use serde::{Deserialize, Serialize};
@@ -53,45 +53,25 @@ struct ValidatorBalance {
 }
 
 fn build_validator_balances(
-    validators: &[Validator],
+    validators: &[(Validator, u64)],
     filter_ids: Option<&Vec<ValidatorID>>,
 ) -> Vec<ValidatorBalance> {
-    let mut allowed_indices: HashSet<u64> = HashSet::new();
-    let mut allowed_addresses: HashSet<PubKey> = HashSet::new();
-
-    if let Some(ids) = filter_ids {
-        for id in ids {
-            match id {
-                ValidatorID::Index(idx) => {
-                    allowed_indices.insert(*idx);
-                }
-                ValidatorID::Address(pk) => {
-                    if validators.iter().any(|v| &v.pubkey == pk) {
-                        allowed_addresses.insert(pk.clone());
-                    }
-                }
+    let filtered_ids = filter_ids.map(|filter_ids| filter_ids.iter().collect::<HashSet<_>>());
+    validators
+        .iter()
+        .enumerate()
+        .filter(|(index, (validator, _))| match &filtered_ids {
+            Some(ids) => {
+                ids.contains(&ValidatorID::Index(*index as u64))
+                    || ids.contains(&ValidatorID::Address(validator.pubkey.clone()))
             }
-        }
-    }
-
-    let mut seen: HashSet<u64> = HashSet::new();
-    let mut balances = Vec::new();
-
-    for (i, validator) in validators.iter().enumerate() {
-        let idx = i as u64;
-        let include = filter_ids.is_none()
-            || allowed_indices.contains(&idx)
-            || allowed_addresses.contains(&validator.pubkey);
-
-        if include && seen.insert(idx) {
-            balances.push(ValidatorBalance {
-                index: idx,
-                balance: validator.effective_balance,
-            });
-        }
-    }
-
-    balances
+            None => true,
+        })
+        .map(|(index, (_, balance))| ValidatorBalance {
+            index: index as u64,
+            balance: *balance,
+        })
+        .collect()
 }
 
 #[get("/beacon/states/{state_id}/validator/{validator_id}")]
@@ -373,11 +353,16 @@ pub async fn get_validator_balances_from_state(
         }
     }
 
-    let validator_balances = build_validator_balances(&state.validators, query.id.as_ref());
+    let validator_balances = build_validator_balances(
+        &state
+            .validators
+            .into_iter()
+            .zip(state.balances.into_iter())
+            .collect::<Vec<_>>(),
+        query.id.as_ref(),
+    );
 
-    Ok(HttpResponse::Ok().json(BeaconResponse::new(
-        validator_balances,
-    )))
+    Ok(HttpResponse::Ok().json(BeaconResponse::new(validator_balances)))
 }
 
 #[post("/beacon/states/{state_id}/validator_balances")]
@@ -395,9 +380,14 @@ pub async fn post_validator_balances_from_state(
         }
     }
 
-    let validator_balances = build_validator_balances(&state.validators, body.id.as_ref());
+    let validator_balances = build_validator_balances(
+        &state
+            .validators
+            .into_iter()
+            .zip(state.balances.into_iter())
+            .collect::<Vec<_>>(),
+        body.id.as_ref(),
+    );
 
-    Ok(HttpResponse::Ok().json(BeaconResponse::new(
-        validator_balances,
-    )))
+    Ok(HttpResponse::Ok().json(BeaconResponse::new(validator_balances)))
 }
```

^ here is code that should work then, I think you had an bug because I don't think balance meant effective balance

crates/rpc/src/handlers/validator.rs
```diff
@@ -305,3 +337,39 @@ pub async fn post_validator_identities_from_state(
...
+    let pairs: Vec<(Validator, u64)> = state
+        .validators
+        .into_iter()
+        .zip(state.balances.into_iter())
+        .collect();
+
+    let validator_balances = build_validator_balances(&pairs, body.id.as_ref());
+
+    Ok(HttpResponse::Ok().json(BeaconResponse::new(validator_balances)))
+}
```
KolbyML, line 375:
```suggestion
#[get("/beacon/states/{state_id}/validator_balances")]
pub async fn get_validator_balances_from_state(
    state_id: Path<ID>,
    query: Query<IdQuery>,
    db: Data<ReamDB>,
) -> Result<impl Responder, ApiError> {
    let state = get_state_from_id(state_id.into_inner(), &db).await?;
    Ok(HttpResponse::Ok().json(BeaconResponse::new(build_validator_balances(&state
        .validators
        .into_iter()
        .zip(state.balances.into_iter())
        .collect(), query.id.as_ref())))
}

#[post("/beacon/states/{state_id}/validator_balances")]
pub async fn post_validator_balances_from_state(
    state_id: Path<ID>,
    body: Json<IdQuery>,
    db: Data<ReamDB>,
) -> Result<impl Responder, ApiError> {
    let state = get_state_from_id(state_id.into_inner(), &db).await?;
    Ok(HttpResponse::Ok().json(BeaconResponse::new(build_validator_balances(&state
        .validators
        .into_iter()
        .zip(state.balances.into_iter())
        .collect(), body.id.as_ref()))))
}
```

Please just inline all of this

Arch0125:
Thanks for merging the PR!, will take care of the conventional commits from next time

PR #333, opened 2025-04-13
fix: preserve first error in loop execution

mdqst:
fixed an issue where `result` was being overwritten on each loop iteration.
this could cause the first error to be lost if a later iteration succeeded.
now, the first encountered error is preserved correctly.

### DIFF
--- a/testing/ef-tests/src/macros/sanity_blocks.rs
+++ b/testing/ef-tests/src/macros/sanity_blocks.rs
@@ -61,10 +61,12 @@ macro_rules! test_sanity_blocks {
                         let signed_block: SignedBeaconBlock = utils::read_ssz_snappy(&block_path)
                             .expect(&format!("cannot find test asset (blocks_{i}.ssz_snappy)"));
 
-                        result = state
-                            .state_transition(&signed_block, validate_result, &mock_engine)
-                            .await
-                            .map_err(|err| err.to_string());
+                        if result.is_ok() {
+                            result = state
+                                .state_transition(&signed_block, validate_result, &mock_engine)
+                                .await
+                                .map_err(|err| err.to_string());
+                        }
                     }
 
                     let expected_post =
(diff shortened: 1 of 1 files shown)


testing/ef-tests/src/macros/sanity_blocks.rs
```diff
@@ -61,10 +61,12 @@ macro_rules! test_sanity_blocks {
...
-                        result = state
-                            .state_transition(&signed_block, validate_result, &mock_engine)
-                            .await
-                            .map_err(|err| err.to_string());
+                        if result.is_ok() {
+                            result = state
+                                .state_transition(&signed_block, validate_result, &mock_engine)
+                                .await
+                                .map_err(|err| err.to_string());
+                        }
```
KolbyML, line 69:
```suggestion
                        
                            result = state
                                .state_transition(&signed_block, validate_result, &mock_engine)
                                .await
                                .map_err(|err| err.to_string());
                        if result.is_err() {
                            break:
                        }
```

Wouldn't it make more sense to do this? then to continue reading blocks from disk?

testing/ef-tests/src/macros/sanity_blocks.rs
```diff
@@ -89,7 +91,6 @@ macro_rules! test_sanity_blocks {
                             );
                         }
                         (Err(_), None) => {
-                            // Expected: invalid operations result in an error and no post state.
```
KolbyML, line 92: why is this log removed?

testing/ef-tests/src/macros/sanity_blocks.rs
KolbyML: with your code change, if state_transition returns an error

```
                        let block_path = case_dir.join(format!("blocks_{}.ssz_snappy", i));
                        if !block_path.exists() {
                            panic!("Test asset not found: {:?}", block_path);
                        }
                        let signed_block: SignedBeaconBlock = utils::read_ssz_snappy(&block_path)
                            .expect(&format!("cannot find test asset (blocks_{i}.ssz_snappy)"));
```

Only these lines will be ran, which isn't doing anything for the test

KolbyML:
I am closing this PR as the concerns weren't addressed, the contributor can reopen the PR if they come back and want to continue

PR #347, opened 2025-04-23
feat: update ethereum_hashing  with poseidon2 hash

SuccinctPaul:
This pr is used for updating ethereum_hashing  with poseidon2 hash.

We can select default hash by features.

* Reference: https://github.com/ReamLabs/ethereum_hashing/pull/1#issuecomment-2822982215

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -817,17 +817,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


unnawut:
Actually once the build passes this PR can be closed without merging as we won't be using poseidon2 right now in our main codebase

PR #349, opened 2025-04-25
chore: add pull request template

KolbyML:
I am adding a pull request template as some of the PR's made have poor descriptions. This pull request template solves 3 key pain points

- a lot of people don't add context on why they are writing a PR
- a lot of people don't add context on how their PR solves the problem
- then a todo section so people can read about conventional commits

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,4 +9,5 @@
 ### To-Do
 
  <!-- Stay ahead of things, add list items here!  -->
+- [ ] I have read [CONTRIBUTING.md](https://github.com/ReamLabs/ream/blob/master/CONTRIBUTING.md).
 - [ ] Uses [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,8 +1,12 @@
-### What was wrong?
+### What are you trying to achieve?
 
-### How was it fixed?
+ <!-- Describe in a sentence or two of the reason for this PR. Link to the issue if possible.  -->
+
+### How was it implemented/fixed?
+
+ <!-- List the approach you used, and/or changes made to the codebase  -->
 
 ### To-Do
 
-[//]: # (Stay ahead of things, add list items here!)
+ <!-- Stay ahead of things, add list items here!  -->
 - [ ] Uses [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).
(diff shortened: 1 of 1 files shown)


unnawut (commented):
Super nitpick: rephrase to cover features + description for each headline

KolbyML:
@unnawut ready for another look

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,8 @@
+### What was wrong?
+
+### How was it fixed?
+
+### To-Do
```
jihoonsong, line 9: The intention looks like a checklist but TODO sounds like there is leftover task after this PR.

unnawut (approved):

jihoonsong (commented):
I think more generalized template would be better. We could refer to https://github.com/OffchainLabs/prysm/blob/develop/.github/PULL_REQUEST_TEMPLATE.md?plain=1 for instance

.github/PULL_REQUEST_TEMPLATE.md
KolbyML: This is a good place for someone to say if there is something that is going to be done in follow up PR's.

jihoonsong: Makes sense. But then it's not the right place for conventional commits. It should have already done, not to be done in subsequent PRs.

jihoonsong: Having `conventional commits` makes it look like a checklist.

KolbyML: It is like a to-do list, where you have things to check off

KolbyML: > Having `conventional commits` makes it look like a checklist.

image

KolbyML: I don't think a `done list` is a thing

unnawut: I think this section should cover things that needs to be completed before merging the PR.

For future work in future PRs, the author should create new issues instead. The todo list in the PR will fade away over time.

unnawut: > I don't think a `done list` is a thing

True that. I don't think it's a done list is required to be in the template.

I still stand that todo should become issues though.

jihoonsong: Missing the point. The conventional commit is not a good example for to-do list as it should be already done in each PR. We won't have a following PR to make a convention commit that current RP is missing.

KolbyML: > Missing the point. The conventional commit is not a good example for to-do list as it should be already done in each PR. We won't have a following PR to make a convention commit that current RP is missing.

the `to-do` section doesn't only have to be used for followup PR's, it is a check list of things to do.

jihoonsong: > This is a good place for someone to say if there is something that is going to be done in follow up PR's.

So it's not for future PRs only. Then, I'd add an item to read contributing.md.

KolbyML: Sure

jihoonsong (approved):
Awesome work, it's good to see it's getting more structured.

PR #363, opened 2025-05-03
chore: bump libraries to latest major/minor version

KolbyML:

Just doing some dependency maintenance, I wanted to bump the Rust version as a feature I am trying to use stabilizes in 1.87, but we   need to wait for Risc0 to bump their version for the zkvm first 

bump all the libraries


Kayden-ML (approved):
looks good 👍

Issue #248, opened 2025-03-26
Implement Discv5 p2p discovery domain: `Sync committee bitfield`


KolbyML:
https://github.com/ethereum/consensus-specs/blob/e13c58d59da21d12f3e5bda8af8427afab1c0c1e/specs/altair/p2p-interface.md#sync-committee-bitfield

vineetpant:
Hi Kolby , I will start looking into P2P discovery domain issues, let me know if they have to be done in some order and feel free to assign issues to me.

KolbyML:
![Image](https://github.com/user-attachments/assets/bb65fab3-50de-4d16-ab82-e0e6990221c5)

@vineetpant #245 and #246 can be done in parallel

 #247 requires #245 and #246

#248 can probably be done whenever, but you could do it after the rest as well if you wanted

bomanaps:
Hello @KolbyML is this been worked on, if not can I get assigned thank you.

KolbyML:
> Hello [@KolbyML](https://github.com/KolbyML) is this been worked on, if not can I get assigned thank you.

Sure here is a link to the spec https://github.com/ethereum/consensus-specs/blob/e13c58d59da21d12f3e5bda8af8427afab1c0c1e/specs/altair/p2p-interface.md#sync-committee-bitfield

bomanaps:
Hello @KolbyML To properly implement the spec's requirement for validators to "select a random number of epochs before the end of the current sync committee period between 1 and SYNC_COMMITTEE_SUBNET_COUNT, inclusive", we would need to add the rand crate as a dependency, should I go ahead with this or am I missing something?

PR #364, opened 2025-05-03
feat: implement sync committee bitfield

bomanaps:

 This PR adds support for discovering peers subscribed to sync committee subnets by extending Ethereum Node Records (ENRs) with a syncnets field. This enables efficient peer discovery for sync committee duties via Discv5, aligning with [issue #248](https://github.com/ReamLabs/ream/issues/248).

 - Extended the ENR structure by introducing a `syncnets` key that stores a `Bitvector[SYNC_COMMITTEE_SUBNET_COUNT]`.
- Implemented logic where each bit in the bitfield represents subscription to a specific `sync_committee_{i}` topic.
- Integrated this field into the Discv5 discovery mechanism, allowing nodes to find peers based on syncnet participation.
- Ensured the implementation aligns with the latest Ethereum consensus discovery specs.

### To-Do

 

KolbyML:
I currently this the code won't work. Could you add a test encoding our enr, then decoding the 2 distinct subnet fields and seeing if that works.

bomanaps:
> I currently this the code will generate invalid Enr's. Could you add a test encoding our enr, then decoding the 2 distinct subnet fields and seeing if that works.

Image 04-05-2025 at 10 08

bomanaps:
Based on the image, the code is properly encoding and decoding subnet fields in ENRs without issues. 
It properly handles: Setting specific subnet bits, encoding those bits into the ENR, decoding the ENR to retrieve the original subnet information

bomanaps:
Please let me know if there’s any other approach or additional checks you’d like me to include to fully address your concern.

KolbyML:
```diff
diff --git a/crates/networking/discv5/src/subnet.rs b/crates/networking/discv5/src/subnet.rs
index ea55c39..9ba4771 100644
--- a/crates/networking/discv5/src/subnet.rs
+++ b/crates/networking/discv5/src/subnet.rs
@@ -508,6 +508,8 @@ mod tests {
             .build(&combined_key)
             .expect("Failed to build ENR");
 
+        panic!("ENR: {:?}", enr.to_base64());
+
         // Decode attestation bitfield
         let decoded_attnets = enr
             .get_decodable::<Subnets>(ATTESTATION_BITFIELD_ENR_KEY)
```
^ doing these generates 
```
enr:-Jq4QI1zb4rdPCyT7CDyreHJ4ps11y73VYX5h7pGGgv-NwuoLuVIzk_fsTB4JCN7eiu8yXjhuKNihAzVYdMgYrOTO80Bh2F0dG5ldHOJIgAAAAAAAAAFgmlkgnY0iXNlY3AyNTZrMaEDGO37MZVYMlh3-nqJqn3c211QDYE9NSvgYPDfGnBjCMmIc3luY25ldHOJIgAAAAAAAAAF
```
https://enr-viewer.com/

Put that enr into this veiwer

image


notice how attnets and syncnets both have the same value of `0x220000000000000005`

- attnets should be 8 bytes
- syncnets should be 1 byte

hence
- attnets should be `0x2200000000000000`
- syncnets should be `0x05`

Attnets and Syncnets should be their own types, not option's in Syncnets, `get_decodable` will return an option if the key isn't avaliable.

bomanaps:
> ```diff
> diff --git a/crates/networking/discv5/src/subnet.rs b/crates/networking/discv5/src/subnet.rs
> index ea55c39..9ba4771 100644
> --- a/crates/networking/discv5/src/subnet.rs
> +++ b/crates/networking/discv5/src/subnet.rs
> @@ -508,6 +508,8 @@ mod tests {
>              .build(&combined_key)
>              .expect("Failed to build ENR");
>  
> +        panic!("ENR: {:?}", enr.to_base64());
> +
>          // Decode attestation bitfield
>          let decoded_attnets = enr
>              .get_decodable::<Subnets>(ATTESTATION_BITFIELD_ENR_KEY)
> ```
> 
> ^ doing these generates
> 
> ```
> enr:-Jq4QI1zb4rdPCyT7CDyreHJ4ps11y73VYX5h7pGGgv-NwuoLuVIzk_fsTB4JCN7eiu8yXjhuKNihAzVYdMgYrOTO80Bh2F0dG5ldHOJIgAAAAAAAAAFgmlkgnY0iXNlY3AyNTZrMaEDGO37MZVYMlh3-nqJqn3c211QDYE9NSvgYPDfGnBjCMmIc3luY25ldHOJIgAAAAAAAAAF
> ```
> 
> https://enr-viewer.com/
> 
> Put that enr into this veiwer
> 
> image
> 
> notice how attnets and syncnets both have the same value of `0x220000000000000005`
> 
>     * attnets should be 8 bytes
> 
>     * syncnets should be 1 byte
> 
> 
> hence
> 
>     * attnets should be `0x2200000000000000`
> 
>     * syncnets should be `0x05`
> 
> 
> Attnets and Syncnets should be their own types, not option's in Syncnets, `get_decodable` will return an option if the key isn't avaliable.

Okay I see the issue form the way i handled  Encodable and Decodable, so i refactored it a little bit
Screenshot 2025-05-05 at 10 47 31

crates/networking/discv5/src/discovery.rs
```diff
@@ -80,6 +84,7 @@ impl Discovery {
         let enr = enr_builder
             .add_value(ENR_ETH2_KEY, &ENRForkID::pectra())
             .add_value(ATTESTATION_BITFIELD_ENR_KEY, &config.subnets)
+            .add_value(SYNC_COMMITTEE_BITFIELD_ENR_KEY, &config.subnets)
```
KolbyML, line 87: Is the ENR being encoded properly here? It looks like the same bug is happening here

KolbyML (commented):
I left some feedback, sorry for the delay

crates/networking/discv5/src/discovery.rs
```diff
@@ -129,21 +137,46 @@ impl Discovery {
...
+
+        self.start_query(query, target_peers);
+    }
+
+    pub fn discover_sync_committee_peers(&mut self, target_peers: usize, subnet_id: Option<u8>) {
+        // If the discv5 service isn't running or we are in the process of a query, don't bother
+        // queuing a new one.
+        if !self.started || self.find_peer_active {
+            return;
+        }
```
KolbyML, line 152: this function isn't used, could it be removed as we could just use start_query directly tbh

crates/networking/discv5/src/config.rs
```diff
@@ -11,16 +14,70 @@ pub struct DiscoveryConfig {
...
+    ) -> u64 {
+        let offset = self.calculate_sync_subnet_join_epoch_offset();
+
+        // If the next period is too close, we join immediately
+        if next_sync_committee_period_start_epoch <= current_epoch + offset {
+            current_epoch
+        } else {
+            next_sync_committee_period_start_epoch - offset
+        }
+    }
```
KolbyML, line 66: These functions were added, but none of them were used can they all be removed?

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,110 +1,452 @@
...
+    }
+}
+
+impl Encodable for SyncCommitteeSubnets {
+    fn encode(&self, out: &mut dyn BufMut) {
+        // Create a Bytes wrapper around the raw bytes to ensure proper RLP encoding
+        let bytes = Bytes::from(self.0.as_ssz_bytes());
+        bytes.encode(out);
+    }
+}
```
KolbyML, line 149:
```suggestion
impl Encodable for SyncCommitteeSubnets {
    fn encode(&self, out: &mut dyn BufMut) {
        let bytes = Bytes::from(self.0.as_ssz_bytes());
        bytes.encode(out);
    }
}
```

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,110 +1,452 @@
...
+        match bitvector {
+            Some(bits) => Ok(SyncCommitteeSubnets(bits)),
+            None => Err(alloy_rlp::Error::Custom(Box::leak(
+                "Failed to decode SyncCommitteeSubnets bitfield"
+                    .to_string()
+                    .into_boxed_str(),
+            ))),
+        }
+    }
+}
```
KolbyML, line 177:
```suggestion
impl Decodable for SyncCommitteeSubnets {
    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
      let bytes = Bytes::decode(buf)?;
        let subnets = BitVector::<U4>::from_ssz_bytes(&bytes).map_err(|err| {
            alloy_rlp::Error::Custom(Box::leak(
                format!("Failed to decode SSZ SyncCommitteeSubnets: {err:?}").into_boxed_str(),
            ))
        })?;
        Ok(subnets)
    }
}
```

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,110 +1,452 @@
...
+
+impl Encodable for AttestationSubnets {
+    fn encode(&self, out: &mut dyn BufMut) {
+        // Create a Bytes wrapper around the raw bytes to ensure proper RLP encoding
+        let bytes = Bytes::from(self.0.as_ssz_bytes());
+        bytes.encode(out);
+    }
+}
```
KolbyML, line 79:
```suggestion
impl Encodable for AttestationSubnets {
    fn encode(&self, out: &mut dyn BufMut) {
        let bytes = Bytes::from(self.0.as_ssz_bytes());
        bytes.encode(out);
    }
}
```

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,110 +1,452 @@
...
+        match bitvector {
+            Some(bits) => Ok(AttestationSubnets(bits)),
+            None => Err(alloy_rlp::Error::Custom(Box::leak(
+                "Failed to decode AttestationSubnets bitfield"
+                    .to_string()
+                    .into_boxed_str(),
+            ))),
+        }
+    }
+}
```
KolbyML, line 107:
```suggestion
impl Decodable for AttestationSubnets {
    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
      let bytes = Bytes::decode(buf)?;
        let subnets = BitVector::<U64>::from_ssz_bytes(&bytes).map_err(|err| {
            alloy_rlp::Error::Custom(Box::leak(
                format!("Failed to decode SSZ AttestationSubnets: {err:?}").into_boxed_str(),
            ))
        })?;
        Ok(subnets)
    }
}
```

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,110 +1,452 @@
...
+
+#[derive(Clone, Debug, Default)]
 pub struct Subnets {
-    attestation_bits: BitVector<U64>,
+    pub attestation_bits: Option<BitVector<U64>>,
+    pub sync_committee_bits: Option<BitVector<U4>>,
 }
```
KolbyML, line 183: Could you remove the whole `Subnets` struct as it is no longer used

crates/networking/discv5/src/subnet.rs
```diff
@@ -134,40 +476,436 @@ pub fn subnet_predicate(subnets: Vec<Subnet>) -> impl Fn(&Enr) -> bool + Send +
...
+
+/// Compute the subnets for a validator's current sync committee assignment
+///
+/// Returns a vector of subnet indices that the validator should be subscribed to
+pub fn compute_subnets_for_sync_committee(sync_committee_indices: &[usize]) -> Vec<u8> {
+    sync_committee_indices
+        .iter()
+        .map(|&index| compute_subnet_for_sync_committee_index(index))
+        .collect()
+}
```
KolbyML, line 573: these functions are only used in tests, could they be removed

crates/networking/discv5/src/discovery.rs
```diff
@@ -129,21 +137,46 @@ impl Discovery {
...
-            Some(id) => QueryType::FindSubnetPeers(vec![Subnet::Attestation(id)]),
-            None => QueryType::FindPeers,
+            Some(id) => QueryType::SubnetPeers(vec![Subnet::Attestation(id)]),
+            None => QueryType::Peers,
+        };
+
+        self.start_query(query, target_peers);
+    }
+
+    pub fn discover_sync_committee_peers(&mut self, target_peers: usize, subnet_id: Option<u8>) {
```
KolbyML, line 147: This function isn't used currently could it be removed for now and added back when it is used

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,110 +1,241 @@
...
+                Ok(())
+            }
+            _ => Err(anyhow!(
+                "Cannot disable non-sync committee subnet in SyncCommitteeSubnets"
+            )),
+        }
     }
 }
 
 pub fn subnet_predicate(subnets: Vec<Subnet>) -> impl Fn(&Enr) -> bool + Send + Sync {
```
KolbyML, line 202: could this be split into 2 different functions 1 for attestation subnets, and 1 for sync committees. As they should  be used in different cases no? And the current implementation has side affects which impacts each other.

bomanaps:
so this means  I will then need to add a functionality to actually use QueryType::SyncCommitteeSubnetPeers

KolbyML:
> so this means I will then need to add a functionality to actually use QueryType::SyncCommitteeSubnetPeers

QueryType::SyncCommitteeSubnetPeers is fine, I would just remove `discover_sync_committee_peers` function for now

bomanaps:
> > so this means I will then need to add a functionality to actually use QueryType::SyncCommitteeSubnetPeers
> 
> QueryType::SyncCommitteeSubnetPeers is fine, I would just remove `discover_sync_committee_peers` function for now

One last thing I want to keep subnet_predicate then add the 2 additional functions to maintain backward compatibility, I fear removing entirely will bring something badly?

KolbyML:
> > > so this means I will then need to add a functionality to actually use QueryType::SyncCommitteeSubnetPeers
> > 
> > 
> > QueryType::SyncCommitteeSubnetPeers is fine, I would just remove `discover_sync_committee_peers` function for now
> 
> One last thing I want to keep subnet_predicate then add the 2 additional functions to maintain backward compatibility, I fear removing it entirely will break something badly?

don't keep `subnet_predicate` just split it into 2 distinct functions. We don't need backwards compatibility

crates/networking/discv5/src/discovery.rs
```diff
@@ -142,9 +152,26 @@ impl Discovery {
...
+                            .collect();
+                        Box::new(attestation_subnet_predicate(subnet_ids))
+                    }
+                    QueryType::SyncCommitteeSubnetPeers(ref subnets) => {
+                        let subnet_ids: Vec<u8> = subnets
+                            .iter()
+                            .filter_map(|subnet| match subnet {
+                                Subnet::SyncCommittee(id) => Some(*id),
+                                _ => None,
+                            })
```
KolbyML, line 172: SubnetPeers should only have `Subnet::Attestation` and `QueryType::SyncCommitteeSubnetPeers` should only have `Subnet::SyncCommittee` could we remove the `Subnet enum` or don't use it like how we currently are as I don't think there is a reason to have this branching logic

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,468 @@
...
+/// let attestation_subnet = Subnet::Attestation(3);
+///
+/// // A sync committee subnet with index 1
+/// let sync_committee_subnet = Subnet::SyncCommittee(1);
+/// ```
 #[derive(Clone, Debug, PartialEq)]
 pub enum Subnet {
     Attestation(u8),
     SyncCommittee(u8),
 }
```
KolbyML, line 37: could you remove this enum, as couldn't all our code work without this?

KolbyML (commented):
I think we can remove the `Subnet` enum, as it is leading us to do a lot of redundent match cases for no reason, when the functions should only be accepting inputs for their respective committees

crates/networking/discv5/src/discovery.rs
bomanaps: refactor this:

So I can only see two possible ways to this either I  remove the Subnet enum and use direct u8 values, or
Keep the enum but structure the code to avoid the filter_map pattern, or what do you suggest?

crates/networking/discv5/src/subnet.rs
```diff
@@ -62,6 +38,20 @@ impl AttestationSubnets {
             .get(index)
             .map_err(|err| anyhow!("Subnet ID out of bounds: {err:?}"))
     }
+
+    pub fn enable_attestation_subnet(&mut self, subnet_id: u8) -> anyhow::Result<()> {
+        if subnet_id < 64 {
```
KolbyML, line 43: Could we use a const for 64 like we are doing for `SYNC_COMMITTEE_SUBNET_COUNT`

crates/networking/discv5/src/subnet.rs
```diff
@@ -112,6 +102,20 @@ impl SyncCommitteeSubnets {
             .get(index)
             .map_err(|err| anyhow!("Subnet ID out of bounds: {err:?}"))
     }
+
+    pub fn enable_sync_committee_subnet(&mut self, subnet_id: u8) -> anyhow::Result<()> {
+        if subnet_id < SYNC_COMMITTEE_SUBNET_COUNT as u8 {
+            self.set(subnet_id as usize, true)?;
+        }
+        Ok(())
```
KolbyML, line 110:
```suggestion
        ensure!(subnet_id < SYNC_COMMITTEE_SUBNET_COUNT as u8);
        self.set(subnet_id as usize, true)?;
        Ok(())
```

could you make other code like this match this pattern of using ensure instead of using an if state. because if something inputs a inappropriate id ideally we throw an error

crates/networking/discv5/src/subnet.rs
```diff
@@ -139,67 +143,9 @@ impl Decodable for SyncCommitteeSubnets {
...
-                Ok(())
-            }
-            _ => Err(anyhow!(
-                "Cannot disable non-sync committee subnet in SyncCommitteeSubnets"
-            )),
-        }
-    }
-}
+pub fn attestation_subnet_predicate(subnets: &[u8]) -> impl Fn(&Enr) -> bool + Send + Sync + use<> {
+    let subnets = subnets.to_vec();
```
KolbyML, line 147: do we need to convert this to a vec? why can't we use the slice directly?

crates/networking/discv5/src/subnet.rs
```diff
@@ -242,7 +188,11 @@ pub fn attestation_subnet_predicate(subnets: Vec<u8>) -> impl Fn(&Enr) -> bool +
     }
 }
 
-pub fn sync_committee_subnet_predicate(subnets: Vec<u8>) -> impl Fn(&Enr) -> bool + Send + Sync {
+pub fn sync_committee_subnet_predicate(
+    subnets: &[u8],
+) -> impl Fn(&Enr) -> bool + Send + Sync + use<> {
+    let subnets = subnets.to_vec();
```
KolbyML, line 194: do we need to convert this to a vec? why can't we use the slice directly?

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,414 @@
...
+        }
+        Ok(())
+    }
+
+    pub fn disable_attestation_subnet(&mut self, subnet_id: u8) -> anyhow::Result<()> {
+        if subnet_id < 64 {
+            self.set(subnet_id as usize, false)?;
+        }
+        Ok(())
+    }
```
KolbyML, line 54: I think we shouldn't have both `set`  and `enable_attestation_subnet`/`disable_attestation_subnet`

ideally we have only one or the other

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,414 @@
...
-                self.attestation_bits
-                    .set(id as usize, false)
-                    .map_err(|err| anyhow!("Subnet ID out of bounds: {err:?}"))?;
-                Ok(())
-            }
-            Subnet::Attestation(_) => Ok(()),
-            Subnet::SyncCommittee(_) => unimplemented!("SyncCommittee support not yet implemented"),
+    pub fn disable_sync_committee_subnet(&mut self, subnet_id: u8) -> anyhow::Result<()> {
+        if subnet_id < SYNC_COMMITTEE_SUBNET_COUNT as u8 {
+            self.set(subnet_id as usize, false)?;
```
KolbyML, line 115: same comment as above

KolbyML (commented):
Here is some feedback

crates/networking/discv5/src/subnet.rs
KolbyML: then the function should take a vector itself

crates/networking/discv5/src/discovery.rs
```diff
@@ -45,9 +48,12 @@ enum EventStream {
 }
 
 #[derive(Debug, Clone, PartialEq)]
+#[allow(clippy::enum_variant_names)]
+#[allow(dead_code)]
```
KolbyML, line 52:
```suggestion
```we shouldn't have these here

File: crates/networking/discv5/src/discovery.rs [hunk 21]
```diff
@@ -146,9 +156,12 @@ impl Discovery {
...
                 match query {
-                    QueryType::FindPeers => Box::new(empty_predicate()),
-                    QueryType::FindSubnetPeers(ref subnets) => {
-                        Box::new(subnet_predicate(subnets.clone()))
+                    QueryType::Peers => Box::new(empty_predicate()),
+                    QueryType::AttestationSubnetPeers(ref subnet_ids) => {
+                        Box::new(attestation_subnet_predicate(subnet_ids.clone()))
+                    }
+                    QueryType::SyncCommitteeSubnetPeers(ref subnet_ids) => {
+                        Box::new(sync_committee_subnet_predicate(subnet_ids.clone()))
```
KolbyML, line 164:
```
                match query.clone() {
                    QueryType::Peers => Box::new(empty_predicate()),
                    QueryType::AttestationSubnetPeers(subnet_ids) => {
                        Box::new(attestation_subnet_predicate(subnet_ids))
                    }
                    QueryType::SyncCommitteeSubnetPeers(subnet_ids) => {
                        Box::new(sync_committee_subnet_predicate(subnet_ids))
```

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,456 @@
...
+                format!("Failed to decode SSZ SyncCommitteeSubnets: {err:?}").into_boxed_str(),
             ))
         })?;
-        Ok(subnets)
+        Ok(Self(subnets))
     }
 }
 
-pub fn subnet_predicate(subnets: Vec<Subnet>) -> impl Fn(&Enr) -> bool + Send + Sync {
+pub fn attestation_subnet_predicate(subnets: Vec<u8>) -> impl Fn(&Enr) -> bool + Send + Sync {
```
KolbyML, line 180:
```suggestion
pub fn attestation_subnet_predicate(subnets: &[u8]) -> impl Fn(&Enr) -> bool + Send + Sync {
```

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,456 @@
...
+            match enr.get_decodable::<AttestationSubnets>(ATTESTATION_BITFIELD_ENR_KEY) {
+                Some(Ok(subnets)) => subnets,
+                _ => {
+                    trace!(
+                        "Peer rejected: invalid or missing attnets field; peer_id: {}",
+                        enr.node_id()
+                    );
+                    return false;
+                }
+            };
```
KolbyML, line 196:
```suggestion
        let attestation_bits =
            match enr.get_decodable::<AttestationSubnets>(ATTESTATION_BITFIELD_ENR_KEY) {
                Some(Ok(attestation_bits)) => attestation_bits,
                _ => {
                    trace!(
                        "Peer rejected: invalid or missing attnets field; peer_id: {}",
                        enr.node_id()
                    );
                    return false;
                }
            };
```

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,456 @@
...
+
+        for subnet_id in &subnets {
```
KolbyML, line 198:
```suggestion
        for subnet_id in subnets {
```
the reason it was erroring earlier was because we can't take a reference to a slice

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,456 @@
...
+                    enr.node_id()
+                );
+            }
+        }
+
+        false
+    }
+}
+
+pub fn sync_committee_subnet_predicate(subnets: Vec<u8>) -> impl Fn(&Enr) -> bool + Send + Sync {
```
KolbyML, line 223:
```suggestion
pub fn sync_committee_subnet_predicate(subnets: &[u8]) -> impl Fn(&Enr) -> bool + Send + Sync {
```

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,456 @@
...
+    move |enr: &Enr| {
+        if subnets.is_empty() {
+            return true;
+        }
+
+        let sync_committee_bits =
+            match enr.get_decodable::<SyncCommitteeSubnets>(SYNC_COMMITTEE_BITFIELD_ENR_KEY) {
+                Some(Ok(subnets)) => subnets,
```
KolbyML, line 231:
```suggestion
        let sync_committee_bits =
            match enr.get_decodable::<SyncCommitteeSubnets>(SYNC_COMMITTEE_BITFIELD_ENR_KEY) {
                Some(Ok(sync_committee_bits)) => sync_committee_bits,
```

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,456 @@
...
+                        "Peer rejected: missing syncnets field; peer_id: {}",
+                        enr.node_id()
+                    );
+                    return false;
                 }
+            };
+
+        for subnet_id in &subnets {
+            if *subnet_id >= SYNC_COMMITTEE_SUBNET_COUNT as u8 {
+                trace!(
```
KolbyML, line 243:
```suggestion
        for subnet_id in subnets {
            if *subnet_id >= SYNC_COMMITTEE_SUBNET_COUNT as u8 {
                trace!(
```

crates/networking/discv5/src/discovery.rs
```diff
@@ -181,19 +194,18 @@ impl Discovery {
...
-                QueryType::FindSubnetPeers(subnets) => {
+                QueryType::AttestationSubnetPeers(subnet_ids) => {
                     self.find_peer_active = false;
                     match query.result {
                         Ok(peers) => {
-                            let predicate = subnet_predicate(subnets.clone());
                             let filtered_peers = peers
                                 .into_iter()
-                                .filter(|enr| predicate(enr))
+                                .filter(|enr| attestation_subnet_predicate(subnet_ids.clone())(enr))
```
KolbyML, line 203:
```suggestion
                                .filter(|enr| attestation_subnet_predicate(enr))
```

crates/networking/discv5/src/discovery.rs
```diff
@@ -181,19 +194,18 @@ impl Discovery {
                         }
                     }
                 }
-                QueryType::FindSubnetPeers(subnets) => {
+                QueryType::AttestationSubnetPeers(subnet_ids) => {
                     self.find_peer_active = false;
                     match query.result {
                         Ok(peers) => {
-                            let predicate = subnet_predicate(subnets.clone());
```
KolbyML, line 188: we want to declare the predicate once like we did before

crates/networking/discv5/src/discovery.rs
```diff
@@ -207,6 +219,33 @@ impl Discovery {
...
                     }
                 }
+                QueryType::SyncCommitteeSubnetPeers(subnet_ids) => {
+                    self.find_peer_active = false;
+                    match query.result {
+                        Ok(peers) => {
+                            let filtered_peers = peers
+                                .into_iter()
+                                .filter(|enr| {
+                                    sync_committee_subnet_predicate(subnet_ids.clone())(enr)
```
KolbyML, line 229: we want to declare the predicate once like we did before

KolbyML (commented):
Here is some feedback

crates/networking/discv5/src/subnet.rs
bomanaps: I cant figure out a way to make the data live long enough for the closure, if i should go with this?

KolbyML: please implement the rest of my feedback I addressed this below

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,452 @@
...
+
+        let expected = BitVector::<U64>::from_bytes(vec![255; 8].into()).unwrap();
+        let attnets = enr
+            .get_decodable::<AttestationSubnets>(ATTESTATION_BITFIELD_ENR_KEY)
+            .unwrap()
+            .unwrap();
+
+        for i in 0..64 {
+            assert_eq!(attnets.get(i).unwrap(), expected.get(i).unwrap());
+        }
```
KolbyML, line 301: same feedback as above

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,452 @@
...
+
+        let decoded_attnets = enr
+            .get_decodable::<AttestationSubnets>(ATTESTATION_BITFIELD_ENR_KEY)
+            .expect("Failed to get attestation subnets")
+            .expect("Failed to decode attestation subnets");
+
+        assert!(decoded_attnets.get(3).unwrap());
+        assert!(decoded_attnets.get(42).unwrap());
+        assert!(!decoded_attnets.get(0).unwrap());
+        assert!(!decoded_attnets.get(10).unwrap());
```
KolbyML, line 379:
```suggestion
        assert!(decoded_attnets.get(3).unwrap());
        assert!(decoded_attnets.get(42).unwrap());
        assert!(!decoded_attnets.get(0).unwrap());
        assert!(!decoded_attnets.get(10).unwrap());
```

could we use assert_eq here and remove the unwrap

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,452 @@
...
+
+        let decoded_syncnets = enr
+            .get_decodable::<SyncCommitteeSubnets>(SYNC_COMMITTEE_BITFIELD_ENR_KEY)
+            .expect("Failed to get sync committee subnets")
+            .expect("Failed to decode sync committee subnets");
+
+        assert!(decoded_syncnets.get(0).unwrap());
+        assert!(decoded_syncnets.get(2).unwrap());
+        assert!(!decoded_syncnets.get(1).unwrap());
+        assert!(!decoded_syncnets.get(3).unwrap());
```
KolbyML, line 389: could we use assert_eq here and remove the unwrap

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,452 @@
...
+            .unwrap();
+
+        let enr = Enr::builder()
+            .add_value(ATTESTATION_BITFIELD_ENR_KEY, &attestation_subnets)
+            .add_value(SYNC_COMMITTEE_BITFIELD_ENR_KEY, &sync_committee_subnets)
+            .build(&combined_key)
+            .expect("Failed to build ENR");
+
+        // Test attestation subnet predicate
+        let att_predicate = attestation_subnet_predicate(&[5]);
```
KolbyML, line 434: can we not use shorthands for the variable name

crates/networking/discv5/src/discovery.rs
```diff
@@ -145,10 +155,15 @@ impl Discovery {
...
-                        Box::new(subnet_predicate(subnets.clone()))
+                match query.clone() {
+                    QueryType::Peers => Box::new(empty_predicate()),
+                    QueryType::AttestationSubnetPeers(subnet_ids) => {
+                        let subnet_ids_clone = subnet_ids.clone();
+                        Box::new(move |enr| attestation_subnet_predicate(&subnet_ids_clone)(enr))
+                    }
+                    QueryType::SyncCommitteeSubnetPeers(subnet_ids) => {
+                        let subnet_ids_clone = subnet_ids.clone();
+                        Box::new(move |enr| sync_committee_subnet_predicate(&subnet_ids_clone)(enr))
```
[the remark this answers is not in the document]
bomanaps: the clones are necessary because of lifetime issues, which leads to compiler error

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,452 @@
...
+        self.0
+            .set(index, value)
+            .map_err(|err| anyhow!("Subnet ID out of bounds: {err:?}"))
+    }
+
+    fn get(&self, index: usize) -> anyhow::Result<bool> {
+        self.0
+            .get(index)
+            .map_err(|err| anyhow!("Subnet ID out of bounds: {err:?}"))
+    }
```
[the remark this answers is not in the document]
bomanaps: If we remove these helper methods, we'd have to copy the same error handling code in multiple places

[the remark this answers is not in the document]
KolbyML: That is fine

crates/networking/discv5/src/discovery.rs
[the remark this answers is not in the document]
KolbyML:
```diff
diff --git a/crates/networking/discv5/src/discovery.rs b/crates/networking/discv5/src/discovery.rs
index c12d471..6ec1f7e 100644
--- a/crates/networking/discv5/src/discovery.rs
+++ b/crates/networking/discv5/src/discovery.rs
@@ -158,12 +158,10 @@ impl Discovery {
                 match query.clone() {
                     QueryType::Peers => Box::new(empty_predicate()),
                     QueryType::AttestationSubnetPeers(subnet_ids) => {
-                        let subnet_ids_clone = subnet_ids.clone();
-                        Box::new(move |enr| attestation_subnet_predicate(&subnet_ids_clone)(enr))
+                        Box::new(attestation_subnet_predicate(subnet_ids))
                     }
                     QueryType::SyncCommitteeSubnetPeers(subnet_ids) => {
-                        let subnet_ids_clone = subnet_ids.clone();
-                        Box::new(move |enr| sync_committee_subnet_predicate(&subnet_ids_clone)(enr))
+                        Box::new(sync_committee_subnet_predicate(subnet_ids))
                     }
                 },
                 target_peers,
@@ -200,16 +198,12 @@ impl Discovery {
                     self.find_peer_active = false;
                     match query.result {
                         Ok(peers) => {
-                            let predicate = attestation_subnet_predicate(&subnet_ids);
+                            let predicate = attestation_subnet_predicate(subnet_ids);
                             let filtered_peers = peers
                                 .into_iter()
                                 .filter(|enr| predicate(enr))
                                 .collect::<Vec<_>>();
-                            info!(
-                                "Found {} peers for subnets {:?}",
-                                filtered_peers.len(),
-                                subnet_ids
-                            );
+                            info!("Found {} peers for subnets", filtered_peers.len(),);
                             let mut peer_map = HashMap::new();
                             for peer in filtered_peers {
                                 peer_map.insert(peer, None);
@@ -226,15 +220,14 @@ impl Discovery {
                     self.find_peer_active = false;
                     match query.result {
                         Ok(peers) => {
-                            let predicate = sync_committee_subnet_predicate(&subnet_ids);
+                            let predicate = sync_committee_subnet_predicate(subnet_ids);
                             let filtered_peers = peers
                                 .into_iter()
                                 .filter(|enr| predicate(enr))
                                 .collect::<Vec<_>>();
                             info!(
-                                "Found {} peers for sync committee subnets {:?}",
+                                "Found {} peers for sync committee subnets",
                                 filtered_peers.len(),
-                                subnet_ids
                             );
                             let mut peer_map = HashMap::new();
                             for peer in filtered_peers {
diff --git a/crates/networking/discv5/src/subnet.rs b/crates/networking/discv5/src/subnet.rs
index f306a25..696879e 100644
--- a/crates/networking/discv5/src/subnet.rs
+++ b/crates/networking/discv5/src/subnet.rs
@@ -177,7 +177,7 @@ impl Decodable for SyncCommitteeSubnets {
     }
 }
 
-pub fn attestation_subnet_predicate(subnets: &[u8]) -> impl Fn(&Enr) -> bool + Send + Sync {
+pub fn attestation_subnet_predicate(subnets: Vec<u8>) -> impl Fn(&Enr) -> bool + Send + Sync {
     move |enr: &Enr| {
         if subnets.is_empty() {
             return true;
@@ -195,7 +195,7 @@ pub fn attestation_subnet_predicate(subnets: &[u8]) -> impl Fn(&Enr) -> bool + S
                 }
             };
 
-        for subnet_id in subnets {
+        for subnet_id in &subnets {
             if *subnet_id >= ATTESTATION_SUBNET_COUNT as u8 {
                 error!(
                     "Peer rejected: subnet ID {} exceeds attestation bitfield length; peer_id: {}",
@@ -220,7 +220,7 @@ pub fn attestation_subnet_predicate(subnets: &[u8]) -> impl Fn(&Enr) -> bool + S
     }
 }
 
-pub fn sync_committee_subnet_predicate(subnets: &[u8]) -> impl Fn(&Enr) -> bool + Send + Sync {
+pub fn sync_committee_subnet_predicate(subnets: Vec<u8>) -> impl Fn(&Enr) -> bool + Send + Sync {
     move |enr: &Enr| {
         if subnets.is_empty() {
             return true;
@@ -238,7 +238,7 @@ pub fn sync_committee_subnet_predicate(subnets: &[u8]) -> impl Fn(&Enr) -> bool
                 }
             };
 
-        for subnet_id in subnets {
+        for subnet_id in &subnets {
             if *subnet_id >= SYNC_COMMITTEE_SUBNET_COUNT as u8 {
                 trace!(
                     "Peer rejected: subnet ID {} exceeds sync committee bitfield length; peer_id: {}",
```

Here is a diff which will resolve the lifetime issues for you

crates/networking/discv5/src/subnet.rs
bomanaps: i got an error while doing which clippy suggests should be replaced with assert!() or assert!(!...)

bomanaps: I implemented a helper function to replace unwrap/expect calls with proper Result assertions since direct assert_eq! won't work with anyhow::Error which doesn't implement PartialEq.

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,485 @@
...
-            Subnet::SyncCommittee(_) => unimplemented!("SyncCommittee support not yet implemented"),
-        };
-        Ok(active)
+    pub fn enable_sync_committee_subnet(&mut self, subnet_id: u8) -> anyhow::Result<()> {
+        ensure!(
+            subnet_id < SYNC_COMMITTEE_SUBNET_COUNT as u8,
+            "Subnet ID {} exceeds maximum sync committee subnet count {}",
+            subnet_id,
+            SYNC_COMMITTEE_SUBNET_COUNT
+        );
```
KolbyML, line 122:
```suggestion
        ensure!(
            subnet_id < SYNC_COMMITTEE_SUBNET_COUNT as u8,
            "Subnet ID {subnet_id} exceeds maximum sync committee subnet count {SYNC_COMMITTEE_SUBNET_COUNT}",
        );
```

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,485 @@
...
+        Ok(())
+    }
+
+    pub fn disable_sync_committee_subnet(&mut self, subnet_id: u8) -> anyhow::Result<()> {
+        ensure!(
+            subnet_id < SYNC_COMMITTEE_SUBNET_COUNT as u8,
+            "Subnet ID {} exceeds maximum sync committee subnet count {}",
+            subnet_id,
+            SYNC_COMMITTEE_SUBNET_COUNT
+        );
```
KolbyML, line 133: same as above

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,485 @@
...
+        Ok(())
+    }
+
+    pub fn is_sync_committee_subnet_enabled(&self, subnet_id: u8) -> anyhow::Result<bool> {
+        ensure!(
+            subnet_id < SYNC_COMMITTEE_SUBNET_COUNT as u8,
+            "Subnet ID {} exceeds maximum sync committee subnet count {}",
+            subnet_id,
+            SYNC_COMMITTEE_SUBNET_COUNT
+        );
```
KolbyML, line 144: same as above

crates/networking/discv5/src/subnet.rs
```diff
@@ -1,173 +1,485 @@
...
+        assert!(
+            attestation_result.is_some(),
+            "Missing attestation subnet field in ENR"
+        );
+        let decoded_attestation_subnets = attestation_result.unwrap().unwrap();
+
+        assert_subnet_value(decoded_attestation_subnets.get(3), true, 3);
+        assert_subnet_value(decoded_attestation_subnets.get(42), true, 42);
+        assert_subnet_value(decoded_attestation_subnets.get(0), false, 0);
+        assert_subnet_value(decoded_attestation_subnets.get(10), false, 10);
```
KolbyML, line 405:
```diff
diff --git a/crates/networking/discv5/src/subnet.rs b/crates/networking/discv5/src/subnet.rs
index 8242e54..2ba89fa 100644
--- a/crates/networking/discv5/src/subnet.rs
+++ b/crates/networking/discv5/src/subnet.rs
@@ -262,18 +262,6 @@ mod tests {
 
     use super::*;
 
-    // Helper function to assert on Result<bool, anyhow::Error> values
-    fn assert_subnet_value(result: Result<bool, anyhow::Error>, expected: bool, subnet_id: usize) {
-        match result {
-            Ok(value) => assert_eq!(
-                value, expected,
-                "Subnet {} value should be {}",
-                subnet_id, expected
-            ),
-            Err(e) => panic!("Failed to get value for subnet {}: {:?}", subnet_id, e),
-        }
-    }
-
     #[test]
     fn test_decodes_subnets() {
         let enr = Enr::from_str("enr:-LS4QLe5eq5PFn1ZynqkrF6yg6ZGoplSDSNEPXtXfQh0vqhrDBQZICVoQu-AdeBOmtOFcAO7a0tJLdSlqStkdxkXnwaCCKSHYXR0bmV0c4gAAAAAAAAAMIRldGgykGqVoakEAAAA__________-CaWSCdjSCaXCEywwIqolzZWNwMjU2azGhA2JDBvnFqwtkUx34b_OdHXN1eO2JBMLWbzZXfGksk3YRg3RjcIIjkYN1ZHCCI5E").unwrap();
@@ -349,10 +337,22 @@ mod tests {
         );
         let decoded_attestation_subnets = attestation_result.unwrap().unwrap();
 
-        assert_subnet_value(decoded_attestation_subnets.get(1), true, 1);
-        assert_subnet_value(decoded_attestation_subnets.get(5), true, 5);
-        assert_subnet_value(decoded_attestation_subnets.get(0), false, 0);
-        assert_subnet_value(decoded_attestation_subnets.get(10), false, 10);
+        assert!(matches!(
+            decoded_attestation_subnets.is_attestation_subnet_enabled(1),
+            Ok(true)
+        ));
+        assert!(matches!(
+            decoded_attestation_subnets.is_attestation_subnet_enabled(5),
+            Ok(true)
+        ));
+        assert!(matches!(
+            decoded_attestation_subnets.is_attestation_subnet_enabled(0),
+            Ok(false)
+        ));
+        assert!(matches!(
+            decoded_attestation_subnets.is_attestation_subnet_enabled(10),
+            Ok(false)
+        ));
 
         let sync_committee_result =
             enr.get_decodable::<SyncCommitteeSubnets>(SYNC_COMMITTEE_BITFIELD_ENR_KEY);
@@ -362,10 +362,22 @@ mod tests {
         );
         let decoded_sync_committee_subnets = sync_committee_result.unwrap().unwrap();
 
-        assert_subnet_value(decoded_sync_committee_subnets.get(0), true, 0);
-        assert_subnet_value(decoded_sync_committee_subnets.get(2), true, 2);
-        assert_subnet_value(decoded_sync_committee_subnets.get(1), false, 1);
-        assert_subnet_value(decoded_sync_committee_subnets.get(3), false, 3);
+        assert!(matches!(
+            decoded_sync_committee_subnets.is_sync_committee_subnet_enabled(0),
+            Ok(true)
+        ));
+        assert!(matches!(
+            decoded_sync_committee_subnets.is_sync_committee_subnet_enabled(2),
+            Ok(true)
+        ));
+        assert!(matches!(
+            decoded_sync_committee_subnets.is_sync_committee_subnet_enabled(1),
+            Ok(false)
+        ));
+        assert!(matches!(
+            decoded_sync_committee_subnets.is_sync_committee_subnet_enabled(3),
+            Ok(false)
+        ));
     }
 
     #[test]
@@ -399,10 +411,22 @@ mod tests {
         );
         let decoded_attestation_subnets = attestation_result.unwrap().unwrap();
 
-        assert_subnet_value(decoded_attestation_subnets.get(3), true, 3);
-        assert_subnet_value(decoded_attestation_subnets.get(42), true, 42);
-        assert_subnet_value(decoded_attestation_subnets.get(0), false, 0);
-        assert_subnet_value(decoded_attestation_subnets.get(10), false, 10);
+        assert!(matches!(
+            decoded_attestation_subnets.is_attestation_subnet_enabled(3),
+            Ok(true)
+        ));
+        assert!(matches!(
+            decoded_attestation_subnets.is_attestation_subnet_enabled(42),
+            Ok(true)
+        ));
+        assert!(matches!(
+            decoded_attestation_subnets.is_attestation_subnet_enabled(0),
+            Ok(false)
+        ));
+        assert!(matches!(
+            decoded_attestation_subnets.is_attestation_subnet_enabled(10),
+            Ok(false)
+        ));
 
         let sync_committee_result =
             enr.get_decodable::<SyncCommitteeSubnets>(SYNC_COMMITTEE_BITFIELD_ENR_KEY);
@@ -412,10 +436,22 @@ mod tests {
         );
         let decoded_sync_committee_subnets = sync_committee_result.unwrap().unwrap();
 
-        assert_subnet_value(decoded_sync_committee_subnets.get(0), true, 0);
-        assert_subnet_value(decoded_sync_committee_subnets.get(2), true, 2);
-        assert_subnet_value(decoded_sync_committee_subnets.get(1), false, 1);
-        assert_subnet_value(decoded_sync_committee_subnets.get(3), false, 3);
+        assert!(matches!(
+            decoded_sync_committee_subnets.is_sync_committee_subnet_enabled(0),
+            Ok(true)
+        ));
+        assert!(matches!(
+            decoded_sync_committee_subnets.is_sync_committee_subnet_enabled(2),
+            Ok(true)
+        ));
+        assert!(matches!(
+            decoded_sync_committee_subnets.is_sync_committee_subnet_enabled(1),
+            Ok(false)
+        ));
+        assert!(matches!(
+            decoded_sync_committee_subnets.is_sync_committee_subnet_enabled(3),
+            Ok(false)
+        ));
 
         let attestation_subnet_predicate_fn = attestation_subnet_predicate(vec![3]);
         assert!(attestation_subnet_predicate_fn(&enr));

```

KolbyML (commented):
Some of the comment's I left weren't addressed, but I was requested for a review again?

crates/networking/discv5/src/discovery.rs
bomanaps: the this is about the dead code we are using this as you suggested, removing the dead code we then have to remove all match arms handling this variant in start_query and process_queries functions, remove any related predicate code specific to sync committee subnet peers

KolbyML:
```diff
diff --git a/crates/networking/discv5/src/discovery.rs b/crates/networking/discv5/src/discovery.rs
index 58dccf4..0d8019b 100644
--- a/crates/networking/discv5/src/discovery.rs
+++ b/crates/networking/discv5/src/discovery.rs
@@ -48,9 +48,7 @@ enum EventStream {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-#[allow(clippy::enum_variant_names)]
-#[allow(dead_code)]
-enum QueryType {
+pub enum QueryType {
     Peers,
     AttestationSubnetPeers(Vec<u8>),
     SyncCommitteeSubnetPeers(Vec<u8>),
@@ -131,7 +129,7 @@ impl Discovery {
         &self.local_enr
     }
 
-    pub fn discover_peers(&mut self, target_peers: usize, subnet_id: Option<u8>) {
+    pub fn discover_peers(&mut self, query: QueryType, target_peers: usize) {
         // If the discv5 service isn't running or we are in the process of a query, don't bother
         // queuing a new one.
         if !self.started || self.find_peer_active {
@@ -139,11 +137,6 @@ impl Discovery {
         }
         self.find_peer_active = true;
 
-        let query = match subnet_id {
-            Some(id) => QueryType::AttestationSubnetPeers(vec![id]),
-            None => QueryType::Peers,
-        };
-
         self.start_query(query, target_peers);
     }
 
@@ -440,7 +433,7 @@ mod tests {
         discovery.discv5.add_enr(peer_enr.clone()).unwrap();
 
         // Discover peers on subnet 0
-        discovery.discover_peers(0, Some(1));
+        discovery.discover_peers(QueryType::AttestationSubnetPeers(vec![1]), 0);
 
         // Mock the query result to bypass async polling
         discovery.discovery_queries.clear();
diff --git a/crates/networking/p2p/src/network.rs b/crates/networking/p2p/src/network.rs
index 86d4e73..81fd736 100644
--- a/crates/networking/p2p/src/network.rs
+++ b/crates/networking/p2p/src/network.rs
@@ -31,7 +31,7 @@ use libp2p::{
 use libp2p_identity::{Keypair, PublicKey, secp256k1};
 use libp2p_mplex::{MaxBufferBehaviour, MplexConfig};
 use parking_lot::Mutex;
-use ream_discv5::discovery::{DiscoveredPeers, Discovery};
+use ream_discv5::discovery::{DiscoveredPeers, Discovery, QueryType};
 use ream_executor::ReamExecutor;
 use tracing::{error, info, trace, warn};
 use yamux::Config as YamuxConfig;
@@ -92,7 +92,7 @@ impl Network {
         let discovery = {
             let mut discovery =
                 Discovery::new(Keypair::from(local_key.clone()), &config.discv5_config).await?;
-            discovery.discover_peers(16, None);
+            discovery.discover_peers(QueryType::Peers, 16);
             discovery
         };
 

```

If you make the changes above you won't get complaints by clippy

crates/networking/discv5/src/discovery.rs
KolbyML, line 52:
```diff
diff --git a/crates/networking/discv5/src/discovery.rs b/crates/networking/discv5/src/discovery.rs
index 58dccf4..0d8019b 100644
--- a/crates/networking/discv5/src/discovery.rs
+++ b/crates/networking/discv5/src/discovery.rs
@@ -48,9 +48,7 @@ enum EventStream {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-#[allow(clippy::enum_variant_names)]
-#[allow(dead_code)]
-enum QueryType {
+pub enum QueryType {
     Peers,
     AttestationSubnetPeers(Vec<u8>),
     SyncCommitteeSubnetPeers(Vec<u8>),
@@ -131,7 +129,7 @@ impl Discovery {
         &self.local_enr
     }
 
-    pub fn discover_peers(&mut self, target_peers: usize, subnet_id: Option<u8>) {
+    pub fn discover_peers(&mut self, query: QueryType, target_peers: usize) {
         // If the discv5 service isn't running or we are in the process of a query, don't bother
         // queuing a new one.
         if !self.started || self.find_peer_active {
@@ -139,11 +137,6 @@ impl Discovery {
         }
         self.find_peer_active = true;
 
-        let query = match subnet_id {
-            Some(id) => QueryType::AttestationSubnetPeers(vec![id]),
-            None => QueryType::Peers,
-        };
-
         self.start_query(query, target_peers);
     }
 
@@ -440,7 +433,7 @@ mod tests {
         discovery.discv5.add_enr(peer_enr.clone()).unwrap();
 
         // Discover peers on subnet 0
-        discovery.discover_peers(0, Some(1));
+        discovery.discover_peers(QueryType::AttestationSubnetPeers(vec![1]), 0);
 
         // Mock the query result to bypass async polling
         discovery.discovery_queries.clear();
diff --git a/crates/networking/p2p/src/network.rs b/crates/networking/p2p/src/network.rs
index 86d4e73..81fd736 100644
--- a/crates/networking/p2p/src/network.rs
+++ b/crates/networking/p2p/src/network.rs
@@ -31,7 +31,7 @@ use libp2p::{
 use libp2p_identity::{Keypair, PublicKey, secp256k1};
 use libp2p_mplex::{MaxBufferBehaviour, MplexConfig};
 use parking_lot::Mutex;
-use ream_discv5::discovery::{DiscoveredPeers, Discovery};
+use ream_discv5::discovery::{DiscoveredPeers, Discovery, QueryType};
 use ream_executor::ReamExecutor;
 use tracing::{error, info, trace, warn};
 use yamux::Config as YamuxConfig;
@@ -92,7 +92,7 @@ impl Network {
         let discovery = {
             let mut discovery =
                 Discovery::new(Keypair::from(local_key.clone()), &config.discv5_config).await?;
-            discovery.discover_peers(16, None);
+            discovery.discover_peers(QueryType::Peers, 16);
             discovery
         };
 

```

If you make the changes above you won't get complaints by clippy about removing the deadcode line

KolbyML (commented):
this will make removing the deadcode line work for you

crates/networking/discv5/src/subnet.rs
bomanaps: i have taken this and `for subnet_id in subnets {` into account still the same end?

KolbyML (approved):
:shipit: the PR looks great :fire:. Thank you for your great work! :rocket:

PR #370, opened 2025-05-04
feat: implemented macro for random and tests

Kayden-ML:

Fixes: https://github.com/ReamLabs/ream/issues/368
This PR implements testing for random

Spec: https://github.com/ethereum/consensus-specs/tree/8fa1aaae5eaca6d9bdcea99658e834963557c20d/tests/formats/random

### To-Do


testing/ef-tests/src/macros/sanity_blocks.rs
```diff
@@ -1,95 +1,97 @@
...
-    () => {
-        #[cfg(test)]
-        #[allow(non_snake_case)]
-        mod test_sanity_blocks {
-            use std::{fs, path::Path};
+    ($operation_name:ident, $path:expr) => {
+        paste::paste! {
+            #[cfg(test)]
+            #[allow(non_snake_case)]
+            mod $operation_name {
```
KolbyML, line 7:
```suggestion
    ($path:expr) => {
        paste::paste! {
            #[cfg(test)]
            #[allow(non_snake_case)]
            mod $path {
```

couldn't we just use the `path` as the mod name?

KolbyML: Oh maybe the / is an issue? if that is the case then this is fine

Kayden-ML: > Oh maybe the / is an issue? if that is the case then this is fine

I was running into an issue with the / so I did it this way

KolbyML (approved):
:shipit: looks good

PR #371, opened 2025-05-04
feat: Implement Checkpoint Sync

### LINKED ISSUE #317: Implement checkpoint sync
### Describe the feature

Since we are only planning to implement the latest fork*, the only way we can sync is checkpoint sync. Hence we want to implement checkpoint sync

### Additional context

_No response_

varun-doshi:

Fixes #317 
This PR implements Checkpoint Sync in Ream.

Ream uses the trusted URL provided to fetch initial block from a checkpoint.(say slot N).
Following that, it should connect to peers and fetch blocks from N+1. Since our p2p is still building up, this mocks receiving blocks from the P2P layers and processes them to do the STF.

To run:
```
cargo run node --data-dir <DIRECTORY> --sync-mode checkpoint-sync --rpc-url <RPC URL>
```

Caution: 
- Currently runs on Sepolia since it supports electra. 
- For RPC: The Sepolia checkpoint sync URLs maintained by the eth-clients team all have a mismatch between the Response structure where the field `finalized: true` is missing. This is only applicable to the Sepolia URLs(for every RPC needed in checkpoint sync). The mainnet URLs all have the exact Response type as per spec. Hence, it is advised to test using a private centralized RPC like quicknode.

### To-Do
- [ ] Implement weak subjectivity check for user provided epoch(will do this next)


KolbyML:
image
@varun-doshi you need to rebase

bin/ream/src/main.rs
```diff
@@ -53,11 +53,32 @@ async fn main() {
...
+                }
+                (_store, _slot) = config
+                    .sync_mode
+                    .sync(ream_db.clone(), &config.rpc_url.unwrap())
+                    .await
+                    .map_err(|err| panic!("{err}"))
+                    .unwrap();
+
+                info!("Checkpoint Sync complete");
+            }
```
KolbyML, line 80:
```suggestion
            if database.is_empty {
                 initialize_database_using_checkpoint
            }
            
            // todo add syncing code once p2p code is more mature.
```

We aren't actually syncing here. this should just check if the database is empty and if it is, it should download the State ,Block, and Blobs. So that we can then sync from P2P

KolbyML (commented):
I left some feedback

KolbyML:
@varun-doshi for the checkpoint_url it should be optional, we should have defaults set for mainnet, and the testnets I would just grab trusted ones from eth-clients or something

crates/common/sync/src/checkpoint.rs
```diff
@@ -0,0 +1,140 @@
...
+    ensure!(
+        block.data.message.slot == state.data.slot,
+        anyhow::anyhow!("Slot mismatch")
+    );
+
+    ensure!(block.data.message.state_root == state.data.state_hash());
+    let mut store = get_forkchoice_store(state.data, block.data.message, db)?;
+
+    // using sepolia genesis time until pectra mainnet
+    let time = 1655733600 + SECONDS_PER_SLOT * (slot + 1);
```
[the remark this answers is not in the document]
varun-doshi: The mainnet genesis time is already stored as a variable...its a number here since this is genesis time of sepolia which we wont use anyways when I update the code after 7th to work for mainnet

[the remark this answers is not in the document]
KolbyML: This code should work for both mainnet and sepolia you should be getting the timestamp like this 
```suggestion
    let time = network_spec().genesis.genesis_time + SECONDS_PER_SLOT * (slot + 1);
```

bin/ream/src/cli/mod.rs
```diff
@@ -86,6 +86,25 @@ pub struct NodeConfig {
...
+        help = "Trusted RPC URL to initiate Checkpoint Sync."
+    )]
+    pub checkpoint_sync_url: Option<String>,
+
+    #[arg(
+        long = "purge-db",
+        help = "Purges the database.",
+        action = clap::ArgAction::SetTrue
+    )]
+    pub purge_db: bool,
```
KolbyML, line 101:
```suggestion
    #[arg(
        long = "purge-db",
        help = "Purges the database.",
    )]
    pub purge_db: bool,
```

we don't need to define this action it is infered

crates/common/sync/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+};
+use ream_fork_choice::{
+    handlers::on_tick,
+    store::{Store, get_forkchoice_store},
+};
+use ream_network_spec::networks::network_spec;
+use ream_rpc::types::response::BeaconVersionedResponse;
+use ream_storage::{db::ReamDB, tables::Table};
+use tracing::{info, warn};
+pub mod checkpoint;
```
KolbyML, line 19:
```suggestion
pub mod checkpoint;

use alloy_primitives::B256;
use anyhow::{anyhow, ensure};
use checkpoint::fetch_default_checkpoint_url;
use ream_consensus::{
    blob_sidecar::{BlobIdentifier, BlobSidecar},
    constants::SECONDS_PER_SLOT,
    electra::{beacon_block::SignedBeaconBlock, beacon_state::BeaconState},
    execution_engine::rpc_types::get_blobs::BlobAndProofV1,
    misc::{compute_epoch_at_slot, compute_start_slot_at_epoch},
};
use ream_fork_choice::{
    handlers::on_tick,
    store::{Store, get_forkchoice_store},
};
use ream_network_spec::networks::network_spec;
use ream_rpc::types::response::BeaconVersionedResponse;
use ream_storage::{db::ReamDB, tables::Table};
use tracing::{info, warn};
```

the mod should be at the top of the file

crates/common/sync/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// Entry point for checkpoint sync.
+pub async fn initialize_db_from_checkpoint(
+    db: ReamDB,
+    rpc: Option<String>,
```
KolbyML, line 24:
```suggestion
    checkpoint_sync_url: Option<String>,
```

crates/common/sync/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+) -> anyhow::Result<(Store, u64)> {
+    let checkpoint_sync_url = match rpc {
+        Some(url) => url,
+        // 0 for Mainnet, 1 for Sepolia
+        None => fetch_default_checkpoint_url().expect("Unable to fetch default checkpoint url")[1]
+            .checkpoint_urls[0]
+            .clone(),
+    };
```
KolbyML, line 32:
```suggestion
    let checkpoint_sync_url = match rpc {
        Some(checkpoint_sync_url) => checkpoint_sync_url,
        None => fetch_default_checkpoint_url().expect("Unable to fetch default checkpoint url")[1]
            .checkpoint_urls[0]
            .clone(),
    };
```

We should be using the network spec to get which network it is and


You can do something like we did for

```
impl Bootnodes {
    pub fn to_enrs(self, network: Network) -> Vec<Enr> {
```

crates/common/sync/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+    info!("Received block from slot:{}", slot);
+    ensure!(
+        block.data.message.slot == state.data.slot,
+        anyhow::anyhow!("Slot mismatch")
+    );
+
+    ensure!(block.data.message.state_root == state.data.state_hash());
+    let mut store = get_forkchoice_store(state.data, block.data.message, db)?;
+
+    // using sepolia genesis time until pectra mainnet
```
KolbyML, line 62: This code isn't network specific right? so we don't need to state sepolia here

crates/storage/src/db.rs
```diff
@@ -191,3 +196,29 @@ impl ReamDB {
...
+        if input.trim().eq_ignore_ascii_case("y") {
+            match fs::remove_dir_all(&db_path) {
+                Ok(_) => println!("Database cleared successfully."),
+                Err(e) => eprintln!("Failed to remove database: {}", e),
+            }
+        } else {
+            println!("Database path does not exist: {:?}", db_path);
+        }
+    } else {
+        println!("Operation canceled.");
```
KolbyML, line 221: we shouldn't be using `println` in production code us `tracing::*!()` instead

crates/common/sync/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+pub async fn fetch_blobs(rpc: &str, store: ReamDB, beacon_block_root: B256) -> anyhow::Result<()> {
+    let body = reqwest::get(format!(
+        "{}{}{}",
+        rpc, "/eth/v1/beacon/blob_sidecars/", beacon_block_root
+    ))
+    .await?
+    .text()
+    .await?;
+
+    let blob_sidecar: BeaconVersionedResponse<Vec<BlobSidecar>> = serde_json::from_str(&body)?;
```
KolbyML, line 136:
```suggestion
    let blob_sidecar = reqwest::get(format!(
        "{rpc}/eth/v1/beacon/blob_sidecars/{}",
        rpc, beacon_block_root
    ))
    .await?
    .json::<BeaconVersionedResponse<Vec<BlobSidecar>>>()
    .await?;
```

can't we do something like this?

crates/common/sync/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    info!("Starting Checkpoint Sync");
+    let current_slot = db
+        .slot_index_provider()
+        .get_highest_slot()
+        .map_err(|err| anyhow!("Unable to fetch highest slot in db:{}", err))?;
+
+    if current_slot.is_some() {
+        warn!("Starting Checkpoint Sync from existing DB. It is advised to start from a fresh DB");
+    }
```
KolbyML, line 42: this function should immediately return if the database was already initialized, we shouldn't be using this code on an existing db

crates/common/sync/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+    let body = reqwest::get(format!("{}{}{}", rpc, "/eth/v2/debug/beacon/states/", slot))
+        .await?
+        .text()
+        .await?;
+
+    let state: BeaconVersionedResponse<BeaconState> = serde_json::from_str(&body)?;
+    info!("State hash:{}", state.data.state_hash());
+
+    Ok(state)
+}
```
KolbyML, line 84:
```suggestion
pub async fn get_state(
    rpc: &str,
    slot: u64,
) -> anyhow::Result<BeaconVersionedResponse<BeaconState>> {
    let state = reqwest::get(format!("{rpc}/eth/v2/debug/beacon/states/{slot}"))
        .await?
        .json<BeaconVersionedResponse<BeaconState>>()
        .await?;

        info!("Fetched State Root:{}", state.data.state_root());

    Ok(state)
}
```

This function isn't fetching a finalized state it is fetching whatever slot we pass

crates/common/sync/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+    info!("Starting Checkpoint Sync");
+    let current_slot = db
+        .slot_index_provider()
+        .get_highest_slot()
+        .map_err(|err| anyhow!("Unable to fetch highest slot in db:{}", err))?;
+
+    if current_slot.is_some() {
+        warn!("Starting Checkpoint Sync from existing DB. It is advised to start from a fresh DB");
+    }
+    let block = fetch_finalized_block(&checkpoint_sync_url, current_slot).await?;
```
KolbyML, line 43:
```suggestion
    let block = fetch_finalized_block(&checkpoint_sync_url).await?;
```

crates/common/sync/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+            format!("{}{}{}", rpc, "/eth/v2/beacon/blocks/", slot + 1)
+        }
+        None => {
+            format!("{}{}", rpc, "/eth/v2/beacon/blocks/finalized")
+        }
+    };
+
+    let body = reqwest::get(url).await?.text().await?;
+
+    let mut block: BeaconVersionedResponse<SignedBeaconBlock> = serde_json::from_str(&body)?;
```
KolbyML, line 103:
```suggestion
pub async fn fetch_finalized_block(
    rpc: &str,
) -> anyhow::Result<BeaconVersionedResponse<SignedBeaconBlock>> {
    let block = reqwest::get(format!("{rpc}/eth/v2/beacon/blocks/finalized")).await?.json<BeaconVersionedResponse<SignedBeaconBlock>>().await?;
```

are we fetching a finalized block or any block? the function name says finalized

crates/common/sync/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+        warn!("Slot {} is not start of epoch", block.data.message.slot);
+        let body = reqwest::get(format!(
+            "{}{}{}",
+            rpc, "/eth/v2/beacon/blocks/", epoch_start_slot
+        ))
+        .await?
+        .text()
+        .await?;
+        block = serde_json::from_str(&body)?;
+    }
```
KolbyML, line 119: Why is this code added? The start of the epoch could be N + 1 if the slot was skipped for N no?

crates/common/consensus/src/electra/beacon_block.rs
```diff
@@ -67,3 +67,9 @@ pub struct BeaconBlock {
     pub state_root: B256,
     pub body: BeaconBlockBody,
 }
+
+impl BeaconBlock {
+    pub fn block_hash(&self) -> B256 {
+        self.tree_hash_root()
+    }
```
KolbyML, line 74: wouldn't this be referred to as a block_root?

crates/common/consensus/src/electra/beacon_state.rs
```diff
@@ -2662,6 +2662,10 @@ impl BeaconState {
         ]
         .concat())
     }
+
+    pub fn state_hash(&self) -> B256 {
+        self.tree_hash_root()
+    }
```
KolbyML, line 2668: isn't this referred to as a state root?

bin/ream/src/main.rs
```diff
@@ -56,11 +57,26 @@ async fn main() {
...
 
             info!("ream database initialized ");
 
+            let _store;
+            let _slot;
+            (_store, _slot) =
+                initialize_db_from_checkpoint(ream_db.clone(), config.checkpoint_sync_url)
+                    .await
+                    .map_err(|err| panic!("{err}"))
+                    .unwrap();
```
KolbyML, line 76:
```suggestion
           initialize_db_from_checkpoint(ream_db.clone(), config.checkpoint_sync_url)
                    .await
                    .map_err(|err| panic!("{err}"))
                    .unwrap();
```

Why are we returning store? if we already have access to it, also slot isn't used so why return it?

bin/ream/src/cli/mod.rs
```diff
@@ -86,6 +86,25 @@ pub struct NodeConfig {
...
+
+    #[arg(
+        long = "ws-epoch",
+        help = "Optional epoch to start Checkpoint Sync from."
+    )]
+    pub ws_epoch: Option<u64>,
```
KolbyML, line 107:
```suggestion
```

could we remove this argument for now, if we need it in the future we can add it then, I am not sure we should give users this option

crates/common/sync/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+ream-storage.workspace = true
+ream-fork-choice.workspace = true
+
+alloy-primitives.workspace = true
+anyhow.workspace = true
+reqwest.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+serde_yaml.workspace = true
+tracing.workspace = true
```
KolbyML, line 25: Can you add comments like we do in our other crates

crates/common/sync/Cargo.toml
KolbyML, line 1: could this crate be called `checkpoint-sync`, as it is more descriptive and `sync` is a bit misleading here

crates/storage/src/db.rs
```diff
@@ -191,3 +196,29 @@ impl ReamDB {
...
+            path
+        );
+        io::stdout().flush().unwrap();
+
+        let mut input = String::new();
+        io::stdin().read_line(&mut input).unwrap();
+        if input.trim().eq_ignore_ascii_case("y") {
+            match fs::remove_dir_all(&db_path) {
+                Ok(_) => println!("Database cleared successfully."),
+                Err(e) => eprintln!("Failed to remove database: {}", e),
```
KolbyML, line 215:
```suggestion
                Err(e) => eprintln!("Failed to remove database: {}", e),
```

shouldn't we return an error if an error happens

KolbyML (commented):
Left some feedback

crates/common/sync/src/lib.rs
varun-doshi: Hmm you're right...this section is not needed

bin/ream/src/cli/mod.rs
varun-doshi: Sounds good, I'll open a separate PR for this after this which will work on weak subjectivity checks

crates/storage/src/db.rs
```diff
@@ -191,3 +197,29 @@ impl ReamDB {
...
+
+pub fn reset_db(path: Option<PathBuf>) -> anyhow::Result<()> {
+    let db_path = path.clone().ok_or_else(|| anyhow!("Path does not exist"))?;
+
+    if db_path.exists() {
+        info!(
+            "Are you sure you want to remove the database at {:?}? (y/n):",
+            path
+        );
+        io::stdout().flush().unwrap();
```
KolbyML, line 209: we shouldn't use unwrap here, also is the y/n used right now?

bin/ream/src/cli/mod.rs
```diff
@@ -86,6 +86,15 @@ pub struct NodeConfig {
         help = "One or more comma-delimited base64-encoded ENR's of peers to initially connect to. Use 'default' to use the default bootnodes for the network. Use 'none' to disable bootnodes."
     )]
     pub bootnodes: Bootnodes,
+
+    #[arg(
+        long = "checkpoint-sync-url",
+        help = "Trusted RPC URL to initiate Checkpoint Sync."
+    )]
+    pub checkpoint_sync_url: Option<String>,
```
KolbyML, line 94:
```suggestion
    pub checkpoint_sync_url: Option<Url>,
```

we should use Url instead of a string

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+) -> anyhow::Result<BeaconVersionedResponse<BeaconState>> {
+    let state = reqwest::get(format!("{rpc}/eth/v2/debug/beacon/states/{slot}"))
+        .await?
+        .json::<BeaconVersionedResponse<BeaconState>>()
+        .await?;
+
+    info!("Fetched State Root:{}", state.data.state_root());
+
+    Ok(state)
+}
```
KolbyML, line 97:
```suggestion
/// Fetch initial state from trusted RPC
pub async fn get_state(
    rpc: &str,
    slot: u64,
) -> anyhow::Result<BeaconVersionedResponse<BeaconState>> {
    Ok(reqwest::get(format!("{rpc}/eth/v2/debug/beacon/states/{slot}"))
        .await?
        .json::<BeaconVersionedResponse<BeaconState>>()
        .await?)
}
```

Maybe we should log outside of this function

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+) -> anyhow::Result<BeaconVersionedResponse<SignedBeaconBlock>> {
+    let block = reqwest::get(format!("{rpc}/eth/v2/beacon/blocks/finalized"))
+        .await?
+        .json::<BeaconVersionedResponse<SignedBeaconBlock>>()
+        .await?;
+
+    info!("Fetched Block Root:{}", block.data.message.block_root());
+
+    Ok(block)
+}
```
KolbyML, line 111:
```suggestion
/// Fetch initial block from trusted RPC
pub async fn fetch_finalized_block(
    rpc: &str,
) -> anyhow::Result<BeaconVersionedResponse<SignedBeaconBlock>> {
    Ok(reqwest::get(format!("{rpc}/eth/v2/beacon/blocks/finalized"))
        .await?
        .json::<BeaconVersionedResponse<SignedBeaconBlock>>()
        .await?)
}
```

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+        store.blobs_and_proofs_provider().insert(
+            BlobIdentifier::new(beacon_block_root, blob_sidecar.index),
+            BlobAndProofV1 {
+                blob: blob_sidecar.blob,
+                proof: blob_sidecar.kzg_proof,
+            },
+        )?;
+    }
+    Ok(())
+}
```
KolbyML, line 133: yet blobs sounds like we are returning the blobs? maybe we should change the function name or move the storing part outside of here. As the function name should describe the function.

crates/storage/src/db.rs
varun-doshi: I feel this is a good warning before deleting the DB just so the user knows its going to be cleared
But the y/n functions exactly as expected

crates/common/checkpoint_sync/src/checkpoint.rs
```diff
@@ -0,0 +1,39 @@
...
+        Network::Hoodi => serde_yaml::from_str(include_str!(
+            "../resources/checkpoint_sync_sources/hoodi.yaml"
+        ))
+        .expect("should deserialize checkpoint sync sources"),
+        Network::Dev => vec![],
+    };
+
+    raw_urls
+        .into_iter()
+        .map(|s| Url::parse(&s).expect("invalid URL in checkpoint sync YAML"))
```
KolbyML, line 37:
```suggestion
        .map(|raw_url| Url::parse(&raw_url).expect("invalid URL in checkpoint sync YAML"))
```

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+pub async fn initialize_db_from_checkpoint(
+    db: ReamDB,
+    checkpoint_sync_url: Option<Url>,
+) -> anyhow::Result<()> {
+    if db.is_initialized() {
+        warn!("Starting Checkpoint Sync from existing DB. It is advised to start from a fresh DB");
+        return Ok(());
+    }
+
+    let checkpoint_sync_url = get_checkpoint_sync_sources(checkpoint_sync_url)[0].clone();
```
KolbyML, line 28:
```suggestion
    let checkpoint_sync_url = get_checkpoint_sync_sources(checkpoint_sync_url).remove(0);
```

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+        db.clone(),
+        block.data.message.block_root(),
+    )
+    .await?;
+    info!("Received block from slot:{}", slot);
+
+    let state = get_state(&checkpoint_sync_url, block.data.message.slot).await?;
+    info!("Fetched State Root:{}", state.data.state_root());
+    ensure!(
+        block.data.message.slot == state.data.slot,
```
KolbyML, line 44: We define a variable `slot = block.data.message.slot`, but then we aren't using it

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+    beacon_block_root: B256,
+) -> anyhow::Result<()> {
+    let blob_sidecar = reqwest::get(format!(
+        "{rpc}/eth/v1/beacon/blob_sidecars/{beacon_block_root}"
+    ))
+    .await?
+    .json::<BeaconVersionedResponse<Vec<BlobSidecar>>>()
+    .await?;
+
+    info!("Got blobs len:{}", blob_sidecar.data.len());
```
KolbyML, line 96: Do we need this log?

crates/storage/src/db.rs
```diff
@@ -190,4 +196,37 @@ impl ReamDB {
             db: self.db.clone(),
         }
     }
+
+    pub fn is_initialized(&self) -> bool {
+        self.slot_index_provider()
+            .get_highest_slot()
+            .map_err(|err| anyhow!("Unable to fetch highest slot in db:{}", err))
```
KolbyML, line 203:
```suggestion
            .map_err(|err| anyhow!("Unable to fetch highest slot in db: {err}"))
```

bin/ream/src/main.rs
```diff
@@ -77,11 +78,21 @@ async fn main() {
                 gossipsub_config,
             };
 
-            let ream_db = ReamDB::new(config.data_dir, config.ephemeral)
+            if config.purge_db {
+                reset_db(config.data_dir.clone()).expect("Unable to delete database");
```
KolbyML, line 82: We shouldn't be passing config.data_dir here. We should be passing the result of

```
let ream_dir =
            dir::setup_data_dir(APP_NAME, data_dir, ephemeral).map_err(StoreError::Io)?;
```

Refactor it out of `ReamDB::new` then pass it in both here `reset_db(`, and in ` `ReamDB::new``            
         
         
  Both those shouldn't be taking Options anymore

crates/storage/src/db.rs
```diff
@@ -190,4 +196,37 @@ impl ReamDB {
...
+                Ok(_) => info!("Database cleared successfully."),
+                Err(err) => return Err(anyhow!("Failed to remove database:{}", err)),
+            }
+        } else {
+            info!("Database path does not exist: {:?}", db_path);
+        }
+    } else {
+        info!("Operation canceled.");
+    }
+    Ok(())
```
KolbyML, line 231:
```suggestion
pub fn reset_db(data_dir: PathBuf) -> anyhow::Result<()> {
    if !fs::read_dir(&db_path)?.next().is_some(); {
        info!("Data directory at {db_path:?} is already empty.");
        return Ok(());
    }

    info!("Are you sure you want to clear the contents of the data directory at {db_path:?}? (y/n):");
    let mut input = String::new();
    io::stdin().read_line(&mut input)?;
    if input.trim().eq_ignore_ascii_case("y") {
        for entry in fs::read_dir(&db_path)? {
            let entry = entry?;
            let path = entry.path();
            if path.is_dir() {
                fs::remove_dir_all(&path)?;
            } else {
                fs::remove_file(&path)?;
            }
        }
        info!("Database contents cleared successfully.");
    } else {
        info!("Operation canceled by user.");
    }

    Ok(())
```

The data directory will always exist, the question is, is it empty or not?

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+        block.data.message.block_root(),
+    )
+    .await?;
+    info!("Received block from slot:{}", slot);
+
+    let state = get_state(&checkpoint_sync_url, block.data.message.slot).await?;
+    info!("Fetched State Root:{}", state.data.state_root());
+    ensure!(
+        block.data.message.slot == state.data.slot,
+        anyhow::anyhow!("Slot mismatch")
```
KolbyML, line 45:
```suggestion
        "Slot mismatch"
```

can't we just do this?

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+    let block = fetch_finalized_block(&checkpoint_sync_url).await?;
+    info!("Fetched Block Root:{}", block.data.message.block_root());
+    let slot = block.data.message.slot;
+    initialize_blobs_in_db(
+        &checkpoint_sync_url,
+        db.clone(),
+        block.data.message.block_root(),
+    )
+    .await?;
+    info!("Received block from slot:{}", slot);
```
KolbyML, line 39: Some of these logs read weird also generally there should be a space between the {} so like `: {}`

KolbyML (commented):
Left some feedback

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+use reqwest::Url;
+use tracing::{info, warn};
+
+/// Entry point for checkpoint sync.
+pub async fn initialize_db_from_checkpoint(
+    db: ReamDB,
+    checkpoint_sync_url: Option<Url>,
+) -> anyhow::Result<()> {
+    if db.is_initialized() {
+        warn!("Starting Checkpoint Sync from existing DB. It is advised to start from a fresh DB");
```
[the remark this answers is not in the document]
varun-doshi: Wdym? I felt this is a good warning to the user. Since other clients dont allow cs from existing DB, we should notify that the best practice is to reset DB and run. Currently, user has option to pres `n` when the DB deleting confirmation pops up in the terminal.
Do you want an improvement on the wordings?

crates/storage/src/db.rs
```diff
@@ -190,4 +188,46 @@ impl ReamDB {
...
+        info!("Operation canceled by user.");
+    }
+    Ok(())
+}
+
+pub fn initialize_db_dir(
+    data_dir: Option<PathBuf>,
+    ephemeral: bool,
+) -> Result<PathBuf, StoreError> {
+    dir::setup_data_dir(APP_NAME, data_dir, ephemeral).map_err(StoreError::Io)
```
KolbyML, line 232: I think we should just use setup_data_dir directly, no need for an abstraction for 1 line

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    initialize_blobs_in_db(
+        &checkpoint_sync_url,
+        db.clone(),
+        block.data.message.block_root(),
+    )
+    .await?;
+    info!("Received blobs from slot: {}", slot);
+
+    let state = get_state(&checkpoint_sync_url, slot).await?;
+    info!("Received State Root: {}", state.data.state_root());
```
KolbyML, line 43: Did a received a state root? or did we just download the whole state?

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+) -> anyhow::Result<()> {
+    if db.is_initialized() {
+        warn!("Starting Checkpoint Sync from existing DB. It is advised to start from a clean DB");
+        return Ok(());
+    }
+
+    let checkpoint_sync_url = get_checkpoint_sync_sources(checkpoint_sync_url).remove(0);
+    info!("Initiating Checkpoint Sync");
+    let block = fetch_finalized_block(&checkpoint_sync_url).await?;
+    info!("Received Block Root: {}", block.data.message.block_root());
```
KolbyML, line 32: Did a received a block root? or did we just download the whole block?

crates/rpc/src/types/response.rs
```diff
@@ -116,3 +116,71 @@ impl<T: Serialize> DataVersionedResponse<T> {
...
+
+    deserializer.deserialize_any(BoolVisitor)
+}
+
+fn option_bool_from_str_or_bool<'de, D>(deserializer: D) -> Result<Option<bool>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    Ok(Some(bool_from_str_or_bool(deserializer)?))
+}
```
KolbyML, line 186:
```suggestion
/// A OptionalBeaconVersionedResponse data struct that can be used to wrap data type
/// used for json rpc responses
///
/// # Example
/// {
///  "data": json!({
///     "version": Some("electra")
///     "execution_optimistic" : Some("false"),
///     "finalized" : None,
///     "data" : T
/// })
/// }
#[derive(Debug, Serialize, Deserialize)]
pub struct OptionalBeaconVersionedResponse<T> {
    pub version: Option<String>,
    pub execution_optimistic: Option<bool>,
    pub finalized: Option<bool>,
    pub data: T,
}
```

if you are going to use this, can this be in the file where you use it. The custom deserializer isn't needed. But honestly we should just remove this and commit to using ssz for checkpoint sync no?

varun-doshi: Do you mean `OptionalBeaconVersionedResponse` completely? We still need it for blocks and blobs though.

And the custom deserializer is needed because some clients return boolean as strings while others do just bool

varun-doshi: Or do you mean to use ssz request for all -state, block, blobs?

KolbyML: > Or do you mean to use ssz request for all -state, block, blobs?

wouldn't it be better to do ssz requests for blocks and blobs as it is quicker

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    deserializer.deserialize_any(BoolVisitor)
+}
+
+fn option_bool_from_str_or_bool<'de, D>(deserializer: D) -> Result<Option<bool>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    Ok(Some(bool_from_str_or_bool(deserializer)?))
+}
```
KolbyML, line 95:
```suggestion
#[derive(Debug, Serialize, Deserialize)]
pub struct OptionalBeaconVersionedResponse<T> {
    pub version: Option<String>,
    pub execution_optimistic: Option<bool>,
    pub finalized: Option<bool>,
    pub data: T,
}
```

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+        .header(ACCEPT, HeaderValue::from_static("application/octet-stream"))
+        .send()
+        .await?
+        .bytes()
+        .await?;
+
+    Ok(BeaconVersionedResponse::new(
+        BeaconState::from_ssz_bytes(&state).expect("Unable to decode from SSZ"),
+    ))
+}
```
KolbyML, line 163:
```suggestion
/// Fetch initial state from trusted RPC
pub async fn get_state(
    rpc: &Url,
    slot: u64,
) -> anyhow::Result<BeaconState> {
    let client = reqwest::Client::new();
    let state = client
        .get(format!("{rpc}eth/v2/debug/beacon/states/{slot}"))
        .header(ACCEPT, HeaderValue::from_static("application/octet-stream"))
        .send()
        .await?
        .bytes()
        .await?;

    Ok(BeaconState::from_ssz_bytes(&state).expect("Unable to decode from SSZ"),
    )
}
```

can you remove the expect and use a `?` as well

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+        .header(ACCEPT, HeaderValue::from_static("application/octet-stream"))
+        .send()
+        .await?
+        .bytes()
+        .await?;
+
+    Ok(BeaconVersionedResponse::new(
+        SignedBeaconBlock::from_ssz_bytes(&raw_bytes).expect("Unable to decode from SSZ"),
+    ))
+}
```
KolbyML, line 181: we don't need `BeaconVersionedResponse` here also please remove the expect

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    header::{ACCEPT, HeaderValue},
+};
+use serde::{Deserialize, Serialize};
+use ssz::Decode;
+use tracing::{info, warn};
+
+pub const VERSION: &str = "electra";
+pub const ETH_CONSENSUS_VERSION_HEADER: &str = "Eth-Consensus-Version";
+const EXECUTION_OPTIMISTIC: bool = false;
+const FINALIZED: bool = false;
```
KolbyML, line 27:
```suggestion
```

we don't need these

bin/ream/src/main.rs
```diff
@@ -77,11 +81,23 @@ async fn main() {
                 gossipsub_config,
             };
 
-            let ream_db = ReamDB::new(config.data_dir, config.ephemeral)
-                .expect("unable to init Ream Database");
+            let ream_dir = setup_data_dir(APP_NAME, config.data_dir.clone(), config.ephemeral)
```
KolbyML, line 84: can you move the APP_NAME const to this crate, as it is no longer used in ream_storage right?

KolbyML (commented):
I left some feedback

crates/common/checkpoint_sync/src/lib.rs
varun-doshi: The expect is for `from_ssz_bytes` not the BeaconVersionedResponse::new.
I'll remove `BeaconVersionedResponse` and use map to handle the Decode err from ssz

bin/ream/src/main.rs
```diff
@@ -14,12 +15,16 @@ use ream_p2p::{
...
+use ream_storage::{
+    db::{ReamDB, reset_db},
+    dir::setup_data_dir,
+};
 use tracing::{error, info};
 use tracing_subscriber::EnvFilter;
 
 #[tokio::main]
 async fn main() {
+    pub const APP_NAME: &str = "ream";
```
KolbyML, line 27: this const should be declared outside of this function

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+///     "finalized" : None,
+///     "data" : T
+/// })
+/// }
+#[derive(Debug, Serialize, Deserialize)]
+pub struct OptionalBeaconVersionedResponse<T> {
+    pub version: Option<String>,
+    #[serde(default)]
+    pub execution_optimistic: Option<Value>,
+    #[serde(default)]
```
KolbyML, line 41:
```suggestion
    pub execution_optimistic: Option<Value>,
```

we don't need to declare this right?

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+    let client = reqwest::Client::new();
+    let state = client
+        .get(format!("{rpc}eth/v2/debug/beacon/states/{slot}"))
+        .header(ACCEPT, HeaderValue::from_static("application/octet-stream"))
+        .send()
+        .await?
+        .bytes()
+        .await?;
+
+    BeaconState::from_ssz_bytes(&state).map_err(|err| anyhow!("{:?}", err))
```
KolbyML, line 101:
```suggestion
    BeaconState::from_ssz_bytes(&state).map_err(|err| anyhow!("{err:?}"))
```

the err should be inlined and we should have a message here not just forwarding the error

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+    let client = reqwest::Client::new();
+    let raw_bytes = client
+        .get(format!("{rpc}eth/v2/beacon/blocks/finalized"))
+        .header(ACCEPT, HeaderValue::from_static("application/octet-stream"))
+        .send()
+        .await?
+        .bytes()
+        .await?;
+
+    SignedBeaconBlock::from_ssz_bytes(&raw_bytes).map_err(|err| anyhow!("{:?}", err))
```
KolbyML, line 115: same comment as above

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+    store: ReamDB,
+    beacon_block_root: B256,
+) -> anyhow::Result<()> {
+    let blob_sidecar = reqwest::get(&format!(
+        "{rpc}eth/v1/beacon/blob_sidecars/{beacon_block_root}"
+    ))
+    .await?
+    .json::<BlobSidercars>()
+    .await
+    .unwrap();
```
KolbyML, line 135: unwrap shouldn't be used

KolbyML (commented):
Left some feedback, the PR is pretty much ready to go but there are a few minor issues still

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+/// {
+///  "data": json!({
+///     "version": Some("electra")
+///     "execution_optimistic" : Some("false"),
+///     "finalized" : None,
+///     "data" : T
+/// })
+/// }
+#[derive(Debug, Serialize, Deserialize)]
+pub struct OptionalBeaconVersionedResponse<T> {
```
KolbyML, line 37:
```suggestion
struct OptionalBeaconVersionedResponse<T> {
```

can you remove the pub for any functions not used outside of this file

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+    let raw_bytes = client
+        .get(format!("{rpc}eth/v2/beacon/blocks/finalized"))
+        .header(ACCEPT, HeaderValue::from_static("application/octet-stream"))
+        .send()
+        .await?
+        .bytes()
+        .await?;
+
+    SignedBeaconBlock::from_ssz_bytes(&raw_bytes)
+        .map_err(|err| anyhow!("Unable to deocde block from ssz byttes: {err:?}"))
```
KolbyML, line 115:
```suggestion
        .map_err(|err| anyhow!("Unable to decode block from ssz bytes: {err:?}"))
```

small spelling issues

KolbyML (approved):
:shipit: PR looks good, thank you for your work. Feel free to merge it after the 2 last concerns are fixed

PR #372, opened 2025-05-04
feat: make NetworkSpec a global, instead of passing it everywhere

KolbyML:

We are going to passing NetworkSpec everywhere, it is something we would only be setting at configuration so we can use a OnceLock, to initialize the global then use it.

Switching to `OnceLock` instead of passing a Arc<> everywhere


Kayden-ML (approved):
Looks good nice work 👍

PR #373, opened 2025-05-05
Sync committee rewards

### LINKED ISSUE #204: Implement /eth/v1/beacon/rewards/sync_committee/{block_id}
https://ethereum.github.io/beacon-APIs/#/Beacon/getSyncCommitteeRewards

aditya172926:

Closes #204 #224

Added logic to calculate sync committee rewards and filter based on the given input of validator index or pubkeys.

An improvement to this implementation can be done to reduce lookup time for fetching the current sync committed validator indices by implementing a cache which maps validator pubkey to its index.

### To-Do

 

crates/rpc/src/handlers/block.rs
```diff
@@ -200,6 +200,103 @@ pub async fn get_beacon_block_from_id(
...
+            .position(|validator| validator.pubkey == *sync_pubkey);
+        match index {
+            Some(index) => sync_committee_indices.push(index),
+            None => {
+                return Err(ApiError::NotFound(format!(
+                    "Failed to find `sync_pubkey` from {sync_pubkey:?}"
+                )));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
KolbyML:
```suggestion
    for (index, validator) in beacon_state.validators.iter().enumerate() {
         if sync_committee.pubkeys.contains(validator.pubkey) {
             sync_committee_indices.push(index)
             sync_commitee.pubkey.remove(validator.pubkey)
         }
    }
    
    if !sync_commitee.is_empty() {
                    return Err(ApiError::NotFound(format!(
                    "Failed to find `sync_pubkey` from {sync_pubkey:?}"
                )));
    }
```

I think we can avoid adding a cache for now. Ideally the outer loop is the `beacon_state.validators` as it is a lot bigger then `sync_committee.pubkeys`

The code change in my suggestion would be O(m) as the m is much much bigger then n. If we want can make `sync_committee.pubkeys` into a `HashSet` so `contains()` becomes `O(1)` instead of `O(n)`

But honestly `O(m)` is probably good enough.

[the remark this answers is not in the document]
aditya172926: > I think we can avoid adding a cache for now. Ideally the outer loop is the `beacon_state.validators` as it is a lot bigger then `sync_committee.pubkeys`
> 
> The code change in my suggestion would be O(m) as the m is much much bigger then n. If we want can make `sync_committee.pubkeys` into a `HashSet` so `contains()` becomes `O(1)` instead of `O(n)`
> 
> But honestly `O(m)` is probably good enough.

Making a hashset seemed like a good idea to go with. Implemented that so that .contains() should be O(1) now.
Can you take a look and review @KolbyML

KolbyML:
Could we refactor the state logic instead of redefining it here. That way the ef-tests test to make sure this code is right was well. Just because this coding being almost duplicated in 2 places is hard to maintain and review to prevent bugs.

aditya172926:
> Could we refactor the state logic instead of redefining it here. That way the ef-tests test to make sure this code is right was well. Just because this coding being almost duplicated in 2 places is hard to maintain and review to prevent bugs.

Are you taking about the function to get beacon state from id? Yeah we can reuse the function defined in state.rs.

KolbyML:
> > Could we refactor the state logic instead of redefining it here. That way the ef-tests test to make sure this code is right was well. Just because this coding being almost duplicated in 2 places is hard to maintain and review to prevent bugs.
> 
> Are you taking about the function to get beacon state from id? Yeah we can reuse the function defined in state.rs.

I mean most of this code can be reused, for example get_sync_aggregate_rewards() is just copied from a function from state.rs

Ideally we refactor the code in state.rs so it can be used here instead of redefined here, as ideally we don't want to main 2 copies of code doing the same thing. Also there are no tests here, so if we refactored the code so we could call it here, then we would know the RPC code is always returning the right result.

aditya172926:
> > > Could we refactor the state logic instead of redefining it here. That way the ef-tests test to make sure this code is right was well. Just because this coding being almost duplicated in 2 places is hard to maintain and review to prevent bugs.
> > 
> > 
> > Are you taking about the function to get beacon state from id? Yeah we can reuse the function defined in state.rs.
> 
> I mean most of this code can be reused, for example get_sync_aggregate_rewards() is just copied from a function from state.rs
> 
> Ideally we refactor the code in state.rs so it can be used here instead of redefined here, as ideally we don't want to main 2 copies of code doing the same thing. Also there are no tests here, so if we refactored the code so we could call it here, then we would know the RPC code is always returning the right result, as the code in state.rs is pretty heavily tested by `ef-tests`

Just to be sure, we are talking to refactor the code in this file right `/src/electra/beacon_state.rs` right?

KolbyML:
> > > > Could we refactor the state logic instead of redefining it here. That way the ef-tests test to make sure this code is right was well. Just because this coding being almost duplicated in 2 places is hard to maintain and review to prevent bugs.
> > > 
> > > 
> > > Are you taking about the function to get beacon state from id? Yeah we can reuse the function defined in state.rs.
> > 
> > 
> > I mean most of this code can be reused, for example get_sync_aggregate_rewards() is just copied from a function from state.rs
> > Ideally we refactor the code in state.rs so it can be used here instead of redefined here, as ideally we don't want to main 2 copies of code doing the same thing. Also there are no tests here, so if we refactored the code so we could call it here, then we would know the RPC code is always returning the right result, as the code in state.rs is pretty heavily tested by `ef-tests`
> 
> Just to be sure, we are talking to refactor the code in this file right `/src/electra/beacon_state.rs` right?

yeah

aditya172926:
Hey @KolbyML i made a refactor commit. Can you review this PR again?

crates/common/consensus/src/electra/beacon_state.rs
```diff
@@ -1593,6 +1627,80 @@ impl BeaconState {
         Ok(())
     }
 
+    pub fn compute_sync_committee_rewards(
+        &self,
+        beacon_block: &SignedBeaconBlock,
+    ) -> anyhow::Result<HashMap<u64, u64>> {
```
KolbyML, line 1633: this function isn't used in `beacon_state.rs` is this code duplicate logic from where else?

crates/common/consensus/src/electra/beacon_state.rs
```diff
@@ -1593,6 +1627,80 @@ impl BeaconState {
...
+
+        let mut balances = HashMap::<usize, u64>::new();
+
+        for validator_index in &sync_committee_indices {
+            balances.insert(
+                *validator_index,
+                *self
+                    .balances
+                    .get(*validator_index)
+                    .ok_or(anyhow!(String::from("Sync Committee Rewards Sync Error",)))?,
```
KolbyML, line 1649:
```suggestion
                    .ok_or(anyhow!("Sync Committee Rewards Sync Error"))?,
```

why is `String::from` here? same for other places in the code

crates/common/consensus/src/electra/beacon_state.rs
aditya172926: I have implemented this function in `handlers/block.rs`, then moved it to the beacon_state.rs in the previous refactor. Should i move it back?

crates/rpc/src/handlers/block.rs
```diff
@@ -43,6 +43,12 @@ pub struct BlockRewards {
     pub attester_slashings: u64,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ValidatorSyncCommitteeReward {
+    pub validator_index: u64,
+    pub reward: u64,
```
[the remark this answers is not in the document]
aditya172926: fixed

crates/common/consensus/src/electra/beacon_state.rs
aditya172926: checked that we don't require it, removed.

crates/common/consensus/src/electra/beacon_state.rs
KolbyML: Most of the code in this function is duplicate logic from `process_sync_aggregate()` right.

Ideally `process_sync_aggregate()` and `compute_sync_committee_rewards()` use the same base function to do the raw calculations.

aditya172926: Makes sense, will do the refactor for this.

unnawut:
@aditya172926 Are you still working on this one? I see there're 3 conflicts rn.

aditya172926:
> @aditya172926 ef-tests should pass before I can continue reviewing the PR

fixed. ef-tests are passing

KolbyML (approved):
:shipit: the PR is good

PR #398, opened 2025-05-13
feat: add multiproof feature in `ream-merkle`

syjn99:

 

This PR adds a [multiproof](https://github.com/ethereum/consensus-specs/blob/dev/ssz/merkle-proofs.md#merkle-multiproofs) feature in our merkle package, `ream-merkle`. This feature will be used in passing `BeaconState` partially to the zkVM.

For general understanding, reading [my note](https://hackmd.io/@reamlabs/B1SJuEoxle#Merkle-Multiproof) might be helpful. Also, you can see an [example](https://hackmd.io/@reamlabs/B1SJuEoxle#Example-first) of using this multiproof.

It modifies slightly the code from [`ssz_rs`](https://github.com/ralexstokes/ssz-rs/blob/main/ssz-rs/src/merkleization/multiproofs.rs) which implements the Python-like [spec code](https://github.com/ethereum/consensus-specs/blob/dev/ssz/merkle-proofs.md#merkle-multiproofs) into Rust.

### To-Do

 
### DIFF
--- a/crates/crypto/merkle/Cargo.toml
+++ b/crates/crypto/merkle/Cargo.toml
@@ -13,6 +13,7 @@ version.workspace = true
 alloy-primitives.workspace = true
 anyhow.workspace = true
 ethereum_hashing.workspace = true
+serde.workspace = true
 
 [features]
 default = []
--- a/crates/crypto/merkle/Cargo.toml
+++ b/crates/crypto/merkle/Cargo.toml
@@ -14,7 +14,3 @@ alloy-primitives.workspace = true
 anyhow.workspace = true
 ethereum_hashing.workspace = true
 serde.workspace = true
-
-[features]
-default = []
-multiproof = []
--- a/crates/crypto/merkle/Cargo.toml
+++ b/crates/crypto/merkle/Cargo.toml
@@ -13,3 +13,7 @@ version.workspace = true
 alloy-primitives.workspace = true
 anyhow.workspace = true
 ethereum_hashing.workspace = true
+
+[features]
+default = []
+multiproof = []
--- a/crates/crypto/merkle/src/hash.rs
+++ b/crates/crypto/merkle/src/hash.rs
@@ -1,6 +1,6 @@
 use alloy_primitives::B256;
 
 /// Common hashing function for Merkle trees.
-pub(crate) fn hash_concat(h1: &[u8], h2: &[u8]) -> B256 {
+pub fn hash_concat(h1: &[u8], h2: &[u8]) -> B256 {
     ethereum_hashing::hash32_concat(h1, h2).into()
 }
--- a/crates/crypto/merkle/src/hash.rs
+++ b/crates/crypto/merkle/src/hash.rs
@@ -1,0 +1,6 @@
+use alloy_primitives::B256;
+
+/// Common hashing function for Merkle trees.
+pub(crate) fn hash_concat(h1: &[u8], h2: &[u8]) -> B256 {
+    ethereum_hashing::hash32_concat(h1, h2).into()
+}
--- a/crates/crypto/merkle/src/index.rs
+++ b/crates/crypto/merkle/src/index.rs
@@ -1,0 +1,31 @@
+/// ``LeafIndex`` is the index of a leaf in the **bottom** layer of the ``tree``.
+pub(crate) type LeafIndex = u64;
+
+/// ``GeneralizedIndex`` is the index of a node in the ``tree``.
+pub(crate) type GeneralizedIndex = u64;
+
+/// Return the given bit of a generalized index.
+/// Note: It is fine to pass ``LeafIndex`` to this function,
+/// as the result will be the same.
+pub(crate) fn get_generalized_index_bit(index: GeneralizedIndex, position: u64) -> bool {
+    (index & (1 << position)) > 0
+}
+
+pub(crate) fn generalized_index_child(
+    index: GeneralizedIndex,
+    right_side: bool,
+) -> GeneralizedIndex {
+    index * 2 + right_side as GeneralizedIndex
+}
+
+pub(crate) fn get_subtree_index(generalized_index: GeneralizedIndex) -> LeafIndex {
+    generalized_index % (1 << (generalized_index as f64).log2().floor() as u64)
+}
+
+/// Return the generalized index of the leaf index with ``depth``.
+pub(crate) fn generalized_index_from_leaf_index(
+    leaf_index: LeafIndex,
+    depth: u64,
+) -> GeneralizedIndex {
+    leaf_index + (1 << depth)
+}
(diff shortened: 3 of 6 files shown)


crates/crypto/merkle/Cargo.toml
```diff
@@ -13,3 +13,8 @@ version.workspace = true
 alloy-primitives.workspace = true
 anyhow.workspace = true
 ethereum_hashing.workspace = true
+serde.workspace = true
+
+[features]
+default = []
+multiproof = []
```
KolbyML, line 20: I don't think this needs to be behind a feature flag

crates/crypto/merkle/src/index.rs
```diff
@@ -0,0 +1,120 @@
+/// ``GeneralizedIndex`` is the index of a node in the ``tree``.
+pub(crate) type GeneralizedIndex = u64;
+
+/// Return the given bit of a generalized index.
+/// Note: It is fine to pass ``LeafIndex`` to this function,
+/// as the result will be the same.
+pub(crate) fn get_generalized_index_bit(index: GeneralizedIndex, position: u64) -> bool {
```
KolbyML, line 7:
```suggestion
pub fn get_generalized_index_bit(index: GeneralizedIndex, position: u64) -> bool {
```

same for everywhere else, we don't need to declare `pub(crate)` if we don't want to export these functions outside the crate we can just not make the mod for index.rs public in lib.rs

crates/crypto/merkle/src/lib.rs
```diff
@@ -118,15 +111,12 @@ mod tests {
...
         ];
 
-        let depth = (leaves.len() as f64).log2().floor() as u64;
+        let depth = (leaves.len() as f64).log2().ceil() as u64;
 
-        let node_2: B256 =
-            ethereum_hashing::hash32_concat(leaves[0].as_slice(), leaves[1].as_slice()).into();
-        let node_3: B256 =
-            ethereum_hashing::hash32_concat(leaves[2].as_slice(), leaves[3].as_slice()).into();
+        let node_2: B256 = hash_concat(leaves[0].as_slice(), leaves[1].as_slice());
```
KolbyML, line 116:
```suggestion
        let node_2 = hash_concat(leaves[0].as_slice(), leaves[1].as_slice());
```

same with below we no longer need these types declarations

crates/crypto/merkle/src/lib.rs
```diff
@@ -145,25 +135,25 @@ mod tests {
...
             &proof_2,
-            2 * depth + 2,
+            2 + (1 << depth),
             root
         ));
         assert!(is_valid_normalized_merkle_branch(
             leaves[3],
             &proof_3,
-            2 * depth + 3,
+            3 + (1 << depth),
```
KolbyML, line 156: These are tests so I think readability is nice, I don't think there really was a reason to do change this right?

crates/crypto/merkle/src/index.rs
```diff
@@ -0,0 +1,120 @@
...
+            assert_eq!(helper_indices[3], 4);
+        }
+    }
+}
+
+#[cfg(feature = "multiproof")]
+pub use multiproof_helpers::{
+    generalized_index_from_leaf_index, generalized_index_parent, generalized_index_sibling,
+    get_helper_indices,
+};
```
KolbyML, line 120: why are import declarations at the bottom of the file?

crates/crypto/merkle/Cargo.toml
syjn99: Yes, then I will delete this `features` section and all `cfg()` lines.

crates/crypto/merkle/src/index.rs
syjn99: My bad, and this will be removed because `features = "multiproof"` will be removed.

crates/crypto/merkle/src/hash.rs
```diff
@@ -0,0 +1,6 @@
+use alloy_primitives::B256;
+
+/// Common hashing function for Merkle trees.
+pub(crate) fn hash_concat(h1: &[u8], h2: &[u8]) -> B256 {
```
KolbyML, line 4:
```suggestion
pub fn hash_concat(h1: &[u8], h2: &[u8]) -> B256 {
```

same reason as before

crates/crypto/merkle/src/index.rs
```diff
@@ -0,0 +1,104 @@
+use std::collections::HashSet;
+
+/// ``GeneralizedIndex`` is the index of a node in the ``tree``.
+pub type GeneralizedIndex = u64;
```
KolbyML, line 4: I think we should just use u64 directly, instead of defining a new type like this, which is a pattern we have done in all other parts of the codebase

crates/crypto/merkle/src/lib.rs
KolbyML: Sounds good, what do you think of moving the addition to the other side then? so `(1 << depth) + 3` I find it a little easier to read, up to you though it isn't the biggest deal.

crates/crypto/merkle/src/multiproof.rs
```diff
@@ -0,0 +1,158 @@
...
+            .map(|&g| (g, tree[g as usize]))
+            .collect();
+
+        Ok(Self { leaves, proofs })
+    }
+
+    /// Return the root of the multiproof.
+    ///
+    /// Most code in this function is borrowed from ssz_rs crate.
+    /// https://github.com/ralexstokes/ssz-rs/blob/main/ssz-rs/src/merkleization/multiproofs.rs
```
KolbyML, line 62: Could we just implement this function from the specs instead of borrowing code from ssz_rs, as I think it is better if we build independently.

crates/crypto/merkle/src/index.rs
syjn99: Make sense, I'm gonna explain more on this at the function comment.

crates/crypto/merkle/src/index.rs
```diff
@@ -0,0 +1,98 @@
...
+    index ^ 1
+}
+
+pub fn generalized_index_child(index: u64, right_side: bool) -> u64 {
+    index * 2 + right_side as u64
+}
+
+pub const fn generalized_index_parent(index: u64) -> u64 {
+    index / 2
+}
```
KolbyML, line 17: do these functions take `generalized_index` or `index`?

syjn99: As the function name implies, it takes `generalized_index` normally.

KolbyML: I meant I think we should call the argument what it is, instead of assuming what it is that from the function name. As we could have a function which takes a normal index for example

syjn99: Yes, this was the main reason for type aliasing like GeneralizedIndex as it is quite confusing. I think renaming parameters(to generalized_index) can be enough then!

PR #406, opened 2025-05-17
feat: fix Beacon API content type

### LINKED ISSUE #405: Fix Beacon API content type
### Describe the feature

Some beacon APIs in Ream have Json encoding for paramaters which means they need to be passed in the Body section. In reality, no getter beacon APIs have Body input requirement(I still need recheck this). They should all be passed in the query itself

### Additional context

_No response_

varun-doshi:

Fixes #405

### Details:

Some `GET` endpoints were using `Json` as their input which meant they need be passed in the `body` section of the request. As per spec, these endpoints should only have `Path` and `Query` types in the request i.e. the endpoints should look something like this:
```
http://localhost/eth/v1/beacon/blob_sidecars/head?indices=1&indices=4
```
But these were failing in Ream since they were in `Json` format. 
Solely switching from `Json` to `Query` type in `actix-web` does not do the trick because `actix` does not support duplicate name keys in the query as per this [discussion](https://github.com/actix/actix-web/issues/1301) .

Hence the only way to format it as per spec is to implement a custom deserializer ourselves(which I've done and changed it to a macro for ease of use).

Note: This issue pertains only to `GET` endpoints. `POST` endpoints generally require `body` so thats okay

### Testing: 

I've tested the updated implementations and they are all working as expected now.

### DIFF
--- a/crates/rpc/src/handlers/blob_sidecar.rs
+++ b/crates/rpc/src/handlers/blob_sidecar.rs
@@ -1,7 +1,8 @@
 use actix_web::{
     HttpResponse, Responder, get,
-    web::{Data, Path, Query},
+    web::{Data, Path},
 };
+use actix_web_lab::extract::Query;
 use ream_consensus::blob_sidecar::BlobIdentifier;
 use ream_storage::{db::ReamDB, tables::Table};
 use tracing::error;
--- a/crates/rpc/src/handlers/blob_sidecar.rs
+++ b/crates/rpc/src/handlers/blob_sidecar.rs
@@ -27,8 +27,7 @@ pub async fn get_blob_sidecars(
         for index in indices {
             if index >= &max_index {
                 return Err(ApiError::BadRequest(format!(
-                    "Invalid blob index: {index}, max index is {}",
-                    max_index
+                    "Invalid blob index: {index}, max index is {max_index}"
                 )));
             }
         }
--- a/crates/rpc/src/handlers/blob_sidecar.rs
+++ b/crates/rpc/src/handlers/blob_sidecar.rs
@@ -1,6 +1,6 @@
 use actix_web::{
     HttpResponse, Responder, get,
-    web::{Data, Json, Path},
+    web::{Data, Path, Query},
 };
 use ream_consensus::blob_sidecar::BlobIdentifier;
 use ream_storage::{db::ReamDB, tables::Table};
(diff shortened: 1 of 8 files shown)


KolbyML:
> Solely switching from `Json` to `Query` type in `actix-web` does not do the trick because `actix` does not support duplicate name keys in the query as per this [discussion](https://github.com/actix/actix-web/issues/1301) .
> 
> Hence the only way to format it as per spec is to implement a custom deserializer ourselves(which I've done and changed it to a macro for ease of use).

Why would we want to support duplicate keys?

varun-doshi:
> > Solely switching from `Json` to `Query` type in `actix-web` does not do the trick because `actix` does not support duplicate name keys in the query as per this [discussion](https://github.com/actix/actix-web/issues/1301) .
> > Hence the only way to format it as per spec is to implement a custom deserializer ourselves(which I've done and changed it to a macro for ease of use).
> 
> Why would we want to support duplicate keys?

this is what i meant:
```
http://localhost/eth/v1/beacon/blob_sidecars/head?indices=1&indices=4
```


Its how it is in the official specs

KolbyML:
> > > Solely switching from `Json` to `Query` type in `actix-web` does not do the trick because `actix` does not support duplicate name keys in the query as per this [discussion](https://github.com/actix/actix-web/issues/1301) .
> > > Hence the only way to format it as per spec is to implement a custom deserializer ourselves(which I've done and changed it to a macro for ease of use).
> > 
> > 
> > Why would we want to support duplicate keys?
> 
> this is what i meant:
> 
> ```
> http://localhost/eth/v1/beacon/blob_sidecars/head?indices=1&indices=4
> ```
> 
> Its how it is in the official specs

Can we remove the custom macro and just use
https://github.com/actix/actix-web/issues/1301#issuecomment-1873831135

image

varun-doshi:
> > > > Solely switching from `Json` to `Query` type in `actix-web` does not do the trick because `actix` does not support duplicate name keys in the query as per this [discussion](https://github.com/actix/actix-web/issues/1301) .
> > > > Hence the only way to format it as per spec is to implement a custom deserializer ourselves(which I've done and changed it to a macro for ease of use).
> > > 
> > > 
> > > Why would we want to support duplicate keys?
> > 
> > 
> > this is what i meant:
> > ```
> > http://localhost/eth/v1/beacon/blob_sidecars/head?indices=1&indices=4
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Its how it is in the official specs
> 
> Can we remove the custom macro and just use [actix/actix-web#1301 (comment)](https://github.com/actix/actix-web/issues/1301#issuecomment-1873831135)
> 
> image

looking into this

varun-doshi:
@KolbyML ready for another look

crates/rpc/src/handlers/blob_sidecar.rs
```diff
@@ -27,7 +28,7 @@ pub async fn get_blob_sidecars(
             if index >= &max_index {
                 return Err(ApiError::BadRequest(format!(
                     "Invalid blob index: {index}, max index is {}",
-                    max_index - 1
+                    max_index
```
KolbyML, line 31: can you inline max_index

Issue #400, opened 2025-05-16
Implement get_eth1_pending_deposit_count


Patchoulis:
### Describe the feature

Implement get_eth1_pending_deposit_count from the electra spec: https://ethereum.github.io/consensus-specs/specs/electra/validator/

PR #415, opened 2025-05-19
Adding get_eth1_pending_deposit_count

Patchoulis:

Fixes https://github.com/ReamLabs/ream/issues/400

Originally I was planning to add it to the new validater crate, but since all the input was sourced from the beacon state, I ended up adding it as a method for the beacon states themselves.

### To-Do

We definitely need more test coverage for the beacon states in general, so a followup would likely be just going through the methods we currently implement for them, and writing tests.


KolbyML (approved):
:shipit: looks good

syjn99 (approved):

PR #419, opened 2025-05-21
chore: improve checkpoint sync logs

Vid201:

Improve logging during checkpoint sync. The state can take some time to download and at the moment it feels like the Ream node freezes.

Added some logs.

### DIFF
--- a/crates/common/checkpoint_sync/src/lib.rs
+++ b/crates/common/checkpoint_sync/src/lib.rs
@@ -47,7 +47,7 @@ pub async fn initialize_db_from_checkpoint(
     checkpoint_sync_url: Option<Url>,
 ) -> anyhow::Result<()> {
     if db.is_initialized() {
-        warn!("Starting checkpoint sync from existing DB. It is advised to start from a clean DB");
+        warn!("DB is already initialized. Skipping checkpoint sync.");
         return Ok(());
     }
 
--- a/crates/common/checkpoint_sync/src/lib.rs
+++ b/crates/common/checkpoint_sync/src/lib.rs
@@ -47,29 +47,34 @@ pub async fn initialize_db_from_checkpoint(
     checkpoint_sync_url: Option<Url>,
 ) -> anyhow::Result<()> {
     if db.is_initialized() {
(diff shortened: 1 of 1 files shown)


varun-doshi (approved):
LGTM

crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -47,29 +47,34 @@ pub async fn initialize_db_from_checkpoint(
     checkpoint_sync_url: Option<Url>,
 ) -> anyhow::Result<()> {
     if db.is_initialized() {
-        warn!("Starting Checkpoint Sync from existing DB. It is advised to start from a clean DB");
+        warn!("Starting checkpoint sync from existing DB. It is advised to start from a clean DB");
```
KolbyML, line 50:
```suggestion
        debug!("Checkpoint sync is already initialized");
```

This log doesn't make sense, as Checkpoint sync doesn't run if the database is already initialized

varun-doshi: We still need to sync from the existing db by querying peers to reach close to the head no? 

So if db already exists maybe we can say something accordingly?

In my mind, checkpoint sync is a two part process
1. Download initial state
2. Progress intial state to the head


@KolbyML what do you think

unnawut: I propose:

```suggestion
        warn!("DB is already initialized. Skipping checkpoint sync.");
```

KolbyML: > We still need to sync from the existing db by querying peers to reach close to the head no?
> 
> So if db already exists maybe we can say something accordingly?
> 
> In my mind, checkpoint sync is a two part process
> 
> 1. Download initial state
> 2. Progress intial state to the head
> 
> @KolbyML what do you think

2. this isn't apart of checkpoint sync, this is just backfill sync from our latest state available. Checkpoint sync is just how we initialize a state if we have none

KolbyML: > I propose this. I think it means exactly what this part of the code does.

This suggestion is good

KolbyML (approved):
:shipit: looks good

Issue #422, opened 2025-05-22
Implement get_head() and get_filtered_block_tree() functions


Kayden-ML:
### Describe the feature

https://ethereum.github.io/consensus-specs/specs/phase0/fork-choice/#get_filtered_block_tree
https://ethereum.github.io/consensus-specs/specs/phase0/fork-choice/#get_head

PR #423, opened 2025-05-22
feat: implemented get_head() and get_filtered_block_tree()

Kayden-ML:

Fixes: https://github.com/ReamLabs/ream/issues/422

https://ethereum.github.io/consensus-specs/specs/phase0/fork-choice/#get_filtered_block_tree
https://ethereum.github.io/consensus-specs/specs/phase0/fork-choice/#get_head

### DIFF
--- a/crates/common/fork_choice/src/store.rs
+++ b/crates/common/fork_choice/src/store.rs
@@ -1,3 +1,5 @@
+use std::cmp::Ordering;
+
 use alloy_primitives::{B256, map::HashMap};
 use anyhow::{anyhow, bail, ensure};
 use ream_bls::BLSSignature;
@@ -138,11 +140,8 @@ impl Store {
     /// whose leaf state's justified/finalized info agrees with that in ``store``.
     pub fn get_filtered_block_tree(&self) -> anyhow::Result<HashMap<B256, BeaconBlock>> {
         let base = self.db.justified_checkpoint_provider().get()?.root;
-
-        let mut blocks: HashMap<B256, BeaconBlock> = HashMap::default();
-
+        let mut blocks = HashMap::default();
         self.filter_block_tree(base, &mut blocks)?;
-
         Ok(blocks)
     }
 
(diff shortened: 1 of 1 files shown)


crates/common/fork_choice/src/store.rs
```diff
@@ -134,6 +134,56 @@ impl Store {
...
+    /// whose leaf state's justified/finalized info agrees with that in ``store``.
+    pub fn get_filtered_block_tree(&self) -> anyhow::Result<HashMap<B256, BeaconBlock>> {
+        let base = self.db.justified_checkpoint_provider().get()?.root;
+
+        let mut blocks: HashMap<B256, BeaconBlock> = HashMap::default();
+
+        self.filter_block_tree(base, &mut blocks)?;
+
+        Ok(blocks)
+    }
```
KolbyML, line 147:
```suggestion
    pub fn get_filtered_block_tree(&self) -> anyhow::Result<HashMap<B256, BeaconBlock>> {
        let base = self.db.justified_checkpoint_provider().get()?.root;
        let mut blocks = HashMap::default();
        self.filter_block_tree(base, &mut blocks)?;
        Ok(blocks)
    }
```

we are fine without the type decoration and the code is simple enough where spaces aren't needed

crates/common/fork_choice/src/store.rs
```diff
@@ -134,6 +134,56 @@ impl Store {
...
+                })
+                .max_by_key(|child| {
+                    let weight = self.get_weight(*child).unwrap_or_default();
+                    (weight, *child)
+                })
+            {
+                head = new_head;
+            } else {
+                return Ok(head);
+            }
```
KolbyML, line 183:
```suggestion
            // Sort by latest attesting balance with ties broken lexicographically
            children.sort_by(|a, b| {
                let weight_a = get_weight(store, a)?;
                let weight_b = get_weight(store, b)?;
                match weight_a.cmp(&weight_b) {
                    Ordering::Equal => a.cmp(b),
                    other => other,
                }
            });   
                     
            // Ties broken by favoring block with lexicographically higher root
            let Some(best_child) = children.last() else {
                bail!("Children should always be present");
            }
            
            head = best_child;
```

I think there we a logic error here, the suggestion above should fix it

KolbyML (approved):
:shipit: PR looks good feel free to merge it when the concerns are resolved

crates/common/fork_choice/src/store.rs
KolbyML: The lambda in the psudo code is a little tricky to read to be honest

KolbyML (approved):
:shipit: looks good

PR #429, opened 2025-05-23
chore: make cargo sort check workspace

KolbyML:
There is a bug and --group isn't working https://github.com/DevinR528/cargo-sort/issues/104

Currently cargo sort is only checking the parent Cargo.toml

Using the ``--workspace`` flag


varun-doshi (approved):
Lgtm

KolbyML:
> Lgtm
> 
> Although looks like currently sort is failing on CI?

I think you shouldn't review a PR if it is in draft tbh

PR #434, opened 2025-05-23
feat: add structure for NetworkManager + BlockByRange syncer

KolbyML:

We want to be able to 
- handle inbound + outbound P2P messages
- make requests to the P2P layer for data so we can sync
- be able to handle on gossip events

There are a lot of people working on the P2P layer right now, this is the foundation so we can all work in parallel. Nothing is set in stone, but this should be enough so we don't bud heads. Normally I would propose this PR when I have a lot more done, but due to splitting work there is a high amount of value getting this in now 

- Create a NetworkManager service which basically coordinates the operations of the Beacon Node
- Add BeaconChain an object which manages the chain
- Syncer crate where we will implement syncers 
- I added the interop between the NetworkManager/Syncers/P2P Layer

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4405,15 +4405,27 @@ dependencies = [
  "ream-consensus",
  "ream-discv5",
  "ream-executor",
+ "ream-manager",
  "ream-network-spec",
  "ream-node",
  "ream-p2p",
  "ream-rpc",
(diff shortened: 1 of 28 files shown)


Kayden-ML (approved):
Looks good nice work 👍

PR #450, opened 2025-05-25
ci: require conventional commit to pass merge queue

KolbyML:

A few PR's not using conventional commits in the title got through

by making the conventional CI pass for the PR to be able to be merge. We should update systems to prevent mistakes like this


Kayden-ML (approved):
Looks good 👍

Issue #421, opened 2025-05-21
Implement get_eth1_vote


Patchoulis:
### Describe the feature

One of the things missing from the validator crate that needs to be done is the get_eth1_vote function from the [electra spec](https://ethereum.github.io/consensus-specs/specs/electra/validator/)

From the looks of it, since the spec shows the function using quite a bit of the field data from BeaconState, it'd be best to add it as a new method for it.

Patchoulis:
Could I work on this?

PR #456, opened 2025-05-25
feat: Added get_eth1_vote

Patchoulis:

fixes https://github.com/ReamLabs/ream/issues/421

Ended up adding it as a beacon state method due to the fields it uses.


KolbyML (approved):
:shipit: looks good

PR #459, opened 2025-05-25
feat: implement sync_committee api

### LINKED ISSUE #192: Implement /eth/v1/beacon/states/{state_id}/sync_committees
https://ethereum.github.io/beacon-APIs/#/Beacon/getEpochSyncCommittees

varun-doshi:

Closes #192 

Implements spec https://ethereum.github.io/beacon-APIs/#/Beacon/getEpochSyncCommittees

### DIFF
--- a/crates/rpc/src/handlers/state.rs
+++ b/crates/rpc/src/handlers/state.rs
@@ -3,7 +3,10 @@ use actix_web::{
     web::{Data, Path, Query},
 };
 use alloy_primitives::B256;
-use ream_consensus::{checkpoint::Checkpoint, electra::beacon_state::BeaconState};
+use ream_consensus::{
+    checkpoint::Checkpoint, constants::SYNC_COMMITTEE_SIZE, electra::beacon_state::BeaconState,
+    misc::compute_start_slot_at_epoch,
+};
 use ream_storage::{
     db::ReamDB,
     tables::{Field, Table},
@@ -18,6 +21,8 @@ use crate::types::{
     query::EpochQuery,
     response::{BeaconResponse, BeaconVersionedResponse},
 };
+
+pub const SYNC_COMMITTEE_SUBNET_COUNT: u64 = 4;
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct CheckpointData {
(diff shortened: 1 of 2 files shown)


crates/rpc/src/handlers/state.rs
```diff
@@ -208,3 +219,44 @@ pub async fn get_pending_partial_withdrawals(
...
+pub async fn get_sync_committees(
+    db: Data<ReamDB>,
+    state_id: Path<ID>,
+    epoch: Query<EpochQuery>,
+) -> Result<impl Responder, ApiError> {
+    let state = if let Some(epoch) = epoch.epoch {
+        get_state_from_id(ID::Slot(compute_start_slot_at_epoch(epoch)), &db).await?
+    } else {
+        get_state_from_id(state_id.into_inner(), &db).await?
+    };
```
KolbyML, line 236: I am not sure if this is implemented correctly

image


It feels off that, epoch would override a required parameter like this, do other clients do this?

varun-doshi: Fixed the logic

crates/rpc/Cargo.toml
```diff
@@ -20,6 +20,7 @@ hashbrown.workspace = true
 parking_lot.workspace = true
 serde.workspace = true
 serde_json.workspace = true
+serde_with.workspace = true
```
KolbyML, line 23: I think we should avoid adding serde_with as we don't really need it

crates/rpc/src/handlers/state.rs
```diff
@@ -51,6 +57,15 @@ impl RandaoResponse {
     }
 }
 
+#[serde_as]
+#[derive(Serialize, Deserialize)]
+struct SyncCommitteeResponse {
+    #[serde(with = "serde_utils::quoted_u64_vec")]
+    pub validators: Vec<u64>,
+    #[serde_as(as = "Vec<Vec<DisplayFromStr>>")]
+    pub validator_aggregates: Vec<Vec<u64>>,
```
KolbyML, line 66: instead of Vec<Vec<u64>> we can define a type Vec<X>

serde(transparent)
struct X{
   serde with u64 ver
   x: Vec<u64>
}

KolbyML (approved):
:shipit: looks great good work 👍

PR #460, opened 2025-05-26
feat: implement get_aggregate_and_proof and get_aggregate_and_proof_signature

### LINKED ISSUE #413: Implement Attestation aggregation functions
### Describe the feature

Implement functions required for Attestation aggregation as per the Altair Honest Validator Spec:
https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/validator.md#aggregation-selection

This will include the following functions:
get_slot_signature
is_aggregator
get_aggregate_signature
get_aggregate_and_proof
get_aggregate_and_proof_signature

### Additional context

_No response_

shariqnaiyer:

Fixes #413

Implemented the get_aggregate_and_proof and get_aggregate_and_proof_signature functions as per the Phase0 Honest Validator Spec:
https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/validator.md#broadcast-aggregate

### To-Do

 
### DIFF
--- a/crates/common/validator/src/aggregate_and_proof.rs
+++ b/crates/common/validator/src/aggregate_and_proof.rs
@@ -1,12 +1,47 @@
-use ream_bls::BLSSignature;
-use ream_consensus::attestation::Attestation;
+use ream_bls::{BLSSignature, PrivateKey, traits::Signable};
+use ream_consensus::{
+    attestation::Attestation,
+    constants::DOMAIN_AGGREGATE_AND_PROOF,
+    electra::beacon_state::BeaconState,
+    misc::{compute_epoch_at_slot, compute_signing_root},
+};
 use serde::{Deserialize, Serialize};
 use ssz_derive::{Decode, Encode};
 use tree_hash_derive::TreeHash;
+
+use crate::attestation::get_slot_signature;
 
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
 pub struct AggregateAndProof {
     pub aggregator_index: u64,
     pub aggregate: Attestation,
     pub selection_proof: BLSSignature,
+}
+
+pub fn get_aggregate_and_proof(
+    state: &BeaconState,
+    aggregator_index: u64,
+    aggregate: Attestation,
+    private_key: PrivateKey,
+) -> anyhow::Result<AggregateAndProof> {
+    Ok(AggregateAndProof {
+        aggregator_index,
+        aggregate: aggregate.clone(),
+        selection_proof: get_slot_signature(state, aggregate.data.slot, private_key)?,
+    })
+}
+
+pub fn get_aggregate_and_proof_signature(
+    state: &BeaconState,
+    aggregate_and_proof: AggregateAndProof,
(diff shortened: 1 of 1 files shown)


crates/common/validator/src/aggregate_and_proof.rs
```diff
@@ -1,12 +1,48 @@
...
+    aggregator_index: u64,
+    aggregate: Attestation,
+    private_key: PrivateKey,
+) -> anyhow::Result<AggregateAndProof> {
+    let slot = aggregate.data.slot;
+    Ok(AggregateAndProof {
+        aggregator_index,
+        aggregate,
+        selection_proof: get_slot_signature(state, slot, private_key)?,
+    })
```
syjn99, line 32:
```suggestion
    Ok(AggregateAndProof {
        aggregator_index,
        aggregate,
        selection_proof: get_slot_signature(state, aggregate.data.slot, private_key)?,
    })
```

I think it will be enough to access directly like this

syjn99 (approved):
LGTM, just one nitpick

crates/common/validator/src/aggregate_and_proof.rs
shariqnaiyer: @syjn99 I initially did this but was finding the aggregate has moved by the time we do aggregate.data.slot. My options were to either clone aggregate or to use a variable prior to the move. I am open to suggestions if there is any other way to handle this.

syjn99: Hmm I see, I'm not sure where this function will be used but it seems passing `Attestation` as a reference, and then clone it is more reasonable IMO.

```RUST
pub fn get_aggregate_and_proof(
    state: &BeaconState,
    aggregator_index: u64,
    aggregate: &Attestation,
    private_key: PrivateKey,
) -> anyhow::Result<AggregateAndProof> {
    Ok(AggregateAndProof {
        aggregator_index,
        aggregate: aggregate.clone(),
        selection_proof: get_slot_signature(state, aggregate.data.slot, private_key)?,
    })
}
```

crates/common/validator/src/aggregate_and_proof.rs
```diff
@@ -1,12 +1,48 @@
...
+}
```
KolbyML:
```suggestion
pub fn get_aggregate_and_proof(
    state: &BeaconState,
    aggregator_index: u64,
    aggregate: Attestation,
    private_key: PrivateKey,
) -> anyhow::Result<AggregateAndProof> {
    Ok(AggregateAndProof {
         selection_proof: get_slot_signature(state, slot, private_key)?,
        aggregator_index,
        aggregate,
    })
}
```

crates/common/validator/src/aggregate_and_proof.rs
shariqnaiyer: @KolbyML @syjn99 The fix was to change the order of elements. I have now fixed it.

crates/common/validator/src/aggregate_and_proof.rs
syjn99: Seems better 👍

Issue #473, opened 2025-05-26
Add CLI for Validator Initialization


Patchoulis:
### Describe the feature

One of the things that need to be done is adding some CLI in order to initialize the validator.

Patchoulis:
Could I be assigned this?

PR #474, opened 2025-05-26
feat: added CLI for validator initialization

Patchoulis:

fixes https://github.com/ReamLabs/ream/issues/473

I added the initial CLI to create a validator config and HTTP client using the info from it.
Since later we plan on adding a key manager server, I added some arguments for that which would be used in the future.

### To-Do

Add a way to serve the key manager api using the information we save in our config.


Patchoulis:
Wait, hold on, I need to fix this up

KolbyML (approved):
:shipit: looks great thank you for setting this up

Issue #462, opened 2025-05-26
Implement engine api eth_blockNumber


Kayden-ML:
### Describe the feature

Implement engine api eth_blockNumber

https://github.com/ethereum/execution-apis/blob/main/src/engine/common.md

PR #476, opened 2025-05-27
feat: Implement engine api eth_blockNumber

Kayden-ML:

Fixes: https://github.com/ReamLabs/ream/issues/462

https://github.com/ethereum/execution-apis/blob/main/src/engine/common.md

### DIFF
--- a/crates/common/execution_engine/src/lib.rs
+++ b/crates/common/execution_engine/src/lib.rs
@@ -120,6 +120,24 @@ impl ExecutionEngine {
             .to_result()
     }
 
+    pub async fn eth_block_number(&self) -> anyhow::Result<String> {
+        let request_body = JsonRpcRequest {
+            id: 1,
+            jsonrpc: "2.0".to_string(),
+            method: "eth_blockNumber ".to_string(),
+            params: vec![],
+        };
+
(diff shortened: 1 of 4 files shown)


crates/common/execution_engine/src/lib.rs
```diff
@@ -120,6 +120,24 @@ impl ExecutionEngine {
...
+            method: "eth_blockNumber ".to_string(),
+            params: vec![],
+        };
+
+        let http_post_request = self.build_request(request_body)?;
+
+        self.http_client
+            .execute(http_post_request)
+            .await?
+            .json::<JsonRpcResponse<String>>()
```
KolbyML, line 136:
```suggestion
    pub async fn eth_block_number(&self) -> anyhow::Result<B64> {
        let request_body = JsonRpcRequest {
            id: 1,
            jsonrpc: "2.0".to_string(),
            method: "eth_blockNumber".to_string(),
            params: vec![],
        };

        let http_post_request = self.build_request(request_body)?;

        self.http_client
            .execute(http_post_request)
            .await?
            .json::<JsonRpcResponse<B64>>()
```

you have an extra space, and we might as well use B64 here

Issue #433, opened 2025-05-23
Connect /eth/v1/events to the validator


Patchoulis:
### Describe the feature

Essentially just connecting the validator to the beacon endpoint for events.

Might require other functionality to be done alongside it.

Patchoulis:
Could I be assigned this?

PR #480, opened 2025-05-28
feat: added event stream

Patchoulis:

fixes https://github.com/ReamLabs/ream/issues/433

Essentially this adds an event stream from the events endpoint for beacon nodes. When I was researching this, I really only found two types of events that would be valuable. Specifically, chain re_orgs, which we would need to listen to, and potentially voluntary exits, which might come in handy for slashing protections.


crates/common/validator/src/beacon_api_client/event.rs
```diff
@@ -0,0 +1,72 @@
...
+        } else if event.event_type == BeaconNodeEventTopic::Attestation.query_value() {
+            Self::from_json(&event.data, Self::Attestation)
+        } else {
+            Err(Self::Error::custom(format!(
+                "Can't create BeaconEvent: unexpected event type: {}",
+                event.event_type,
+            )))
+        }
+    }
+}
```
KolbyML, line 72: Instead of using `BeaconNodeEventTopic` lets just define our own enum called ``EventTopic`` which implements the traits `FromStr` and `Display`, that way we can use a match case, which looks a lot nicer

KolbyML (approved):
:shipit: looks good

PR #485, opened 2025-05-30
feat: added keystore serialization and deserialization

Patchoulis:

Added serialization, deserialization, and tests for keystores.

Ended up running https://github.com/ethereum/staking-deposit-cli to generate a keystore json, and added it as a test.

### To-Do

Using the kdf parameters, we would need to implement password validation, as well as decryption to get a private key.


crates/crypto/keystore/src/encrypted_keystore.rs
```diff
@@ -0,0 +1,206 @@
...
+        #[serde(serialize_with = "buffer_to_hex", deserialize_with = "hex_to_buffer")]
+        iv: Vec<u8>,
+    },
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+#[serde(tag = "function", content = "params")]
+pub enum ChecksumParams {
+    #[serde(rename = "sha256")]
+    Sha256 {},
```
KolbyML, line 69:
```suggestion
#[serde(tag = "function", content = "params", rename_all = "kebab-case")]
pub enum ChecksumParams {
    Sha256 {},
```

crates/crypto/keystore/src/encrypted_keystore.rs
```diff
@@ -0,0 +1,206 @@
...
+impl EncryptedKeystore {
+    pub fn load_from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let file_contents = fs::read_to_string(path)?;
+        let keystore: EncryptedKeystore = serde_json::from_str(&file_contents)?;
+
+        Ok(keystore)
+    }
+
+    pub fn save_to_file<P: AsRef<Path>>(&self, path: P) -> Result<()> {
+        let serialized = serde_json::to_string(self)?;
```
KolbyML, line 98: the struct and impl definition should be right next to each other, not separate by random objects

```rust
enum X {}

impl X;

```

Don't put objects between these, ideally put the other objects below the object and impl definitions

PR #489, opened 2025-05-30
feat: hmac for pbkdf2

Patchoulis:

Creating the HMAC helper function in order to read off the encrypted keystore files normally generated by any of the staking CLI.

Originally I intended to just do it for sha256, but I noticed it apparently would work for any hash function. I instead generalized it to work with most of the hash functions we already use in the odd case we may need it for something else.

### To-Do

This is meant as a part of the PBKDF2 implementation, so this would be next. Additionally, scrypt requires PBKDF2-HMAC-SHA256, so it's a mandatory function for keystore decryption.

### DIFF
--- a/crates/crypto/keystore/src/hmac.rs
+++ b/crates/crypto/keystore/src/hmac.rs
@@ -1,0 +1,98 @@
+use sha2::{Digest, digest::crypto_common::BlockSizeUser};
+use ssz_types::FixedVector;
+
+// Going off of this
+// https://en.wikipedia.org/wiki/HMAC#:~:text=In%20cryptography%2C%20an%20HMAC%20(sometimes,and%20a%20secret%20cryptographic%20key.
+pub fn hmac<T: Digest + BlockSizeUser>(
+    key: &[u8],
+    message: &[u8],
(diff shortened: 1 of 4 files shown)


crates/crypto/keystore/src/hmac.rs
```diff
@@ -0,0 +1,98 @@
...
+    inner_hasher.update(&i_key_pad);
+    inner_hasher.update(message);
+    let inner_hash = inner_hasher.finalize();
+
+    // Compute outer hash
+    let mut outer_hasher = T::new();
+    outer_hasher.update(&o_key_pad);
+    outer_hasher.update(&inner_hash);
+
+    FixedVector::<u8, T::OutputSize>::from(outer_hasher.finalize().to_vec())
```
KolbyML, line 32: which should use from, we should use new, from can lead to unexpected behavior

KolbyML:
Since this function will only be used for sha256 we should only support sha256, please return a B256 and remove the generics as they aren't needed

Patchoulis:
@KolbyML Resolved the issues

PR #492, opened 2025-05-30
feat: add pbkdf2 for keystores

Patchoulis:

A small PR which implements one of the two functions we require in order to obtain a keystore decryption key.

Approach was based on the spec in https://www.ietf.org/rfc/rfc2898.txt
Additionally, I added another text vector from the spec above, which we could use to validate decryption and password validation.

### To-Do

Once this is merged, scrypt can be done, with password validation being trivial after.


crates/crypto/keystore/src/pbkdf2.rs
```diff
@@ -0,0 +1,56 @@
...
+        let block = pbkdf2_helper(password, salt, c, i);
+        if i == l {
+            derived_key.extend_from_slice(&block[..r as usize]);
+        } else {
+            derived_key.extend_from_slice(block.as_ref());
+        }
+    }
+
+    derived_key
+}
```
KolbyML, line 36: I don't think there is a reason to have 1 letter variable names

crates/crypto/keystore/assets/Pbkdf2TestKeystore.json
KolbyML, line 1: can this be mini=fided so it is 1 line

KolbyML (commented):
Left some feedback

Patchoulis:
@KolbyML Fixed it

KolbyML (approved):
:shipit: looks good

PR #496, opened 2025-05-30
feat: handle ReqResp P2P messages in NetworkManager

KolbyML:

We want to handle to requests we are sent over the Req Response P2P domain

Implementing handling the responses some in the network service itself, and others which require access to the database in NetworkManager


Kayden-ML (approved):
Amazing work Looks good 👍

PR #505, opened 2025-05-31
feat: implement /eth/v1/node/syncing in the validator

### LINKED ISSUE #504: Connect GET /eth/v1/node/syncing to the Validator
### Describe the feature

Prior to being able to perform our protocol duties as a validator, we need to check if the beacon node is synced. 

We want to implement the caller to endpoint `eth/v1/node/syncing` in our Validator Client: 
https://ethereum.github.io/beacon-APIs/#/ValidatorRequiredApi/getSyncingStatus

### Additional context

_No response_

shariqnaiyer:

Fixes #504 

I created the expected response struct `SyncStatus` in the beacon_api_crate.
I then created a function called `get_node_syncing_status` in the beacon_api_client to make the request to a beacon node and parse the json response. 

### To-Do

 
### DIFF
--- a/crates/common/beacon_api_types/src/sync.rs
+++ b/crates/common/beacon_api_types/src/sync.rs
@@ -5,7 +5,8 @@ use ssz_derive::{Decode, Encode};
 pub struct SyncStatus {
     #[serde(with = "serde_utils::quoted_u64")]
     pub head_slot: u64,
-    pub sync_distance: usize,
+    #[serde(with = "serde_utils::quoted_u64")]
+    pub sync_distance: u64,
     pub is_syncing: bool,
     pub is_optimistic: bool,
     pub el_offline: bool,
--- a/crates/common/beacon_api_types/src/sync.rs
+++ b/crates/common/beacon_api_types/src/sync.rs
@@ -1,0 +1,12 @@
+use serde::{Deserialize, Serialize};
+use ssz_derive::{Decode, Encode};
+
+#[derive(Debug, Deserialize, Serialize, Encode, Decode)]
+pub struct SyncStatus {
+    #[serde(with = "serde_utils::quoted_u64")]
+    pub head_slot: u64,
+    pub sync_distance: usize,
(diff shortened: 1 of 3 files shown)


crates/common/beacon_api_types/src/sync.rs
```diff
@@ -0,0 +1,12 @@
+use serde::{Deserialize, Serialize};
+use ssz_derive::{Decode, Encode};
+
+#[derive(Debug, Deserialize, Serialize, Encode, Decode)]
+pub struct SyncStatus {
+    #[serde(with = "serde_utils::quoted_u64")]
+    pub head_slot: u64,
+    pub sync_distance: usize,
```
KolbyML, line 8:
```suggestion
#[derive(Debug, Deserialize, Serialize, Encode, Decode)]
pub struct SyncStatus {
    #[serde(with = "serde_utils::quoted_u64")]
    pub head_slot: u64,
    #[serde(with = "serde_utils::quoted_u64")]
    pub sync_distance: u64,
```

crates/common/validator/src/beacon_api_client/mod.rs
```diff
@@ -73,10 +75,34 @@ impl BeaconApiClient {
...
+        if !response.status().is_success() {
+            return Err(ValidatorError::RequestFailed {
+                status_code: response.status(),
+            });
+        }
+
+        response
+            .json()
+            .await
+            .map_err(|err| ValidatorError::JsonDecodeError(err.to_string()))
```
KolbyML, line 99:
```suggestion
        response
            .json()
            .await
```

we should just be able to do this no? as from should be implemented right?

KolbyML (commented):
Left some feedback

crates/common/validator/src/beacon_api_client/mod.rs
KolbyML: fix this in other places as well

KolbyML (approved):
:shipit: looks good

Issue #506, opened 2025-05-31
Add Keystore password validation


Patchoulis:
### Describe the feature

Essentially this issue is to finish up password validation for keystores in accordance to https://eips.ethereum.org/EIPS/eip-2335

Patchoulis:
Could I be assigned this?

PR #507, opened 2025-05-31
feat: add password validation and finish up scrypt for keystores

Patchoulis:

fixes https://github.com/ReamLabs/ream/issues/506

It also adds the missing scrypt compatability

Uses thread to handle the parallelization factor used in scrypt.
Also added parameter verification for pbkdf2 to ensure validity.
Added a test vector from https://eips.ethereum.org/EIPS/eip-2335

Salsa and a lot of the Scrypt specific methods perform in-place outputs due to the high memory usage of scrypt.

### To-Do

We do need to also do AES-CTR for decryption.


crates/crypto/keystore/src/scrypt.rs
```diff
@@ -0,0 +1,274 @@
...
+                &mut thread_chunk,
+                cost_parameter as usize,
+            );
+            (thread_index, thread_chunk)
+        }));
+    }
+    for thread_handle in thread_handles {
+        let (thread_index, thread_result) = thread_handle
+            .join()
+            .map_err(|err| anyhow!(format!("Thread panicked: {err:?}")))?;
```
KolbyML:
```suggestion
            .map_err(|err| anyhow!(format!("Failed to join thread while doin scrypt_romix: {err:?}")))?;
```

crates/crypto/keystore/src/scrypt.rs
```diff
@@ -0,0 +1,274 @@
...
+        rearranged_blocks[block_index * 64..(block_index + 1) * 64]
+            .copy_from_slice(&input_output[block_index * 128..block_index * 128 + 64]);
+        rearranged_blocks[(block_size + block_index) * 64..(block_size + block_index + 1) * 64]
+            .copy_from_slice(&input_output[block_index * 128 + 64..(block_index + 1) * 128]);
+    }
+
+    input_output.copy_from_slice(&rearranged_blocks);
+}
+
+fn scrypt_romix(block_size: usize, input_output: &mut [u8], cost_parameter: usize) {
```
KolbyML, line 31:
```suggestion
fn scrypt_remix(block_size: usize, input_output: &mut [u8], cost_parameter: usize) {
```

did you mean romix? is that supposed to be random_mix or remix?

Patchoulis: In https://datatracker.ietf.org/doc/html/rfc7914#page-7

It's referred to as "The scryptROMix Algorithm"

Issue #510, opened 2025-06-01
Add Keystore Decryption


Patchoulis:
### Describe the feature

Now that we have password validation for keystores, the next logical step would be to add keystore decryption using those passwords.

This would allow us to get public and private key pairs from the staking cli as an example, which our validator could use.

Patchoulis:
Could I be assigned this?

PR #511, opened 2025-06-01
feat: add keystore decryption

Patchoulis:

 fixes https://github.com/ReamLabs/ream/issues/510

Adds a method to get actual key pairs from an encrypted keystore, such as those generated with the staking CLI.

### To-Do

The next logical step is to use these key pairs for our validator client.

### DIFF
--- a/crates/crypto/keystore/src/decrypt.rs
+++ b/crates/crypto/keystore/src/decrypt.rs
@@ -1,0 +1,18 @@
+use aes::{
+    cipher::{BlockEncrypt, KeyInit, generic_array::GenericArray},
+    Aes128,
+};
+
+pub fn aes128_ctr(buf: &mut [u8], key: [u8; 16], iv: &[u8; 16]) {
+    let cipher = Aes128::new(&key.into());
+    let mut ctr = u128::from_be_bytes(*iv);
+
+    for chunk in buf.chunks_mut(16) {
+        let mut block = GenericArray::from(ctr.to_be_bytes());
+        cipher.encrypt_block(&mut block);
+        for (b, k) in chunk.iter_mut().zip(block.iter()) {
+            *b ^= k;
+        }
+        ctr = ctr.wrapping_add(1);
+    }
+}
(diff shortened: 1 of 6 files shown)


crates/crypto/keystore/src/decrypt.rs
```diff
@@ -0,0 +1,18 @@
...
+
+    for chunk in buf.chunks_mut(16) {
+        let mut block = GenericArray::from(ctr.to_be_bytes());
+        cipher.encrypt_block(&mut block);
+        for (b, k) in chunk.iter_mut().zip(block.iter()) {
+            *b ^= k;
+        }
+        ctr = ctr.wrapping_add(1);
+    }
+}
```
KolbyML: is there any way to make this more readable? I don't like `b` `k` or `ctr`, `iv` what do these variables mean

crates/crypto/keystore/src/keystore.rs
```diff
@@ -17,6 +18,11 @@ pub struct EncryptedKeystore {
     pub version: u64,
 }
 
+pub struct Keystore {
+    pub public_key: PubKey,
```
KolbyML: can you rename the PubKey object to PublicKey

crates/crypto/keystore/src/keystore.rs
```diff
@@ -49,6 +55,54 @@ impl EncryptedKeystore {
...
+            "Password provided is invalid!"
+        );
+
+        let mut private_key = PrivateKey {
+            inner: B256::from_slice(self.crypto.cipher.message.as_slice()),
+        };
+        match &self.crypto.cipher.params {
+            CipherParams::Aes128Ctr { iv } => {
+                let key_param: [u8; 16] = derived_key[0..16].try_into().map_err(|err| {
+                    anyhow::anyhow!(format!(
```
KolbyML:
```suggestion
                    anyhow!(format!(
```

do this for all other parts of the code

KolbyML (approved):
:shipit: looks good

Issue #482, opened 2025-05-29
Connect POST /eth/v1/validator/duties/attester/{epoch} to the validator client


Patchoulis:
### Describe the feature

Essentially the goal here is to make a method to get the attestation duties, which would be used to check if the validator is currently supposed to attest.

PR #512, opened 2025-06-01
feat: connect POST duties/attester/{epoch} to the validator

Patchoulis:

fixes https://github.com/ReamLabs/ream/issues/482

Fairly small PR. I made it in the same style as get_proposer_duties. With it, you should be able to pass in a list of validator_indicies to get the duties.

### To-Do

Once the validator is up and running(as well as a beacon or beam node), we need to add integration tests using endpoints like this.


Patchoulis:
If I recall correctly, this is one of the json only APIs? Correct me if I'm wrong, but I didn't see any mention of it accepting or returning ssz bytes according to https://ethereum.github.io/beacon-APIs

PR #513, opened 2025-06-01
fix: failing to create status reply on non-fully synced db

KolbyML:

We were failing to create the reply status message

If we can't build the head due to not having enough history, just return our head as the finalized root, in the case I am testing that is the only root we have anyways


Kayden-ML (approved):
Looks good 👍

Issue #519, opened 2025-06-02
Implement a slot and epoch clock for the validator


Patchoulis:
### Describe the feature

From my knowledge, most validators work by running on a loop, initially getting the genesis timestamp, and then running on a set clock schedule by using the network parameters to time when to perform tasks like fetching the attester duties and proposer duties.

We need a way to do this ourselves in order to get a working validator up.

PR #520, opened 2025-06-02
feat: added a new slot and epoch clock.

Patchoulis:

fixes https://github.com/ReamLabs/ream/issues/519

My thought process behind it was that we would want the actual timer to be on a blocking thread. Otherwise the plan is we would call all of our "on epoch" code and "on slot" code by using the timer.

Didn't want to use more than one dedicated thread for timing-related code, especially since you could determine when a new epoch occurs at the same time as when you'd determine when a new slot occurs.

slot and epoch are also stored in the clock in case we would want a way to modify or read this from another thread.

### To-Do

I added a parameter for genesis_time, so when the PR for /eth/v1/beacon/genesis is merged in, we would make an initial call for the genesis information from a beacon node, and use the timestamp from the response to compute the current slot and epoch.

We would need to pass that in once ready.

### DIFF
--- a/crates/common/validator/src/ticker.rs
+++ b/crates/common/validator/src/ticker.rs
@@ -1,79 +1,55 @@
-use std::{
-    sync::{
-        Arc,
-        atomic::{AtomicU64, Ordering},
-    },
-    thread::sleep,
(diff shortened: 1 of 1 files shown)


KolbyML:
Why not use an Interval?

Issue #529, opened 2025-06-04
Add CLI support for importing keystores


Patchoulis:
### Describe the feature

One of the tasks we need to do is add CLI for initializing a validator by importing keystores. Since we have a way of decrypting keystores using passwords, it makes sense to actually build up the validator service using this.

PR #531, opened 2025-06-04
feat: added validator service and import-keystores for cli

Patchoulis:

fixes https://github.com/ReamLabs/ream/issues/529
fixes https://github.com/ReamLabs/ream/issues/519

 This PR sets up the initial validator service. Due to the nature of the APIs, as well as for convenience purposes, I've made it  so the validator service can handle multiple individual validators. I've also set up the CLI to accept either a password or a password plaintext file for keystore decryption.

As listed above, the validator service can handle a variety of individual validators. Perhaps we should add another parameter in the CLI for the number of validators?

### To-Do

It might be a good idea to use the /eth/v1/beacon/states/{state_id}/validators endpoint instead of /eth/v1/beacon/states/{state_id}/validators/{validator_id} for multiple validator queries. This would lower the number of API calls for those running multiple validators.


bin/ream/src/cli/import_keystores.rs
```diff
@@ -0,0 +1,61 @@
...
+
+use anyhow::anyhow;
+use ream_keystore::keystore::EncryptedKeystore;
+use unicode_normalization::UnicodeNormalization;
+
+pub fn load_password_file(path: &PathBuf) -> anyhow::Result<String> {
+    let contents = read_to_string(path)
+        .map_err(|err| anyhow!(format!("Unable to load password file: {err:?}")))?;
+    Ok(contents.trim_end_matches(&['\n', '\r'][..]).to_string())
+}
```
KolbyML:
```suggestion
pub fn load_password_file(path: &PathBuf) -> anyhow::Result<String> {
    Ok(read_to_string(path)
        .map_err(|err| anyhow!("Unable to load password file: {err:?}"))?.trim_end_matches(&['\n', '\r'][..]).to_string())
}
```

we don't need to use format!() in anyhow!, please remove that in all other places in this PR, and we might as inline all of this

bin/ream/src/cli/import_keystores.rs
```diff
@@ -0,0 +1,61 @@
...
+
+pub fn process_password(password: String) -> String {
+    password
+        .nfkd()
+        .filter(|&character_unprocessed| {
+            let character = character_unprocessed as u32;
+            !((character == 0x7F) || (character <= 0x1F) || (0x80..=0x9F).contains(&character))
+        })
+        .collect::<String>()
```
KolbyML:
```suggestion
pub fn process_password(password: String) -> String {
    password
        .nfkd()
        .filter(|&character_unprocessed| {
            let character = character_unprocessed as u32;
            !((character == 0x7F) || (character <= 0x1F) || (0x80..=0x9F).contains(&character))
        })
        .collect()
```

shouldn't we be fine not defining the type

bin/ream/src/cli/validator_node.rs
```diff
@@ -25,6 +27,32 @@ pub struct ValidatorNodeConfig {
...
+    pub suggested_fee_recipient: Address,
+
+    #[arg(long, help = "The plaintext password file to use for keystores")]
+    pub password_file: Option<PathBuf>,
+
+    #[arg(
+        long,
+        help = "The password to use for keystores. It's recommended to use password-file over this in order to prevent your keystore password from appearing in the shell history"
+    )]
+    pub password: Option<String>,
```
KolbyML: do other clients support both?, if we want to support both we should set up the flags so you can only submit one or the other

crates/common/validator/src/validator.rs
```diff
@@ -15,3 +33,112 @@ pub fn check_if_validator_active(
...
+    }
+
+    pub fn on_slot(slot: u64) {
+        println!("Current Slot: {}", slot);
+    }
+
+    pub fn on_epoch(epoch: u64) {
+        println!("Current Epoch: {}", epoch);
+    }
+}
```
KolbyML:
```suggestion
    pub async fn start(self) {
        let genesis_info = self
            .beacon_api_client
            .get_genesis()
            .await
            .expect("Could not retrieve genesis information");
        let clock_handle = self.start_clock(genesis_info.data.genesis_time);

        let seconds_per_slot = self.network.seconds_per_slot;
        let genesis_instant = UNIX_EPOCH + Duration::from_secs(genesis_time);
        let elapsed = SystemTime::now()
            .duration_since(genesis_instant)
            .expect("System Time is before the genesis time");

        let mut slot = elapsed.as_secs() / seconds_per_slot;
        let mut epoch = compute_epoch_at_slot(slot);

        let mut interval = {
            let interval_start =
                Instant::now() - (elapsed - Duration::from_secs(slot * seconds_per_slot));
            interval_at(interval_start, Duration::from_secs(seconds_per_slot))
        };
        interval.set_missed_tick_behavior(MissedTickBehavior::Burst);

       loop {
            select! {
                _ => interval.tick() {
                        slot += 1;
                        let current_epoch = compute_epoch_at_slot(slot);
        
                        if current_epoch != epoch {
                            epoch = current_epoch;
                            self.on_epoch(epoch);
                        }
                        self.on_slot(slot);
                }
            }
       }
    }

    pub fn on_slot(&self, slot: u64) {
        println!("Current Slot: {}", slot);
    }

    pub fn on_epoch(&self, epoch: u64) {
        println!("Current Epoch: {}", epoch);
    }
}
```

I think this service should look more like this, what do you think?

KolbyML (commented):
Left some feedback

crates/common/validator/src/validator.rs
```diff
@@ -15,3 +33,112 @@ pub fn check_if_validator_active(
...
+    pub suggested_fee_recipient: Arc<Address>,
+    pub network: Arc<NetworkSpec>,
+    pub executor: ReamExecutor,
+}
+
+impl ValidatorService {
+    pub fn new(
+        keystores: Vec<Keystore>,
+        suggested_fee_recipient: Address,
+        network: Arc<NetworkSpec>,
```
KolbyML:
```suggestion
pub struct ValidatorService {
    pub beacon_api_client: Arc<BeaconApiClient>,
    pub validators: Vec<Arc<ValidatorInfo>>,
    pub suggested_fee_recipient: Arc<Address>,
    pub executor: ReamExecutor,
}

impl ValidatorService {
    pub fn new(
        keystores: Vec<Keystore>,
        suggested_fee_recipient: Address,
```

set the global like we do in the beacon client, because functions depend on the global being set

KolbyML (approved):
:shipit: looks good

Issue #532, opened 2025-06-05
Connect POST /eth/v1/beacon/states/{state_id}/validators


Patchoulis:
### Describe the feature

Previously we did /eth/v1/beacon/states/{state_id}/validators/{validator_id}, but if we want to do a multi-validator system for the validator service, we should add a way to query for multiple validators with a single call.

PR #533, opened 2025-06-05
feat: connected POST /eth/v1/beacon/states/{state_id}/validators

Patchoulis:

fixes https://github.com/ReamLabs/ream/issues/532

Similar style to how we connected the GET /eth/v1/beacon/states/{state_id}/validators/{validator_id} endpoint, but instead, it has two list options for filters.

I also converted status for ValidatorStatus into an enum to account to limit the number of options available for status, and simplify future status checks, and noticed on the list of beacon APIs, the post body is apparently supposed to contain "statuses" instead of "status", so I made a quick fix for that.

### To-Do

Use the new method to perform a check for whether or not the validators are ready in the validator service.


KolbyML (approved):
:shipit: looks good

Issue #451, opened 2025-05-25
Connect p2p gossip messages to network manager


Kayden-ML:
### Describe the feature

Connect p2p gossip messages to network manager

I will also plug in the respective on_* functions in the fork choice code, which are accessible from the BeaconChain struct

PR #534, opened 2025-06-05
feat: Connect p2p gossip messages to network manager and add respective on_*

Kayden-ML:

Fixes: https://github.com/ReamLabs/ream/issues/451

Connected p2p gossip messages to network manager and added respective on_


Kayden-ML:
@KolbyML  ready for review

Issue #539, opened 2025-06-07
Fetch Validator Indicies


Patchoulis:
### Describe the feature

When the validator service starts, all we have is the public and private key.

We need a way to fetch the indicies associated with each validator, since they're used later for both proposer and attester duties.

PR #540, opened 2025-06-07
feat: fetch validator indicies

Patchoulis:

fixes https://github.com/ReamLabs/ream/issues/539

I had to modify the validator info to instead be the keystore(once decrypted), and a hashmap associated with the indicies for faster lookups.

### To-Do

Use the indicies to later fetch the duties


KolbyML (approved):
:shipit: looks good

PR #544, opened 2025-06-08
feat: Initial implementation of ream's account manager

ch4r10t33r:

Initial implementation of ream's `account_manager` tool. This tool will replace the existing account_manager tool used to generate validator keys for the beacon chain.

### To-Do

 
### DIFF
--- a/bin/ream/src/cli/account_manager.rs
+++ b/bin/ream/src/cli/account_manager.rs
@@ -1,0 +1,46 @@
+use clap::Parser;
+
+#[derive(Debug, Parser)]
+pub struct AccountManagerConfig {
+    /// Verbosity level
+    #[arg(short, long, default_value_t = 3)]
+    pub verbosity: u8,
+
+    /// Account lifetime in 2 ** lifetime slots
+    #[arg(short, long, default_value_t = 28)]
+    pub lifetime: u64,
+
+    /// Chunk size for messages
+    #[arg(short, long, default_value_t = 5)]
+    pub chunk_size: u64,
+}
+
+impl Default for AccountManagerConfig {
+    fn default() -> Self {
+        Self {
+            verbosity: 3,
+            lifetime: 28,
+            chunk_size: 5,
+        }
+    }
+}
+
+impl AccountManagerConfig {
+    pub fn new() -> Self {
+        Self::parse()
+    }
+
+    pub fn validate(&self) -> anyhow::Result<()> {
+        // Validate chunk size
+        if self.chunk_size < 4 {
+            anyhow::bail!("Chunk size must be at least 4");
+        }
+
+        // Validate lifetime
+        if self.lifetime < 18 {
+            anyhow::bail!("Lifetime must be at least 18");
+        }
+
+        Ok(())
+    }
+}
--- a/bin/ream/src/main.rs
+++ b/bin/ream/src/main.rs
@@ -1,5 +1,4 @@
-use std::env;
-use std::process;
+use std::{env, process};
 
 use clap::Parser;
 use ream::cli::{
--- a/bin/ream/src/main.rs
+++ b/bin/ream/src/main.rs
@@ -1,8 +1,10 @@
 use std::env;
+use std::process;
 
 use clap::Parser;
 use ream::cli::{
     Cli, Commands,
+    account_manager::AccountManagerConfig,
     beacon_node::BeaconNodeConfig,
     import_keystores::{load_keystore_directory, load_password_file, process_password},
     validator_node::ValidatorNodeConfig,
@@ -45,6 +47,7 @@ async fn main() {
             run_beacon_node(*config, async_executor, main_executor).await
         }
         Commands::ValidatorNode(config) => run_validator_node(*config, async_executor).await,
+        Commands::AccountManager(config) => run_account_manager(*config).await,
     }
 }
 
@@ -151,4 +154,23 @@ pub async fn run_validator_node(config: ValidatorNodeConfig, async_executor: Rea
     .expect("Failed to create validator service");
 
     validator_service.start().await;
+}
+
+pub async fn run_account_manager(config: AccountManagerConfig) {
+    info!("starting up account manager...");
+
+    // Validate the configuration
+    config
+        .validate()
+        .expect("Invalid account manager configuration");
+
+    info!(
+        "Account manager configuration: lifetime={}, chunk_size={}",
+        config.lifetime, config.chunk_size
+    );
+
+    ream_account_manager::generate_keys();
+
+    info!("Account manager completed successfully");
+    process::exit(0);
 }
--- a/crates/account_manager/Readme.md
+++ b/crates/account_manager/Readme.md
@@ -1,0 +1,25 @@
+# Ream Account Manager
+
+**Ream Chain Account Manager** is a CLI tool written in **Rust** for generating **Winternitz One-Time Signature (WOTS)** validator keys for use on the **Beam Chain**.
+
+> ⚡ Secure, deterministic, and purpose-built for lightweight cryptographic identity management.
+
+---
+
+## 🚀 Features
+
+- 🔐 Generate Winternitz OTS (WOTS) key pairs
+- 🧾 Export public/private keys to files or standard output
+- 🏷️ Tag key generation with metadata or labels
+- 🛡️ Designed with validator node integration in mind
+
+---
+
+## 📦 Installation
+
+### Prerequisites
+- Rust & Cargo: [Install Rust](https://www.rust-lang.org/tools/install)
+
+### Clone and Build
+
+TODO
--- a/crates/account_manager/src/lib.rs
+++ b/crates/account_manager/src/lib.rs
@@ -1,5 +1,23 @@
+use std::time::Instant;
+
+use hashsig::signature::{
+    SignatureScheme,
+    generalized_xmss::instantiations_poseidon::lifetime_2_to_the_20::winternitz::SIGWinternitzLifetime20W4,
+};
+use hashsig::signature::SignatureScheme;
+use rand::{Rng, thread_rng};
 use tracing::info;
 
 pub fn generate_keys() {
     info!("Generating beam chain validator keys.....");
+    let mut rng = thread_rng();
+    // measure_time::<SIGWinternitzLifetime20W4, _>("Poseidon - L 20 - Winternitz - w 4", &mut rng);
+    info!("Key generation complete");
+}
+
+fn measure_time<T: SignatureScheme, R: Rng>(description: &str, rng: &mut R) {
+    let start = Instant::now();
+    // let (_pk, _sk) = T::gen(rng);
+    let duration = start.elapsed();
+    info!("{} - Gen: {:?}", description, duration);
 }
--- a/crates/account_manager/src/lib.rs
+++ b/crates/account_manager/src/lib.rs
@@ -1,0 +1,5 @@
+use tracing::info;
+
+pub fn generate_keys() {
+    info!("Generating beam chain validator keys.....");
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1138,6 +1138,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "blake2b_simd"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "06e903a20b159e944f91ec8499fe1e55651480c541ea0a584f5d967c49ad9d99"
+dependencies = [
+ "arrayref",
+ "arrayvec",
+ "constant_time_eq",
+]
+
+[[package]]
 name = "block-buffer"
 version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1157,13 +1168,26 @@ dependencies = [
 
 [[package]]
 name = "bls12_381"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a3c196a77437e7cc2fb515ce413a6401291578b5afc8ecb29a3c7ab957f05941"
+dependencies = [
+ "ff 0.12.1",
+ "group 0.12.1",
+ "pairing 0.22.0",
+ "rand_core 0.6.4",
+ "subtle",
+]
+
+[[package]]
+name = "bls12_381"
 version = "0.8.0"
 source = "git+https://github.com/zkcrypto/bls12_381?rev=9ea427c0eb1a7e2ac16902a322aea156c496ddb0#9ea427c0eb1a7e2ac16902a322aea156c496ddb0"
 dependencies = [
  "digest 0.10.7",
- "ff",
- "group",
- "pairing",
+ "ff 0.13.1",
+ "group 0.13.0",
+ "pairing 0.23.0",
  "rand_core 0.6.4",
  "subtle",
 ]
@@ -1459,6 +1483,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "constant_time_eq"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7c74b8349d32d297c9134b8c88677813a227df8f779daa29bfc29c183fe3dca6"
+
+[[package]]
 name = "cookie"
 version = "0.16.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1539,6 +1569,16 @@ version = "0.5.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "82b8f8f868b36967f9606790d1903570de9ceaf870a7bf9fbbd3016d636a2cb2"
 dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
+name = "crossbeam-deque"
+version = "0.8.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9dd111b7b7f7d55b72c0a6ae361660ee5853c9af73f70c3c2ef6858b950e2e51"
+dependencies = [
+ "crossbeam-epoch",
  "crossbeam-utils",
 ]
 
@@ -2003,9 +2043,9 @@ dependencies = [
  "base16ct",
  "crypto-bigint",
  "digest 0.10.7",
- "ff",
+ "ff 0.13.1",
  "generic-array",
- "group",
+ "group 0.13.0",
  "pem-rfc7468",
  "pkcs8",
  "rand_core 0.6.4",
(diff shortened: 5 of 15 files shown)


crates/account_manager/Readme.md
```diff
@@ -0,0 +1,25 @@
+# Ream Account Manager
+
+**Ream Chain Account Manager** is a CLI tool written in **Rust** for generating **Winternitz One-Time Signature (WOTS)** validator keys for use on the **Beam Chain**.
```
syjn99:
```suggestion
**Ream Account Manager** is a CLI tool written in **Rust** for generating **Winternitz One-Time Signature (WOTS)** validator keys for use on the **Beam Chain**.
```

bin/ream/src/cli/account_manager.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub fn validate(&self) -> anyhow::Result<()> {
+        // Validate chunk size
+        if self.chunk_size < 4 {
+            anyhow::bail!("Chunk size must be at least 4");
+        }
+
+        // Validate lifetime
+        if self.lifetime < 18 {
+            anyhow::bail!("Lifetime must be at least 18");
+        }
```
syjn99: Thinking of user experience, I'd like to suggest set the minimum value in these cases. Open to discuss those issues

crates/account_manager/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+use rand::SeedableRng;
+use rand_chacha::ChaCha20Rng;
+use tracing::info;
+
+pub fn generate_keys(seed_phrase: &str) {
+    info!("Generating beam chain validator keys.....");
+
+    // Convert seed phrase to bytes for RNG initialization
+    let seed_bytes = seed_phrase.as_bytes();
+    let mut rng = ChaCha20Rng::from_seed(seed_bytes.try_into().unwrap_or([0; 32]));
```
syjn99: I don't think this is right, as `try_into()` will always fail as `seed_bytes` are not actually 32 bytes. So in this implementation the seed will always be zero bytes.

I'm not so sure about the specification but we can hash the `seed_phrase` to fit into 32 bytes.

syjn99:
```rust
    // Convert seed phrase to bytes for RNG initialization
    let seed_bytes = seed_phrase.as_bytes();
    let actual_seed = seed_bytes.try_into().unwrap_or([0; 32]);
    println!("Seed bytes: {:?}", actual_seed);
    let mut rng = ChaCha20Rng::from_seed(actual_seed);
```

This code prints like:

```
2025-06-09T15:05:04.588150Z  INFO ream: starting up account manager...
2025-06-09T15:05:04.588272Z  INFO ream: Account manager configuration: lifetime=28, chunk_size=5
2025-06-09T15:05:04.589092Z  WARN ream::cli::account_manager: ⚠️  IMPORTANT: Generated new seed phrase: erosion duty ball onion hole squirrel liberty chair rare insane alpha grunt coyote cloud food abuse reason unable book misery auto kid impact equal
2025-06-09T15:05:04.589114Z  WARN ream::cli::account_manager: ⚠️  Please save this seed phrase somewhere safe. You will need it to recover your keys.
2025-06-09T15:05:04.589214Z  INFO ream_account_manager: Generating beam chain validator keys.....
Seed bytes: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

ch4r10t33r: Well spotted @syjn99 . I'll investigate and change this. I hadn't fully tested the functionality yet.

bin/ream/src/cli/account_manager.rs
ch4r10t33r: Will do. I was planning to do this in subsequent iterations. But I guess, adding a minimum value wouldn't take too long.

crates/account_manager/src/lib.rs
ch4r10t33r: > I don't think this is right, as `try_into()` will always fail as `seed_bytes` are not actually 32 bytes. So in this implementation the seed will always be zero bytes.
> 
> I'm not so sure about the specification but we can hash the `seed_phrase` to fit into 32 bytes.

Alternatively, I can use the first public/private key derived from the seed phrase as the initial salt for generating the validator keys. Once the key’s lifetime expires, the next public/private key from the same seed phrase would serve as the new salt. wdyt?

syjn99: The one drawback of your suggestion is that users cannot derive a proper keypair from a given seed phrase at certain time, as there's no way to know how many derivation is needed.
AFAIK there's no common specification for Winternitz-OTS right now (which corresponds to [ERC-2333 ~ ERC-2335](https://ethereum-magicians.org/t/erc-2333-erc-2334-erc-2335-bls12-381-key-generation-deterministic-account-hierarchy-keystore/19566)), it seems quite a huge agenda to decide those things. We need some kind of "index" thing to derive, and one of my naive idea is using a timestamp or epoch as an index.

ch4r10t33r: > it seems quite a huge agenda to decide those things. We need some kind of "index" thing to derive.

Yes, in the approach I’m proposing, we’d need to store an index in the keystore to ensure that a new set of keys is generated each time.

> and one of my naive idea is using a timestamp or epoch as an index.

That could work, but even with this approach, we’d still need to store the last used timestamp or epoch in the keystore to avoid regenerating the same set of keys.

I agree there’s no perfect solution at the moment, and much of this may evolve once the Beam spec is released—assuming it addresses this topic at all. I suggest we proceed with a functional implementation for now and revisit it once the spec is available.

ch4r10t33r: > I don't think this is right, as `try_into()` will always fail as `seed_bytes` are not actually 32 bytes. So in this implementation the seed will always be zero bytes.

This issue has now been addressed. The `seed_bytes` are no longer zero. I'll address the other comment in a new PR as it is bigger change. wdyt? @syjn99

syjn99: > I suggest we proceed with a functional implementation for now and revisit it once the spec is available.

Yes, I also agree with this point. Seems the issue is resolved!

bin/ream/src/cli/account_manager.rs
```diff
@@ -0,0 +1,70 @@
...
+            warn!("Lifetime must be at least 18, resetting to 18");
+            self.lifetime = 18;
+        }
+
+        Ok(())
+    }
+
+    pub fn get_seed_phrase(&self) -> String {
+        if let Some(phrase) = &self.seed_phrase {
+            phrase.clone()
```
syjn99, line 59: I think we need to sanitize this argument to follow BIP-32 mnemonic form.

bin/ream/src/cli/account_manager.rs
```diff
@@ -0,0 +1,70 @@
...
+        } else {
+            let mnemonic = Mnemonic::random(OsRng, Default::default());
+            let phrase = mnemonic.phrase().to_string();
+            warn!("⚠️  IMPORTANT: Generated new seed phrase: {}", phrase);
+            warn!(
+                "⚠️  Please save this seed phrase somewhere safe. You will need it to recover your keys."
+            );
+            phrase
```
syjn99, line 67: At this point, the instance of `AccountManagerConfig` doesn't hold the newly generated phrase. I wonder whether this is intended or not.

crates/account_manager/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+    // Hash the seed phrase to get a 32-byte seed
+    let mut hasher = Sha256::new();
+    hasher.update(seed_phrase.as_bytes());
+    let seed = hasher.finalize().into();
+    info!("Seed: {:?}", seed);
+
+    let mut rng = <ChaCha20Rng as SeedableRng>::from_seed(seed);
+
+    // measure_time::<SIGWinternitzLifetime20W4, _>("Poseidon - L 20 - Winternitz - w 4", &mut rng);
+    let (_pk, _sk) = SIGWinternitzLifetime20W4::r#gen(&mut rng);
```
syjn99: Later, we might need an abstraction layer for this, by adding a new crate in our `crypto` folder. (Just like what BLS related codes are placed in)

crates/account_manager/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+use tracing::info;
+
+pub fn generate_keys(seed_phrase: &str) {
+    info!("Generating beam chain validator keys.....");
+
+    // Hash the seed phrase to get a 32-byte seed
+    let mut hasher = Sha256::new();
+    hasher.update(seed_phrase.as_bytes());
+    let seed = hasher.finalize().into();
+    info!("Seed: {:?}", seed);
```
KolbyML:
```suggestion
    info!("Seed: {seed:?}");
```

It is best if we inline this

crates/account_manager/src/lib.rs
KolbyML: can we not use shorthands like pk and sk, it isn't that readable

bin/ream/src/cli/account_manager.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub fn get_seed_phrase(&self) -> String {
+        if let Some(phrase) = &self.seed_phrase {
+            phrase.clone()
+        } else {
+            let mnemonic = Mnemonic::random(OsRng, Default::default());
+            let phrase = mnemonic.phrase().to_string();
+            warn!("⚠️  IMPORTANT: Generated new seed phrase: {}", phrase);
+            warn!(
+                "⚠️  Please save this seed phrase somewhere safe. You will need it to recover your keys."
+            );
```
KolbyML:
```suggestion
            warn!("⚠️ IMPORTANT: Generated new seed phrase: {phrase}");
            warn!(
                "⚠️ Please save this seed phrase somewhere safe. You will need it to recover your keys."
            );
```

bin/ream/src/cli/account_manager.rs
```diff
@@ -0,0 +1,70 @@
...
+        // Validate chunk size
+        if self.chunk_size < 4 {
+            warn!("Chunk size must be at least 4, resetting to 4");
+            self.chunk_size = 4;
+        }
+
+        // Validate lifetime
+        if self.lifetime < 18 {
+            warn!("Lifetime must be at least 18, resetting to 18");
+            self.lifetime = 18;
```
KolbyML: Can we define constants for these

KolbyML (commented):
@Patchoulis and @shariqnaiyer could you review this PR, because you guys are working on the validator for the Beacon Chain. I think coordinating with how we are going to handle Account Management for the Beacon Chain and also transition to the Beam Chain credentials, would be helpful here. So you guys should coordinate with @ch4r10t33r going forward.

bin/ream/src/cli/account_manager.rs
ch4r10t33r: Yes, I will need to add further validations here. Can I do this in the next PR? I need to refactor a lot of code to consume the new `hash-sig` crate.

bin/ream/src/cli/account_manager.rs
ch4r10t33r: Yes, This was intended for now. I'll make these changes in the next PR as described above.

crates/account_manager/src/lib.rs
ch4r10t33r: Noted. Yes, will do.

ch4r10t33r: Noted. This was just for testing. I'll change the variables names to be more meaningful.

bin/ream/src/main.rs
```diff
@@ -152,3 +154,23 @@ pub async fn run_validator_node(config: ValidatorNodeConfig, async_executor: Rea
...
     validator_service.start().await;
 }
+
+pub async fn run_account_manager(mut config: AccountManagerConfig) {
+    info!("starting up account manager...");
+
+    // Validate the configuration
+    config
+        .validate()
+        .expect("Invalid account manager configuration");
```
jihoonsong, line 179: `validate` doesn't raise an error.

ch4r10t33r: `validate` doesn't raise an error when you skip the parameters (as we already have defaults values for all the parameters). However, if you input invalid values, it reports an error.

jihoonsong: Isn't it just print warning log?

ch4r10t33r: I am sorry. I am not sure I understand you. Do you want it to print the log? Or you are saying the code under `expect` is redundant?

bin/ream/src/cli/account_manager.rs
```diff
@@ -0,0 +1,79 @@
...
+        Self::parse()
+    }
+
+    pub fn validate(&mut self) -> anyhow::Result<()> {
+        // Validate chunk size
+        if self.chunk_size < MIN_CHUNK_SIZE {
+            warn!(
+                "Chunk size must be at least {}, resetting to {}",
+                MIN_CHUNK_SIZE, MIN_CHUNK_SIZE
+            );
```
KolbyML:
```suggestion
            warn!("Chunk size must be at least {MIN_CHUNK_SIZE}, resetting to {MIN_CHUNK_SIZE}");
```

we should inline this

bin/ream/src/cli/account_manager.rs
```diff
@@ -0,0 +1,79 @@
...
+            self.chunk_size = MIN_CHUNK_SIZE;
+        }
+
+        // Validate lifetime
+        if self.lifetime < MIN_LIFETIME {
+            warn!(
+                "Lifetime must be at least {}, resetting to {}",
+                MIN_LIFETIME, MIN_LIFETIME
+            );
```
KolbyML:
```suggestion
            warn!("Lifetime must be at least {MIN_LIFETIME}, resetting to {MIN_LIFETIME}");
```

bin/ream/src/main.rs
KolbyML: @ch4r10t33r Jihoon is commenting on `pub fn validate(&mut self) -> anyhow::Result<()> {` not using the Result, if the function doesn't use the result you could remove the return or

```rust
    pub fn validate(&mut self) -> anyhow::Result<()> {
        ensure!(self.chunk_size >= MIN_CHUNK_SIZE, "Chunk size must be at least {MIN_CHUNK_SIZE}");
        ensure!(self.lifetime >= MIN_LIFETIME, "Lifetime must be at least {MIN_LIFETIME}");
        Ok(())
    }
```

Maybe something like this ^, I think this would be better just because then the user is very aware that they miss-configured it, logs can be missed etc. Also then the function is using the result

bin/ream/src/main.rs
```diff
@@ -166,3 +168,23 @@ pub async fn run_validator_node(config: ValidatorNodeConfig, async_executor: Rea
...
+    info!(
+        "Account manager configuration: lifetime={}, chunk_size={}",
+        config.lifetime, config.chunk_size
+    );
+
+    let seed_phrase = config.get_seed_phrase();
+    ream_account_manager::generate_keys(&seed_phrase);
+
+    info!("Account manager completed successfully");
+    process::exit(0);
```
```suggestion
```

process exit shouldn't be needed right?

KolbyML (commented):
Left some finally feedback, once it is resolved I think this PR is good to go.

bin/ream/src/main.rs
jihoonsong: Sorry let me be more precise. The way that the validate method of account manager config is implemented is it just sets config's value to the minimum value and prints warning log when any of its values is below it, instead of raising an exception. So current implementation doesn't look like validation from my perspective nor the except is necessary. (I'm not pointing out the naming though.) Not sure if you're planning to add some code that raises an exception in the following PRs.

bin/ream/src/main.rs
jihoonsong: Without it, it dies with error saying "runtime is dropped within..." It's same with after validator node finishes. It's coming from L42 and L44. One of solutions is declaring executor only when it's used. (e.g., bring those lines into L47)

bin/ream/src/main.rs
ch4r10t33r: Yes, future PRs will introduce additional parameters that will need proper validation and error handling. So if that’s alright, I’ll keep it as is for now and update it later?

bin/ream/src/main.rs
ch4r10t33r: This is necessary—if I skip it, the program panics because the async function doesn't complete properly. In fact, the same issue also occurs with the ream node command.

KolbyML: could you move this outside the function and put it at the end of main() then

ch4r10t33r: https://github.com/ReamLabs/ream/pull/544/#discussion_r2148809154

As explained above, I'll be adding additional parameters to this command, so if its ok, I'll retain the function as-is for now. @KolbyML

bin/ream/src/main.rs
jihoonsong: I'm fine with leaving the except but I'd prefer Kolby's suggestion. I think being explicit to user's misconfiguration gives better UX.

ch4r10t33r:
> Left some finally feedback, once it is resolved I think this PR is good to go.

All done now. Thank you.

bin/ream/src/main.rs
ch4r10t33r: Done. Changed as suggested. @jihoonsong @KolbyML

bin/ream/src/main.rs
jihoonsong: Given that it doesn't exit properly with the validator command as well, I think this should be addressed. But it looks like out of this PR's scope.

KolbyML: I am sure why this was set as resolved. 



> Without it, it dies with error saying "runtime is dropped within..." It's same with after validator node finishes. It's coming from L42 and L44. One of solutions is declaring executor only when it's used. (e.g., bring those lines into L47)

The problem Jihoon mentioned is present for the Beacon Node and Validator is it not?, in that case process::exit(0); should be below

```rust
    match cli.command {
        Commands::BeaconNode(config) => {
            run_beacon_node(*config, async_executor, main_executor).await
        }
        Commands::ValidatorNode(config) => run_validator_node(*config, async_executor).await,
        Commands::AccountManager(config) => run_account_manager(*config).await,
    }
    process::exit(0);
```

instead of in the `run_account_manager` function

jihoonsong (approved):
Great work!

bin/ream/src/main.rs
jihoonsong: Putting process::exit(0); at the end of main solves it but can we have more elegant solution?

ch4r10t33r: I had set it as resolved as it was outside the scope of this PR. However, I can amend the code to what you've suggested!

ch4r10t33r: I've amended the code as suggested. However, we might need to look into this under a separate issue?  wdyt @jihoonsong @KolbyML

jihoonsong: I'm of the opinion that we can find a better solution but it would be different PR.

Putting process::exit(0); at the end of main doesn't look sexy. Having it until we have a better one is okay though.

PR #546, opened 2025-06-09
fix: ReqResp handler crashing when handling outbound connection

KolbyML:

When I tried to send outbound ReqResp messages, the handler would panic with 

```OutboundStreamState should always be present, poll() should not be in parallel```

We were returning before removing the dead outbound streams, update the code to remove the dead stream before returning


Kayden-ML (approved):
Looks good nice work 👍

PR #547, opened 2025-06-09
feat: send status requests when dialing peers to properly connect with peers

KolbyML:

When dialing a peer, the spec says we must follow up with sending a status message

this serves 2 purposes
- find out if the peer is on the same network or not
- see what block the peer is on so we know what range we can download blocks from them for syncing

Send a status message when dialing a peer, once we get a response we consider a peer connected


crates/networking/manager/src/service.rs
```diff
@@ -286,51 +290,26 @@ impl ManagerService {
...
                                                 connection_id,
                                                 stream_id,
-                                                &format!("Failed to get block for head root {head_root}: {err:?}"),
+                                                &format!("Failed to build status request: {err}"),
                                             );
                                             continue;
                                         }
                                     };
 
+
```
Kayden-ML: can you remove this extra line

Kayden-ML (approved):
Looks good great work 👍

KolbyML:
I am working on an updated test

Issue #543, opened 2025-06-08
Connect POST /eth/v2/beacon/blocks to the Validator Client


shariqnaiyer:
### Describe the feature

Our validator must publish the produced and signed block if it is chosen as a Block Proposer.

This will be accomplished by implementing a POST request to  /eth/v2/beacon/blocks as per the Beacon API Spec:
https://ethereum.github.io/beacon-APIs/#/ValidatorRequiredApi/publishBlockV2

Issue #483, opened 2025-05-29
Connect POST /eth/v2/beacon/blinded_blocks to the Validator Client


shariqnaiyer:
### Describe the feature

Our validator must publish the produced and signed block if it is chosen as a Block Proposer. 

The /eth/v2/beacon/blinded_blocks endpoint enables our Validator to publish a block.

shariqnaiyer:
@KolbyML Could I work on this issue?

PR #551, opened 2025-06-10
feat: add publish_block and publish_blinded_block

shariqnaiyer:

Fixes: #483 
Fixes: #543  

I created POST requests in the beacon_api_client to publish a block and to publish a blinded beacon block as per the Beacon API specs:
https://ethereum.github.io/beacon-APIs/#/ValidatorRequiredApi/publishBlockV2
https://ethereum.github.io/beacon-APIs/#/ValidatorRequiredApi/publishBlindedBlockV2

### To-Do

 
### DIFF
--- a/crates/common/validator/src/beacon_api_client/http_client.rs
+++ b/crates/common/validator/src/beacon_api_client/http_client.rs
@@ -17,12 +17,22 @@ pub enum ContentType {
     Ssz,
 }
 
+impl ContentType {
+    pub fn to_header_value(&self) -> HeaderValue {
+        match self {
+            ContentType::Json => HeaderValue::from_static(JSON_CONTENT_TYPE),
+            ContentType::Ssz => HeaderValue::from_static(SSZ_CONTENT_TYPE),
+        }
+    }
+}
+
 #[derive(Debug, Clone)]
 pub struct ClientWithBaseUrl {
     client: Client,
     base_url: Url,
     content_type: ContentType,
 }
+
 impl ClientWithBaseUrl {
     pub fn new(
         url: Url,
@@ -73,17 +83,10 @@ impl ClientWithBaseUrl {
     ) -> anyhow::Result<RequestBuilder> {
         let url = self.base_url.join(url.as_str())?;
 
-        let mut headers = HeaderMap::new();
-        match content_type {
-            ContentType::Json => {
-                headers.insert(CONTENT_TYPE, HeaderValue::from_static(JSON_CONTENT_TYPE));
-            }
-            ContentType::Ssz => {
-                headers.insert(CONTENT_TYPE, HeaderValue::from_static(SSZ_CONTENT_TYPE));
-            }
-        }
-
-        Ok(self.client.post(url).headers(headers))
+        Ok(self
+            .client
+            .post(url)
+            .header(CONTENT_TYPE, content_type.to_header_value()))
     }
 
     pub async fn execute(&self, request: Request) -> Result<Response, reqwest::Error> {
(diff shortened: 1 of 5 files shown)


crates/common/validator/src/beacon_api_client/mod.rs
```diff
@@ -479,6 +485,64 @@ impl BeaconApiClient {
...
+            .execute(
+                self.http_client
+                    .post("/eth/v2/beacon/blinded_blocks".to_string())?
+                    .query(&[(
+                        "broadcast_validation",
+                        serde_json::to_string(&broadcast_validation)?,
+                    )])
+                    .header(ETH_CONSENSUS_VERSION_HEADER, VERSION)
+                    .body(signed_blinded_beacon_block.as_ssz_bytes())
+                    .build()?,
```
KolbyML: The `Content-Type` header must be specified right? or is that done automatically?

shariqnaiyer:
```rust
impl BeaconApiClient {
    pub fn new(beacon_api_endpoint: Url, request_timeout: Duration) -> anyhow::Result<Self> {
        Ok(Self {
            http_client: ClientWithBaseUrl::new(
                beacon_api_endpoint,
                request_timeout,
                ContentType::Ssz,
            )?,
        })
    }
```
    
The http client is initialized with the content header for ssz

KolbyML: This discussion shouldn't be marked as resolved, `The http client is initialized with the content header for ssz` this is set for the get case, I don't think the post case is handled properly here

crates/common/validator/src/beacon_api_client/http_client.rs
```diff
@@ -62,12 +51,39 @@ impl ClientWithBaseUrl {
...
+        match content_type {
+            ContentType::Json => {
+                headers.insert(CONTENT_TYPE, HeaderValue::from_static(JSON_CONTENT_TYPE));
+            }
+            ContentType::Ssz => {
+                headers.insert(CONTENT_TYPE, HeaderValue::from_static(SSZ_CONTENT_TYPE));
+            }
+        }
+
+        Ok(self.client.post(url).headers(headers))
```
KolbyML: are you able to use `header()` instead of `headers()` here

```suggestion
        Ok(self.client.post(url).header(CONTENT_TYPE, HeaderValue::from_static(match content_type {
            ContentType::Json => JSON_CONTENT_TYPE,
            ContentType::Ssz => SSZ_CONTENT_TYPE,
            })
        }
))
```

something like this or implement a function and do

```suggestion
        Ok(self.client.post(url).header(CONTENT_TYPE, content_type.to_header_value()))
```
^ I think this is cleaner

crates/common/validator/src/beacon_api_client/http_client.rs
```diff
@@ -62,12 +51,39 @@ impl ClientWithBaseUrl {
...
+            ContentType::Json => {
+                headers.insert(CONTENT_TYPE, HeaderValue::from_static(JSON_CONTENT_TYPE));
+            }
+            ContentType::Ssz => {
+                headers.insert(CONTENT_TYPE, HeaderValue::from_static(SSZ_CONTENT_TYPE));
+                headers.insert(ACCEPT, HeaderValue::from_static(ACCEPT_PRIORITY));
+            }
+        }
+
+        Ok(self.client.get(url).headers(headers))
```
KolbyML, line 76: this is good

crates/common/validator/src/beacon_api_client/http_client.rs
```diff
@@ -21,34 +21,23 @@ pub enum ContentType {
 pub struct ClientWithBaseUrl {
     client: Client,
     base_url: Url,
+    content_type: ContentType,
 }
-
 impl ClientWithBaseUrl {
```
KolbyML: image

new line is missing now?

KolbyML (approved):
:shipit: looks good

Issue #554, opened 2025-06-10
Connect POST /eth/v1/beacon/pool/sync_committees for sync committee processing to the validator


Patchoulis:
### Describe the feature

This is another one of the end points we would need in order to perform sync committees.

PR #555, opened 2025-06-10
feat: added /eth/v1/beacon/pool/sync_committees

Patchoulis:

fixes https://github.com/ReamLabs/ream/issues/554

Essentially follows the same format as the subscription endpoint.

### To-Do

With this and the other PR, the goal is to fetch the fork and root, use it to produce a committee message, and then publish that committee message. Aggregators also have to be dealt with later.

### DIFF
--- a/crates/common/beacon_api_types/src/request.rs
+++ b/crates/common/beacon_api_types/src/request.rs
@@ -11,9 +11,6 @@ pub struct ValidatorsPostRequest {
 }
 
 #[derive(Debug, Deserialize, Serialize)]
-pub struct SyncCommitteePostRequest(Vec<SyncCommitteeRequestItem>);
-
-#[derive(Debug, Deserialize, Serialize)]
 pub struct SyncCommitteeRequestItem {
     #[serde(with = "serde_utils::quoted_u64")]
     pub slot: u64,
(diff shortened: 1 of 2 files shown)


crates/common/beacon_api_types/src/request.rs
```diff
@@ -7,3 +9,16 @@ pub struct ValidatorsPostRequest {
     pub ids: Option<Vec<ValidatorID>>,
     pub statuses: Option<Vec<ValidatorStatus>>,
 }
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct SyncCommitteePostRequest(Vec<SyncCommitteeRequestItem>);
```
KolbyML:
```suggestion
```

we don't need this wrapper just inline it

crates/common/validator/src/beacon_api_client/mod.rs
```diff
@@ -158,6 +158,29 @@ impl BeaconApiClient {
         Ok(response.json().await?)
     }
 
+    pub async fn publish_sync_committee_signature(
+        &self,
+        sync_committee_request: SyncCommitteePostRequest,
```
KolbyML:
```suggestion
        sync_committee_request: Vec<SyncCommitteeRequestItem>,
```

KolbyML (approved):
:shipit: looks good

Kayden-ML (approved):
Run cargo and then it looks good nice work 👍

Issue #558, opened 2025-06-12
Implement block proposal logic in the validator


shariqnaiyer:
### Describe the feature

The validator must propose blocks in the flow as defined by the Beacon Spec:
https://ethereum.github.io/beacon-APIs/validator-flow.md

PR #559, opened 2025-06-12
feat: add functionality to propose block in the validator

shariqnaiyer:

Fixes: #558
https://ethereum.github.io/beacon-APIs/validator-flow.md

We implemented a function to generate a radao reveal.
We then generated 2 functions to sign blocks and blinded blocks.
We used these functions to propose the block logic as per the spec.

### To-Do

 

KolbyML (approved):
:shipit: PR generally looks good, but I want a review from @Patchoulis before merging this, because he works on this code with you

Patchoulis:
It seems good from a quick look at it. Normally I'd say we should avoid the .cloned() for the keystores if we're only going to use it once(my impression), but if it's an arc, it should be alright.

Issue #563, opened 2025-06-13
Implement sync committee submissions for non-aggregators


Patchoulis:
### Describe the feature

We currently have methods for both the submission of sync committee messages, as well as duty retrieval for sync duties.

The next step is to add a method which combines all of this in order to actually process the sync committees for non-aggregators.

PR #564, opened 2025-06-13
feat: added the method for sync committees for non-aggregators

Patchoulis:

fixes https://github.com/ReamLabs/ream/issues/563

I modified some of the current code to instead use just the block root instead of the entire beacon block, which would lower the response sizes.

Additionally, the sync committees are submitted with a single request for all the validators to lower the number of requests we need.

### To-Do

The next step is to deal with aggregators, and then hook it up to the duties.

I'll also see if there's any consensus tests specifically for validators so we can validate our validation.

### DIFF
--- a/crates/common/validator/src/contribution_and_proof.rs
+++ b/crates/common/validator/src/contribution_and_proof.rs
@@ -2,8 +2,9 @@ use alloy_primitives::B256;
 use ream_bls::{BLSSignature, PrivateKey, traits::Signable};
 use ream_consensus::{
     electra::beacon_state::BeaconState,
-    misc::{compute_epoch_at_slot, compute_signing_root},
+    misc::{compute_domain, compute_epoch_at_slot, compute_signing_root},
 };
+use ream_network_spec::networks::network_spec;
 use serde::{Deserialize, Serialize};
 use ssz_derive::{Decode, Encode};
 use ssz_types::{BitVector, typenum::U128};
@@ -38,18 +39,12 @@ pub struct SignedContributionAndProof {
     pub signature: BLSSignature,
 }
 
-pub fn get_contribution_and_proof(
-    state: &BeaconState,
-    aggregator_index: u64,
-    contribution: SyncCommitteeContribution,
-    private_key: PrivateKey,
-) -> anyhow::Result<ContributionAndProof> {
+pub fn get_contribution_and_proof(contribution: SyncCommitteeContribution, aggregator_index: u64, privkey: &PrivateKey) -> anyhow::Result<ContributionAndProof> {
     Ok(ContributionAndProof {
         selection_proof: get_sync_committee_selection_proof(
-            state,
             contribution.slot,
             contribution.subcommittee_index,
-            private_key,
+            privkey,
         )?,
         aggregator_index,
         contribution,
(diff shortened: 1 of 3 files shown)


crates/common/validator/src/contribution_and_proof.rs
```diff
@@ -39,33 +37,29 @@ pub struct SignedContributionAndProof {
 }
 
 pub fn get_contribution_and_proof(
-    state: &BeaconState,
-    aggregator_index: u64,
     contribution: SyncCommitteeContribution,
-    private_key: PrivateKey,
+    aggregator_index: u64,
+    privkey: &PrivateKey,
```
KolbyML:
```suggestion
    private_key: &PrivateKey,
```

I am not sure why we are reverting this to make it less readable

crates/common/validator/src/sync_committee.rs
```diff
@@ -136,14 +134,14 @@ pub fn process_sync_committee_contributions(
 }
 
 pub fn get_sync_committee_selection_proof(
-    state: &BeaconState,
     slot: u64,
     subcommittee_index: u64,
-    private_key: PrivateKey,
+    privkey: &PrivateKey,
```
KolbyML: here as well

crates/common/validator/src/validator.rs
```diff
@@ -261,6 +266,49 @@ impl ValidatorService {
...
+                {
+                    if let Ok(signature) = keystore.get().private_key.sign(signing_root.as_ref()) {
+                        return Some(SyncCommitteeRequestItem {
+                            slot,
+                            beacon_block_root,
+                            validator_index,
+                            signature,
+                        });
+                    }
+                }
```
KolbyML: why are we suppressing returning an error if signing fails?

Patchoulis: Oh, this was because if it fails, I didn't want it to prevent the other validators from submitting.
I added a warning message in this scenario though for clarity, and I think another PR we should do in the near future is to validate the actual private keys to prevent us from ever importing invalid key pairs.

crates/common/validator/src/validator.rs
```diff
@@ -261,6 +266,51 @@ impl ValidatorService {
...
+                            validator_index,
+                            signature,
+                        });
+                    } else {
+                        info!("invalid private key detected for validator at index {validator_index:?}")
+                    }
+                }
+                None
+            })
+            .collect::<Vec<_>>();
```
KolbyML: Signing a root should never fail, because we should verify private_keys are valid before making this call, can you update this code to return an error if one signing fails

crates/common/validator/src/validator.rs
```diff
@@ -261,6 +266,51 @@ impl ValidatorService {
...
+                {
+                    if let Ok(signature) = keystore.get().private_key.sign(signing_root.as_ref()) {
+                        return Some(SyncCommitteeRequestItem {
+                            slot,
+                            beacon_block_root,
+                            validator_index,
+                            signature,
+                        });
+                    } else {
+                        info!("invalid private key detected for validator at index {validator_index:?}")
```
KolbyML: you should never use info! for an error use warn! or error! depending on the context, it doesn't make sense here regardless because you should just be passing up an error

crates/common/validator/src/validator.rs
```diff
@@ -261,6 +266,50 @@ impl ValidatorService {
         Ok(())
     }
 
+    pub async fn submit_sync_committee(
+        &self,
+        slot: u64,
+        validator_indices: Vec<u64>,
```
KolbyML: Why did you change this to a vector?

KolbyML (approved):
:shipit: looks good

PR #567, opened 2025-06-13
feat: add "zkvm" flag that reduces VariableList to a manageable size by zkVMs

unnawut:

This PR adds "zkvm" feature flag so that external programs can config ream codebase in the way that can be run on zkVMs.

In this PR, the feature flag would drop `VariableList` that are 2<sup>40</sup> in size down to 2<sup>29</sup> when enabled, so that 32-bit zkVMs can work with our `BeaconState`.

See https://github.com/ReamLabs/consenzero-bench/pull/24 for how this feature flag is used.

The BeaconState contains some "zkvm" features where 32-bit zkVMs would fail on constructing a `VariableList` larger than 2<sup>32</sup> size (i.e. 2<sup>40</sup>). When "zkvm" feature is enabled, it would construct the BeaconState with 2<sup>29</sup> list instead.

To remain conformant with `consensus-specs`, this feature needs to be used with the modified `ssz_types` crate at https://github.com/ReamLabs/ssz_types/tree/magic-extended-list where the crate would detect 2<sup>29</sup> as a magic number when computing the root hash, and it will compute as a 2<sup>40</sup> list root instead.

Tbh I'm not sure I actually want to add this to ream's main codebase, but it's needed to run `ream-consensus` on zkVMs with correct results. If it's not ideal to merge, I'm happy to keep it as a long running `zkvm` branch on ream as well. Happy to discuss this and also in case anyone has a better idea to manage these differences.

### To-Do


crates/common/consensus/src/electra/beacon_state.rs
```diff
@@ -127,6 +129,14 @@ pub mod quoted_u8_var_list {
...
 }
 
+// The BeaconState contains some "zkvm" features where 32-bit zkVMs would fail
+// on constructing a VariableList larger than 2^32 size (i.e. 2^40). When "zkvm" feature
+// is enabled, it would construct the BeaconState with 2^29 list instead.
+//
+// This feature needs to be used with the modified ssz_types crate at
+// https://github.com/ReamLabs/ssz_types/tree/magic-extended-list
+// where the crate would detect 2^29 as a magic number when computing the root hash,
+// and it will compute as a 2^40 list root instead.
```
KolbyML: this should be a doc comment

crates/common/consensus/src/electra/beacon_state.rs
```diff
@@ -151,9 +161,16 @@ pub struct BeaconState {
     pub eth1_deposit_index: u64,
 
     // Registry
+    #[cfg(not(feature = "zkvm"))]
     pub validators: VariableList<Validator, U1099511627776>,
+    #[cfg(feature = "zkvm")]
+    pub validators: VariableList<Validator, U536870912>,
```
KolbyML: instead of doing it like this, could you pull out U1099511627776 and U536870912 define them as a type like `ValiatorLength` or whatever the constant is called in the spec, rather then defining optional fields. If you could apply this to the rest of the PR

```
    #[cfg(not(feature = "zkvm"))]
    type Hi = U32;
    #[cfg(feature = "zkvm")]
    type Hi = U16
    
    pub struct Car {
         pub validators: VariableList<Validator, Hi>,
    }
```

^ like this

KolbyML (commented):
Left some feedback

crates/common/consensus/src/electra/beacon_state.rs
KolbyML: In the future we should be able to configure this using a network preset like mainnet/minimal

crates/common/consensus/src/electra/beacon_state.rs
```diff
@@ -164,9 +173,9 @@ pub struct BeaconState {
 
     // Participation
     #[serde(with = "quoted_u8_var_list")]
-    pub previous_epoch_participation: VariableList<u8, U1099511627776>,
+    pub previous_epoch_participation: VariableList<u8, MaxAttestationsPerEpoch>,
```
jihoonsong: Any reason this doesn't use `ValidatorRegistryLimit`?

unnawut: Ah good catch! I was referring to [Phase0](https://ethereum.github.io/consensus-specs/specs/phase0/beacon-chain/) but realizing just now it was modified in [Altair](https://ethereum.github.io/consensus-specs/specs/altair/beacon-chain/#beaconstate)

PR #572, opened 2025-06-14
chore: clean up noisy logs + clean up peer connection code

KolbyML:

- The logs were so noisy I couldn't see anything
- We had a few problems in our peer connection logic

- remove too noisy logs
- clean up peer connection logic


Kayden-ML (approved):
Looks  good nice work 👍

PR #577, opened 2025-06-15
fix: failing to BlockBlock requests over p2p

KolbyML:

When I was trying to request blocks by range I was getting errors, due to the full payload not being ready and we would keep trying to decode the length every time we would check if we got enough data. This was leading to cases where we would either
- decode the current or next payload improperly

Also we were erroring if we got a `ErrorKind::UnexpectedEof`, but this is fine it just means we didn't get enough data yet

Only decode the length once per a piece of content

Don't error if we haven't gotten enough content yet


Kayden-ML (approved):
Looks good nice work 👍

PR #578, opened 2025-06-15
feat: remove excess Runtimes from 3 to 1 + properly shutdown runtime

KolbyML:

- We are currently spawning 3 tokio async runtimes, we should only be spawning 1 runtime

- Only spawn 1 runtime
- shutdown the tokio threads
- shutdown the runtime


Kayden-ML (approved):
Looks good nice work 👍

Issue #561, opened 2025-06-13
Implement make attestation logic in the validator


shariqnaiyer:
### Describe the feature

An attesting validator must perform it's duties as per the Beacon API spec:
https://ethereum.github.io/beacon-APIs/validator-flow.md

The implementation will include a function that can be called for attesting validators to perform their protocol duty end to end.

PR #582, opened 2025-06-16
feat: implement make attestation

shariqnaiyer:

Fixes: #561 

Implements the following portion of the [validator flow](https://ethereum.github.io/beacon-APIs/validator-flow.md): 
> 3. [Fetch AttestationData](#/ValidatorRequiredApi/produceAttestationData)
> 4. [Submit Attestation](#/ValidatorRequiredApi/submitPoolAttestations) (AttestationData + aggregation bits)
    - Aggregation bits are `Bitlist` with length of committee (received in AttesterDuty)
    with bit on position `validator_committee_index` (see AttesterDuty) set to true

I fetch an attestation data from the beacon node.
I then sign the attestation data and create a single attestation object that we use to submit an attestation.

### To-Do

 
### DIFF
--- a/crates/common/validator/src/validator.rs
+++ b/crates/common/validator/src/validator.rs
@@ -5,7 +5,7 @@ use std::{
 };
 
 use alloy_primitives::Address;
-use anyhow::anyhow;
+use anyhow::{anyhow, bail};
 use ream_beacon_api_types::{
     block::{BroadcastValidation, ProduceBlockData},
     duties::{AttesterDuty, ProposerDuty, SyncCommitteeDuty},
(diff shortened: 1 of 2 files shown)


crates/common/validator/src/validator.rs
```diff
@@ -310,6 +312,35 @@ impl ValidatorService {
...
+                    signature: sign_attestation_data(&attestation_data, &keystore.private_key)?,
+                    data: attestation_data,
+                }])
+                .await?)
+        } else {
+            Err(anyhow!(
+                "Keystore not found for validator: {validator_index}"
+            ))
+        }
+    }
```
KolbyML:
```suggestion
    pub async fn make_attestation(
        &self,
        slot: u64,
        validator_index: u64,
        committee_index: u64,
    ) -> anyhow::Result<()> {
        let Some(keystore) = self.validator_index_to_keystore.get(&validator_index) else {
            bail!("Keystore not found for validator: {validator_index}");
        }
   
        let attestation_data = self
            .beacon_api_client
            .get_attestation_data(slot, committee_index)
            .await?
            .data;

        Ok(self
            .beacon_api_client
            .submit_attestation(vec![SingleAttestation {
                attester_index: validator_index,
                committee_index,
                signature: sign_attestation_data(&attestation_data, &keystore.private_key)?,
                data: attestation_data,
            }])
            .await?)
    }
```

In the future if let Some should normally changed into let Some like I showed above

Issue #528, opened 2025-06-04
MEV Boost Builder


varun-doshi:
### Describe the feature

Implement support for MEV Boost

mw2000:
Should this be closed since this is implemented in #584 and connected to the beacon api via #1048

PR #584, opened 2025-06-16
feat: implement builder specs

varun-doshi:

First step to #528 

https://github.com/ethereum/builder-specs/tree/main


varun-doshi:
Ready for review @KolbyML

crates/common/builder/src/blinded_beacon_block.rs
```diff
@@ -0,0 +1,21 @@
...
+    pub parent_root: B256,
+    pub state_root: B256,
+    pub body: BlindedBeaconBlockBody,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
+pub struct SignedBlindedBeaconBlock {
+    pub message: BlindedBeaconBlock,
+    pub signature: BLSSignature,
+}
```
KolbyML: this code is already defined in the codebase, please removed this and use the types already defined

crates/common/builder/src/builder_api/header.rs
```diff
@@ -0,0 +1,24 @@
+use alloy_primitives::B256;
+use anyhow::Ok;
+use ream_bls::PubKey;
+
+use crate::{BuilderConfig, builder_bid::SignedBuilderBid};
+
+pub async fn get_builder_header(
+    config: BuilderConfig,
+    parent_hash: B256,
+    pubkey: &PubKey,
```
KolbyML:
```suggestion
pub async fn get_builder_header(
    config: BuilderConfig,
    parent_hash: B256,
    public_key: &PubKey,
```

we might as well fully spell public to be consistent and for readability, please update all other parts of your PR with this change

crates/common/builder/src/registration.rs
```diff
@@ -0,0 +1,81 @@
...
+    ensure!(
+        validator_pubkeys.contains(&registration.message.pubkey),
+        "Validator not found"
+    );
+
+    let index = validator_pubkeys
+        .iter()
+        .position(|k| *k == registration.message.pubkey)
+        .unwrap();
+    let validator = state.validators[index].clone();
```
KolbyML: Why is `unwrap()` here, use find instead of position here, so we don't need to do the raw index below

crates/common/builder/src/registration.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+pub fn process_registration(
+    state: BeaconState,
+    registration: SignedValidatorRegistrationV1,
+    registrations: HashMap<PubKey, ValidatorRegistrationV1>,
+    current_timestamp: u64,
+) -> anyhow::Result<()> {
+    let validator_pubkeys: Vec<PubKey> =
+        state.validators.iter().map(|v| v.pubkey.clone()).collect();
```
KolbyML:
```suggestion
    let validator_pubkeys =
        state.validators.iter().map(|validator| validator.pubkey.clone()).collect::<Vec<_>>();
```

crates/common/builder/src/verify.rs
```diff
@@ -0,0 +1,28 @@
...
+pub fn verify_blinded_block_signature(
+    state: BeaconState,
+    signed_block: SignedBlindedBeaconBlock,
+) -> anyhow::Result<bool> {
+    let proposer_index = state.get_beacon_proposer_index(Some(state.slot))?;
+
+    let proposer = state
+        .validators
+        .get(proposer_index as usize)
+        .ok_or(anyhow::anyhow!("Invalid proposer index"))?;
```
KolbyML:
```suggestion
pub fn verify_blinded_block_signature(
    state: BeaconState,
    signed_block: SignedBlindedBeaconBlock,
) -> anyhow::Result<bool> {
    let proposer_index = state.get_beacon_proposer_index(Some(state.slot))?;

    let proposer = state
        .validators
        .get(proposer_index as usize)
        .ok_or(anyhow!("Invalid proposer index"))?;
```

crates/common/consensus/src/constants.rs
```diff
@@ -16,6 +16,7 @@ pub const CHURN_LIMIT_QUOTIENT: u64 = 65536;
 pub const CURRENT_SYNC_COMMITTEE_INDEX: u64 = 22;
 pub const DEPOSIT_CONTRACT_TREE_DEPTH: u64 = 32;
 pub const DOMAIN_AGGREGATE_AND_PROOF: B32 = fixed_bytes!("0x06000000");
+pub const DOMAIN_APPLICATION_BUILDER: B32 = fixed_bytes!("0x00000001");
```
KolbyML: this const isn't used in this crate, so it shouldn't be defined here

crates/common/consensus/src/constants.rs
```diff
@@ -51,6 +52,7 @@ pub const MAX_DEPOSITS: u64 = 16;
 pub const MAX_SEED_LOOKAHEAD: u64 = 4;
 pub const MAX_PER_EPOCH_ACTIVATION_CHURN_LIMIT: u64 = 8;
 pub const MAX_RANDOM_VALUE: u64 = 65535;
+pub const MAX_REGISTRATION_LOOKAHEAD: u64 = 10;
```
KolbyML: this const isn't used in this crate, so it shouldn't be defined here

crates/common/builder/src/builder_api/header.rs
```diff
@@ -0,0 +1,24 @@
...
+        "/eth/v1/builder/header/{slot}/{parent_hash:?}/{pubkey:?}"
+    ))?;
+    let signed_blinded_bid = reqwest::Client::new()
+        .get(get_header_endpoint)
+        .send()
+        .await?
+        .json::<SignedBuilderBid>()
+        .await?;
+
+    Ok(signed_blinded_bid)
```
KolbyML:
```suggestion
    Ok(reqwest::Client::new()
        .get(get_header_endpoint)
        .send()
        .await?
        .json::<SignedBuilderBid>()
        .await?)
```

we might as well inline this, also could you make a struct for this function, so we can avoid reconstructing an HTTP client everytime we use this function

crates/common/builder/src/bid.rs
```diff
@@ -0,0 +1,75 @@
...
+        header,
+        blob_kzg_commitments: signed_blinded_bid.message.blob_kzg_commitments,
+        execution_requests: signed_blinded_bid.message.execution_requests,
+        value,
+        pubkey,
+    })
+}
+
+pub fn get_bid_signature(
+    _state: BeaconState,
```
KolbyML: why is state here if it isn't used

crates/common/builder/src/builder_bid.rs
```diff
@@ -0,0 +1,24 @@
...
+use ssz_types::{VariableList, typenum::U4096};
+use tree_hash_derive::TreeHash;
+
+#[derive(Debug, PartialEq, Eq, Clone, TreeHash, Serialize, Deserialize)]
+pub struct BuilderBid {
+    pub header: ExecutionPayloadHeader,
+    pub blob_kzg_commitments: VariableList<KZGCommitment, U4096>,
+    pub execution_requests: ExecutionRequests,
+    pub value: U256,
+    pub pubkey: PubKey,
```
KolbyML: use serde rename here

KolbyML (commented):
Here is some initial feedback. It would be nice if the PR's description was a bit more fleshed out.

crates/common/builder/src/builder_bid.rs
varun-doshi: Do you mean for pubkey=>public_key?

crates/common/builder/src/builder_api/header.rs
varun-doshi: Putting this under a single BuilderClient since there are other APIs which will come into the picture once I glue this with the validator client...let me know if that is okay

crates/common/builder/src/builder_bid.rs
KolbyML: use public_key for the field name, use serde rename so it serializes as pubkey

crates/common/builder/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+use alloy_primitives::{aliases::B32, fixed_bytes};
+use url::Url;
+
+pub mod bid;
+pub mod blobs;
+pub mod builder_bid;
+pub mod builder_client;
+pub mod registration;
+pub mod validator_registration;
+pub mod verify;
```
KolbyML: Normally we should target

- mod's
- import's
- consts
- ~types
- functions/structs/enum
- tests

crates/common/builder/src/builder_client.rs
```diff
@@ -0,0 +1,39 @@
...
+
+pub struct BuilderClient {
+    client: reqwest::Client,
+    mev_relay_url: Url,
+}
+
+impl BuilderClient {
+    pub fn new(config: BuilderConfig) -> anyhow::Result<Self> {
+        Ok(Self {
+            client: reqwest::Client::new(),
```
KolbyML:
```suggestion
pub struct BuilderClient {
    client: Client,
    mev_relay_url: Url,
}

impl BuilderClient {
    pub fn new(config: BuilderConfig) -> anyhow::Result<Self> {
        Ok(Self {
            client: Client::new(),
```

can't we just import Client

bin/ream/src/cli/validator_node.rs
```diff
@@ -58,6 +58,20 @@ pub struct ValidatorNodeConfig {
         help = "The password to use for keystores. It's recommended to use password-file over this in order to prevent your keystore password from appearing in the shell history"
     )]
     pub password: Option<String>,
+
+    #[arg(
+        long,
+        help = "Enable external block builder",
+        action(ArgAction::SetTrue)
+    )]
+    pub enable_builder: bool,
```
KolbyML:
```suggestion
    #[arg(
        long,
        help = "Enable external block builder",
    )]
    pub enable_builder: bool,
```

why would this be enabled by default?

crates/common/builder/src/bid.rs
```diff
@@ -0,0 +1,73 @@
...
+        value,
+        public_key,
+    })
+}
+
+pub fn get_bid_signature(bid: BuilderBid, private_key: PrivateKey) -> anyhow::Result<BLSSignature> {
+    let domain = compute_domain(DOMAIN_APPLICATION_BUILDER, None, None);
+    let signing_root = compute_signing_root(bid, domain);
+    Ok(private_key.sign(signing_root.as_ref())?)
+}
```
KolbyML: I don't think we should add this code. As isn't this code meant for the block builder? https://github.com/ethereum/builder-specs/blob/main/specs/bellatrix/builder.md

We are not building blocks, we are just asking the builder to build us the `validator` a block so shouldn't we only be implementing https://github.com/ethereum/builder-specs/blob/main/specs/bellatrix/validator.md

crates/common/builder/src/blobs.rs
```diff
@@ -0,0 +1,24 @@
...
+    pub blobs: VariableList<KZGCommitment, U1048576>,
+    pub commitments: VariableList<VariableList<u8, U96>, U1024>,
+    pub proofs: VariableList<VariableList<u8, U96>, U1024>,
+}
+
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
+pub struct ExecutionPayloadAndBlobsBundle {
+    pub execution_payload: ExecutionPayload,
+    pub blobs_bundle: BlobsBundle,
+}
```
KolbyML: this is for the builder and not apart of electra, wasn't this added in fulu so it should be removed for the same reason above?

crates/common/builder/src/builder_client.rs
```diff
@@ -0,0 +1,39 @@
...
+        ))?;
+
+        Ok(self
+            .client
+            .get(get_header_endpoint)
+            .send()
+            .await?
+            .json::<SignedBuilderBid>()
+            .await?)
+    }
```
KolbyML: shouldn't we implement 3 other endpoints?

- blinded_blocks
- status
- validators

crates/common/builder/src/registration.rs
```diff
@@ -0,0 +1,90 @@
...
+        );
+    }
+
+    ensure!(
+        verify_registration_signature(state, registration),
+        "Registration signature is invalid"
+    );
+
+    Ok(())
+}
```
KolbyML: We don't need these for the reason above?

bin/ream/src/cli/validator_node.rs
varun-doshi: It isn't enabled by default...ArgAction::SetTrue means if its passed it will enable, if it it is disabled
```
--enable-builder //enabled
```
else disabled

Correct me if Im wrong here

crates/common/builder/src/blobs.rs
varun-doshi: This was actually added in Deneb and is a response from the Builder API so we need it here

varun-doshi:
Interestingly the clippy fail in CI does not hit on my local machine...Pubkey is very much in the root in my local

bin/ream/src/cli/validator_node.rs
KolbyML: That is the default behavior so we don't need `action(ArgAction::SetTrue)`, I think we have discussed this same topic quite a few PR's back as well

crates/common/builder/src/blobs.rs
KolbyML: image

which endpoint returns this then?

crates/common/builder/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+pub mod builder_bid;
+pub mod builder_client;
+pub mod validator_registration;
+pub mod verify;
+
+use alloy_primitives::{aliases::B32, fixed_bytes};
+use url::Url;
+
+pub const DOMAIN_APPLICATION_BUILDER: B32 = fixed_bytes!("0x00000001");
+pub const MAX_REGISTRATION_LOOKAHEAD: u64 = 10;
```
KolbyML: this const isn't used anymore

crates/common/builder/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[derive(Debug, Clone)]
+pub struct BuilderConfig {
+    pub builder_enabled: bool,
+    pub mev_relay_url: Url,
+}
```
KolbyML: this is only used in one file, can you move this to the file that uses it

crates/common/builder/src/builder_client.rs
```diff
@@ -0,0 +1,121 @@
...
+            .client
+            .post(register_validator_endpoint)
+            .headers(self.get_header_with_json())
+            .json(&signed_registration)
+            .send()
+            .await?;
+
+        match response.status() {
+            StatusCode::OK => Ok(()),
+            StatusCode::BAD_REQUEST => Err(anyhow::anyhow!("unknown validator")),
```
KolbyML:
```suggestion
            StatusCode::BAD_REQUEST => Err(anyhow!("unknown validator")),
```

please address this everywhere else in the PR as well

crates/common/builder/src/builder_client.rs
```diff
@@ -0,0 +1,121 @@
...
+            StatusCode::UNSUPPORTED_MEDIA_TYPE => Err(anyhow::anyhow!("unsupported media type")),
+            StatusCode::INTERNAL_SERVER_ERROR => Err(anyhow::anyhow!("builder internal error")),
+            _ => Err(anyhow::anyhow!("internal error")),
```
KolbyML:
```suggestion
            err => Err(anyhow!("internal error: {err:?}")),
```

please do this in other parts of the PR as well

crates/common/builder/src/builder_client.rs
```diff
@@ -0,0 +1,121 @@
...
+use url::Url;
+
+use crate::{
+    BuilderConfig, blobs::ExecutionPayloadAndBlobsBundle, builder_bid::SignedBuilderBid,
+    validator_registration::SignedValidatorRegistrationV1,
+};
+
+pub const VERSION: &str = "electra";
+pub const ETH_CONSENSUS_VERSION_HEADER: &str = "Eth-Consensus-Version";
+pub const JSON_CONTENT_TYPE: &str = "application/json";
```
KolbyML: we shouldn't be redeclaring consts that already exist in the codebase

crates/common/builder/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+serde.workspace = true
+serde_json.workspace = true
+ssz_types.workspace = true
+tree_hash.workspace = true
+tree_hash_derive.workspace = true
+url.workspace = true
+
+# ream dependencies
+ream-bls.workspace = true
+ream-consensus.workspace = true
```
KolbyML: https://github.com/ReamLabs/ream/pull/584#pullrequestreview-2940463870

> I left some feedback, I don't think it makes sense for us to build a builder, so we should only be implementing the validator specs to support mev-relays no?

^ this earlier comment seems to have been missed.

I don't think there is really a reason to make this it's own crate, it is minimal enough it might as well be a module in the validator crate. It doesn't seem like there will be a usecase outside of the validator as well.

KolbyML (commented):
PR overall is looking good, just a few concerns to work through now

varun-doshi:
Unable to reply directly to your question above https://github.com/ReamLabs/ream/pull/584#discussion_r2159802425
But here you go, https://github.com/ethereum/builder-specs/blob/main/apis/builder/blinded_blocks.yaml
From deneb onwards, it returns `ExecutionPayloadAndBlobsBundle` which in turn has `BlobsBundle`

KolbyML:
> @KolbyML Unable to reply directly to your question above [#584 (comment)](https://github.com/ReamLabs/ream/pull/584#discussion_r2159802425) But here you go, https://github.com/ethereum/builder-specs/blob/main/apis/builder/blinded_blocks.yaml From deneb onwards, it returns `ExecutionPayloadAndBlobsBundle` which in turn has `BlobsBundle`

Thank you for sharing, that clears up my confusion for that concern, good catch :+1:

crates/common/validator/src/builder/bid.rs
```diff
@@ -0,0 +1,23 @@
+use alloy_primitives::Address;
+use anyhow::ensure;
+use ream_consensus::electra::beacon_state::BeaconState;
+
+use super::builder_bid::SignedBuilderBid;
+use crate::builder::verify::verify_bid_signature;
+
+pub fn process_bid(
+    state: BeaconState,
+    bid: SignedBuilderBid,
```
KolbyML: shouldn't these be references?

crates/common/validator/src/builder/builder_client.rs
```diff
@@ -0,0 +1,125 @@
...
+use crate::beacon_api_client::http_client::JSON_CONTENT_TYPE;
+
+#[derive(Debug, Clone)]
+pub struct BuilderConfig {
+    pub builder_enabled: bool,
+    pub mev_relay_url: Url,
+}
+
+pub struct BuilderClient {
+    client: Client,
```
KolbyML:
```suggestion
    client: ClientWithBaseUrl,
```

Could you use ClientWithBaseUrl here, just because we are reintroducing logic we already have here

Look here for example of usage https://github.com/ReamLabs/ream/blob/master/crates/common/validator/src/beacon_api_client/mod.rs

crates/common/validator/src/builder/builder_client.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub async fn get_builder_status(&self) -> anyhow::Result<()> {
+        let builder_statis_endpoint = self.mev_relay_url.join("/eth/v1/builder/status")?;
+
+        let response = self.client.get(builder_statis_endpoint).send().await?;
+        match response.status() {
+            StatusCode::OK => Ok(()),
+            StatusCode::INTERNAL_SERVER_ERROR => {
+                Err(anyhow!("internal error: builder internal error"))
+            }
+            _ => Err(anyhow!("failed to get builder status")),
```
KolbyML:
```suggestion
            status => Err(anyhow!("failed to get builder status: {status:?}")),
```

crates/common/validator/src/builder/builder_client.rs
```diff
@@ -0,0 +1,125 @@
...
+            .json(&signed_registration)
+            .send()
+            .await?;
+
+        match response.status() {
+            StatusCode::OK => Ok(()),
+            StatusCode::BAD_REQUEST => Err(anyhow!("unknown validator")),
+            StatusCode::UNSUPPORTED_MEDIA_TYPE => Err(anyhow!("unsupported media type")),
+            StatusCode::INTERNAL_SERVER_ERROR => Err(anyhow!("builder internal error")),
+            _ => Err(anyhow!("internal error")),
```
KolbyML:
```suggestion
            status => Err(anyhow!("internal error: {err:?}")),
```

crates/common/validator/src/builder/validator_registration.rs
```diff
@@ -0,0 +1,36 @@
...
+use tree_hash::TreeHash;
+use tree_hash_derive::TreeHash;
+
+use super::DOMAIN_APPLICATION_BUILDER;
+
+#[derive(Debug, PartialEq, Eq, Clone, TreeHash, Serialize, Deserialize)]
+pub struct ValidatorRegistrationV1 {
+    pub fee_recipient: Address,
+    pub gas_limit: u64,
+    pub timestamp: u64,
```
KolbyML: shouldn't we have the serde hex u64 thing here?

KolbyML (commented):
Here is some more feedback

crates/common/validator/src/builder/builder_client.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    pub fn get_header_with_json(&self) -> HeaderMap {
+        let mut headers = HeaderMap::new();
+        headers.insert(CONTENT_TYPE, HeaderValue::from_static(JSON_CONTENT_TYPE));
+        headers.insert(
+            ETH_CONSENSUS_VERSION_HEADER,
+            HeaderValue::from_static(VERSION),
+        );
+        headers
+    }
```
KolbyML: please remove this

crates/common/validator/src/builder/builder_client.rs
```diff
@@ -0,0 +1,132 @@
...
+    ) -> anyhow::Result<ExecutionPayloadAndBlobsBundle> {
+        let get_blinded_blocks_endpoint = self
+            .client
+            .base_url()
+            .join("/eth/v1/builder/blinded_blocks")?;
+
+        let response = self
+            .client
+            .post(get_blinded_blocks_endpoint, ContentType::Json)?
+            .headers(self.get_header_with_json())
```
KolbyML:
```suggestion
            .header(ETH_CONSENSUS_VERSION_HEADER, VERSION)
```

crates/common/validator/src/builder/builder_client.rs
```diff
@@ -0,0 +1,132 @@
...
+    ) -> anyhow::Result<()> {
+        let register_validator_endpoint = self
+            .client
+            .base_url()
+            .join("/eth/v1/builder/register_validator")?;
+
+        let response = self
+            .client
+            .post(register_validator_endpoint, ContentType::Json)?
+            .headers(self.get_header_with_json())
```
KolbyML:
```suggestion
            .header(ETH_CONSENSUS_VERSION_HEADER, VERSION)
```

crates/common/validator/src/builder/builder_client.rs
```diff
@@ -0,0 +1,132 @@
...
+        public_key: &PublicKey,
+        slot: u64,
+    ) -> anyhow::Result<SignedBuilderBid> {
+        let get_header_endpoint = self.client.base_url().join(&format!(
+            "/eth/v1/builder/header/{slot}/{parent_hash:?}/{public_key:?}"
+        ))?;
+
+        Ok(self
+            .client
+            .get(get_header_endpoint)?
```
KolbyML:
```
    pub fn get<U: IntoUrl>(&self, url: U) -> anyhow::Result<RequestBuilder> {
        let url = self.base_url.join(url.as_str())?;

        let mut headers = HeaderMap::new();
        match self.content_type {
            ContentType::Json => {
                headers.insert(CONTENT_TYPE, HeaderValue::from_static(JSON_CONTENT_TYPE));
            }
            ContentType::Ssz => {
                headers.insert(CONTENT_TYPE, HeaderValue::from_static(SSZ_CONTENT_TYPE));
                headers.insert(ACCEPT, HeaderValue::from_static(ACCEPT_PRIORITY));
            }
        }

        Ok(self.client.get(url).headers(headers))
    }

    pub fn post<U: IntoUrl>(
        &self,
        url: U,
        content_type: ContentType,
    ) -> anyhow::Result<RequestBuilder> {
        let url = self.base_url.join(url.as_str())?;

        Ok(self
            .client
            .post(url)
            .header(CONTENT_TYPE, content_type.to_header_value()))
    }
```

see get/post now do the base join for you so just

```suggestion
    /// Get an execution payload header.
    pub async fn get_builder_header(
        &self,
        parent_hash: B256,
        public_key: &PublicKey,
        slot: u64,
    ) -> anyhow::Result<SignedBuilderBid> {
        Ok(self
            .client
            .get(format!( "/eth/v1/builder/header/{slot}/{parent_hash:?}/{public_key:?}"))?
```

please apply this to the other endpoints you added

crates/common/validator/src/builder/builder_client.rs
varun-doshi: I guess we dont need to do this here since `.post(get_blinded_blocks_endpoint, ContentType::Json)` already does that internally?

KolbyML: post does content-type internally not `.header(ETH_CONSENSUS_VERSION_HEADER, VERSION)`

varun-doshi: Ahh yes , my bad

KolbyML (approved):
:shipit: PR looks great, thank you for your hard work :muscle:

PR #603, opened 2025-06-20
feat: removed abbreviated err and validators for better readability

Kayden-ML:

Currently we have cases where when we where calling a validator it was abbreviated to v this pr updates it to be more readable 

renaming v to validator and e to err


KolbyML (approved):
:shipit: looks good

PR #606, opened 2025-06-21
feat: use create_dir_all instead of create_dir

Kayden-ML:

Before I created a pr that stored the blobs in the data directory, instead of the database but i used create_dir which cause it to crash if the folder already exists.

I used create_dir_all instead

### To-Do

 

KolbyML (approved):
:shipit: looks good

PR #607, opened 2025-06-21
feat: fixed beacon node initialization from panicking

Kayden-ML:

when running the beacon node I was running into a tokio panic error due to trying to unwrap the execution engine which is only optionally initialized

fixed beacon node initialization by removing the expect


KolbyML (approved):
:shipit: looks good

Issue #610, opened 2025-06-23
Connect POST /eth/v1/beacon/pool/voluntary_exits to the Validator


shariqnaiyer:
### Describe the feature

An active validator can choose to voluntarily exit from the pool by submitting a voluntary exit to the beacon node. A validator that goes offline without submitting a voluntary exit will incur penalties for missing it's validator duties. 

A voluntary exit is accomplished by initially submitting a POST request to /eth/v1/beacon/pool/voluntary_exits to the Beacon node. The implementation will entail creating the the necessary structs as well as making the request as per the Beacon API Spec:

https://ethereum.github.io/beacon-APIs/#/Beacon/submitPoolVoluntaryExit

PR #611, opened 2025-06-23
feat: implement submit_voluntary_exit in the validator

shariqnaiyer:

Fixes: #610 

The request was implemented as per the Beacon API spec. Signing of the voluntary exit was implemented. Up next we may want to make a cli command for exiting with the exit flow (submit voluntary exit and monitor exit status till successfully exited).

### To-Do

 
### DIFF
--- a/crates/common/validator/src/validator.rs
+++ b/crates/common/validator/src/validator.rs
@@ -571,7 +571,6 @@ impl ValidatorService {
                 &self
                     .validator_index_to_keystore
                     .get(&validator_index)
-                    .cloned()
                     .ok_or_else(|| anyhow!("Keystore not found for validator: {validator_index}"))?
                     .private_key,
             )?)


crates/common/validator/src/validator.rs
```diff
@@ -380,6 +381,27 @@ impl ValidatorService {
...
+        self.beacon_api_client
+            .submit_signed_voluntary_exit(sign_voluntary_exit(
+                epoch,
+                validator_index,
+                &self
+                    .validator_index_to_keystore
+                    .get(&validator_index)
+                    .cloned()
+                    .ok_or_else(|| anyhow!("Keystore not found for validator: {validator_index}"))?
+                    .private_key,
```
KolbyML: why are we cloning the private key if we only need a reference?

crates/common/validator/src/voluntary_exit.rs
```diff
@@ -0,0 +1,36 @@
...
+                    &voluntary_exit,
+                    compute_domain(
+                        DOMAIN_VOLUNTARY_EXIT,
+                        Some(network_spec().electra_fork_version),
+                        None,
+                    ),
+                )
+                .as_ref(),
+            )
+            .map_err(|e| anyhow!("Failed to sign voluntary exit: {e}"))?,
```
KolbyML:
```suggestion
            .map_err(|err| anyhow!("Failed to sign voluntary exit: {err}"))?,
```

PR #613, opened 2025-06-23
feat: add endpoint for blinded block

### LINKED ISSUE #207: Implement /eth/v1/beacon/blinded_blocks/{block_id}
https://ethereum.github.io/beacon-APIs/#/Beacon/getBlindedBlock

aditya172926:

 
Fixes #207 

Taking most of the data from `SignedBeaconBlock` and using `execution_payload_header` from its execution payload. Then checking for the request header if `application/octet-stream` returning a string response instead of json.

### To-Do

 
### DIFF
--- a/crates/common/consensus/src/electra/beacon_block.rs
+++ b/crates/common/consensus/src/electra/beacon_block.rs
@@ -10,6 +10,10 @@ use super::beacon_block_body::BeaconBlockBody;
 use crate::{
     beacon_block_header::{BeaconBlockHeader, SignedBeaconBlockHeader},
     blob_sidecar::BlobSidecar,
+    electra::{
+        blinded_beacon_block::{BlindedBeaconBlock, SignedBlindedBeaconBlock},
+        blinded_beacon_block_body::BlindedBeaconBlockBody,
+    },
     execution_engine::rpc_types::get_blobs::{Blob, BlobAndProofV1},
     polynomial_commitments::kzg_proof::KZGProof,
 };
@@ -70,6 +74,37 @@ impl SignedBeaconBlock {
                 self.blob_sidecar(BlobAndProofV1 { blob, proof }, index as u64)
             })
             .collect::<anyhow::Result<Vec<_>>>()
+    }
+
+    pub fn to_signed_blinded_beacon_block(&self) -> anyhow::Result<SignedBlindedBeaconBlock> {
+        Ok(SignedBlindedBeaconBlock {
+            message: BlindedBeaconBlock {
+                slot: self.message.slot,
+                proposer_index: self.message.proposer_index,
+                parent_root: self.message.parent_root,
+                state_root: self.message.state_root,
+                body: BlindedBeaconBlockBody {
+                    randao_reveal: self.message.body.randao_reveal.clone(),
+                    eth1_data: self.message.body.eth1_data.clone(),
+                    graffiti: self.message.body.graffiti,
+                    proposer_slashings: self.message.body.proposer_slashings.clone(),
+                    attester_slashings: self.message.body.attester_slashings.clone(),
+                    attestations: self.message.body.attestations.clone(),
+                    deposits: self.message.body.deposits.clone(),
+                    voluntary_exits: self.message.body.voluntary_exits.clone(),
+                    sync_aggregate: self.message.body.sync_aggregate.clone(),
+                    execution_payload_header: self
+                        .message
+                        .body
+                        .execution_payload
+                        .to_execution_payload_header(),
+                    bls_to_execution_changes: self.message.body.bls_to_execution_changes.clone(),
+                    blob_kzg_commitments: self.message.body.blob_kzg_commitments.clone(),
+                    execution_requests: self.message.body.execution_requests.clone(),
+                },
+            },
+            signature: self.signature.clone(),
+        })
     }
 }
 
--- a/crates/common/consensus/src/electra/beacon_block.rs
+++ b/crates/common/consensus/src/electra/beacon_block.rs
@@ -76,8 +76,8 @@ impl SignedBeaconBlock {
             .collect::<anyhow::Result<Vec<_>>>()
     }
 
-    pub fn to_signed_blinded_beacon_block(&self) -> anyhow::Result<SignedBlindedBeaconBlock> {
-        Ok(SignedBlindedBeaconBlock {
+    pub fn as_signed_blinded_beacon_block(&self) -> SignedBlindedBeaconBlock {
+        SignedBlindedBeaconBlock {
             message: BlindedBeaconBlock {
                 slot: self.message.slot,
                 proposer_index: self.message.proposer_index,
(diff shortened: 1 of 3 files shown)


crates/rpc/src/handlers/block.rs
```diff
@@ -330,3 +336,52 @@ pub async fn get_beacon_heads(db: Data<ReamDB>) -> Result<impl Responder, ApiErr
...
     Ok(HttpResponse::Ok().json(DataResponse::new(leaves)))
 }
+
+#[get("/beacon/blind_block/{block_id}")]
+pub async fn get_blind_block(
+    req: HttpRequest,
+    db: Data<ReamDB>,
+    block_id: Path<ID>,
+) -> Result<impl Responder, ApiError> {
+    // This endpoint is not implemented yet.
```
KolbyML: I don't think this comment is supposed to be here

crates/rpc/src/handlers/block.rs
```diff
@@ -330,3 +336,52 @@ pub async fn get_beacon_heads(db: Data<ReamDB>) -> Result<impl Responder, ApiErr
...
+    let beacon_block: SignedBeaconBlock =
+        get_beacon_block_from_id(block_id.into_inner(), &db).await?;
```
KolbyML:
```suggestion
    let beacon_block = get_beacon_block_from_id(block_id.into_inner(), &db).await?;
```

crates/rpc/src/handlers/block.rs
```diff
@@ -330,3 +336,52 @@ pub async fn get_beacon_heads(db: Data<ReamDB>) -> Result<impl Responder, ApiErr
...
+                    .body
+                    .execution_payload
+                    .to_execution_payload_header(),
+                bls_to_execution_changes: beacon_block.message.body.bls_to_execution_changes,
+                blob_kzg_commitments: beacon_block.message.body.blob_kzg_commitments,
+                execution_requests: beacon_block.message.body.execution_requests,
+            },
+        },
+        signature: beacon_block.signature.clone(),
+    };
```
KolbyML: make a function on SignedBeaconBlock called `toSignedBlindedBeaconBlock`

crates/rpc/src/handlers/block.rs
```diff
@@ -330,3 +336,52 @@ pub async fn get_beacon_heads(db: Data<ReamDB>) -> Result<impl Responder, ApiErr
...
+    if let Some(header) = req.headers().get("application/octet-stream") {
+        if header.to_str().unwrap_or_default() == "application/octet-stream" {
+            let ssz_response = Bytes::from(signed_blind_block.as_ssz_bytes()).to_string();
+            return Ok(HttpResponse::Ok()
+                .content_type("application/octet-stream")
+                .body(ssz_response));
+        }
+    }
+
+    Ok(HttpResponse::Ok().json(BeaconVersionedResponse::new(signed_blind_block)))
```
KolbyML: could you do a match case like done in other parts of the code, also please use the consts for `application/octet-stream`

KolbyML (commented):
Left some feedback

crates/rpc/src/handlers/block.rs
```diff
@@ -26,10 +26,13 @@ use ream_storage::{
     tables::{Field, Table},
 };
 use serde::{Deserialize, Serialize};
+use ssz::Encode;
 use tracing::error;
 
 use crate::handlers::state::get_state_from_id;
 
+pub const OCTET_STREAM_HEADER: &str = "application/octet-stream";
```
KolbyML: we already have a const in the codebase for this right value, could you reuse it

crates/common/consensus/src/electra/beacon_block.rs
```diff
@@ -71,6 +75,37 @@ impl SignedBeaconBlock {
             })
             .collect::<anyhow::Result<Vec<_>>>()
     }
+
+    pub fn to_signed_blinded_beacon_block(&self) -> anyhow::Result<SignedBlindedBeaconBlock> {
+        Ok(SignedBlindedBeaconBlock {
```
KolbyML:
```suggestion
    pub fn as_signed_blinded_beacon_block(&self) -> SignedBlindedBeaconBlock {
        SignedBlindedBeaconBlock {
```

I think we should call this as, as we are only taking a reference to self, we aren't converting self to `SignedBlindedBeaconBlock`

also we don't need to result here so please remove it

crates/rpc/src/handlers/block.rs
```diff
@@ -338,3 +341,31 @@ pub async fn get_beacon_heads(db: Data<ReamDB>) -> Result<impl Responder, ApiErr
...
+                    .body(ssz_response))
+            }
+            _ => Ok(HttpResponse::Ok().json(BeaconVersionedResponse::new(signed_blind_block))),
+        }
+    } else {
+        Err(ApiError::InternalError(
+            "Failed to convert beacon block to signed blinded beacon block".to_string(),
+        ))
+    }
+}
```
KolbyML:
```suggestion
#[get("/beacon/blind_block/{block_id}")]
pub async fn get_blind_block(
    http_request: HttpRequest,
    db: Data<ReamDB>,
    block_id: Path<ID>,
) -> Result<impl Responder, ApiError> {
    let beacon_block = get_beacon_block_from_id(block_id.into_inner(), &db).await?;
    let blinded_beacon_block = beacon_block.as_signed_blinded_beacon_block() 
    match http_request
        .headers()
        .get(OCTET_STREAM_HEADER)
        .and_then(|header| header.to_str().ok())
    {
        Some(OCTET_STREAM_HEADER) => {
            Ok(HttpResponse::Ok()
                .content_type(OCTET_STREAM_HEADER)
                .body(blinded_beacon_block.as_ssz_bytes()))
        }
        _ => Ok(HttpResponse::Ok().json(BeaconVersionedResponse::new(blinded_beacon_block))),
    }
}
```

The code should look like this, make sure you update the constant though

KolbyML (commented):
Left some feedback

crates/rpc/src/handlers/block.rs
```diff
@@ -338,3 +340,23 @@ pub async fn get_beacon_heads(db: Data<ReamDB>) -> Result<impl Responder, ApiErr
...
+
+#[get("/beacon/blind_block/{block_id}")]
+pub async fn get_blind_block(
+    req: HttpRequest,
+    db: Data<ReamDB>,
+    block_id: Path<ID>,
+) -> Result<impl Responder, ApiError> {
+    let beacon_block = get_beacon_block_from_id(block_id.into_inner(), &db).await?;
+    let blinded_beacon_block = beacon_block.as_signed_blinded_beacon_block();
+    match req
```
KolbyML:
```suggestion
    http_request: HttpRequest,
    db: Data<ReamDB>,
    block_id: Path<ID>,
) -> Result<impl Responder, ApiError> {
    let beacon_block = get_beacon_block_from_id(block_id.into_inner(), &db).await?;
    let blinded_beacon_block = beacon_block.as_signed_blinded_beacon_block();
    match http_request
```

KolbyML (approved):
After this last concern is resolved I will merge the PR

PR #618, opened 2025-06-24
feat: forward and validate beacon block

### LINKED ISSUE #589: gossipsub validate and forward: beacon_block
### Describe the feature

Validate each received `beacon_block` message and forward it upon successful validation.

#### Acceptance criteria:
- Each received `beacon_block` gossipsub message is validated according to the following specs:
  - https://ethereum.github.io/consensus-specs/specs/phase0/p2p-interface/#beacon_block
  - https://ethereum.github.io/consensus-specs/specs/altair/p2p-interface/#beacon_block
  - https://ethereum.github.io/consensus-specs/specs/bellatrix/p2p-interface/#beacon_block
  - https://ethereum.github.io/consensus-specs/specs/capella/p2p-interface/#beacon_block
  - https://ethereum.github.io/consensus-specs/specs/deneb/p2p-interface/#beacon_block
  - https://ethereum.github.io/consensus-specs/specs/electra/p2p-interface/#beacon_block
- The message is dropped if it matches at least one `[IGNORE]` or `[REJECT]` rule
- The message is forwarded if it does not match any `[IGNORE]` or `[REJECT]` rule

### Additional context

_No response_

varun-doshi:

Fixes #589 

https://ethereum.github.io/consensus-specs/specs/phase0/p2p-interface/#beacon_block
### To-Do

 
### DIFF
--- a/crates/storage/src/db.rs
+++ b/crates/storage/src/db.rs
@@ -1,12 +1,14 @@
 use std::{fs, io, path::PathBuf, sync::Arc};
 
-use anyhow::Result;
+use anyhow::{Result, anyhow};
+use ream_consensus::electra::beacon_block::SignedBeaconBlock;
 use redb::{Builder, Database};
 use tracing::info;
 
 use crate::{
     errors::StoreError,
     tables::{
+        Table,
         beacon_block::{BEACON_BLOCK_TABLE, BeaconBlockTable},
         beacon_state::{BEACON_STATE_TABLE, BeaconStateTable},
         blobs_and_proofs::{BLOB_FOLDER_NAME, BlobsAndProofsTable},
@@ -194,6 +196,20 @@ impl ReamDB {
             Ok(Some(slot)) => slot > 0,
             _ => false,
         }
+    }
+    pub fn get_latest_block(&self) -> anyhow::Result<SignedBeaconBlock> {
+        let slot = self
+            .slot_index_provider()
+            .get_highest_root()
+            .map_err(|err| anyhow!("Failed to get latest root, error: {err:?}"))?
+            .ok_or_else(|| anyhow!("Unable to fetch latest root"))?;
+
+        let latest_block = self
+            .beacon_block_provider()
+            .get(slot)?
+            .ok_or_else(|| anyhow!("Unable to fetch latest block"))?;
+
+        Ok(latest_block)
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5384,6 +5384,8 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "discv5",
+ "ethereum_ssz",
+ "ethereum_ssz_derive",
  "libp2p",
  "ream-beacon-chain",
  "ream-consensus",
@@ -5453,6 +5455,7 @@ dependencies = [
  "ream-executor",
  "ream-light-client",
  "ream-network-spec",
+ "ream-storage",
  "ream-validator",
  "serde",
  "serde_yaml",
--- a/crates/networking/manager/Cargo.toml
+++ b/crates/networking/manager/Cargo.toml
@@ -12,6 +12,8 @@ version.workspace = true
 [dependencies]
 anyhow.workspace = true
 discv5.workspace = true
+ethereum_ssz.workspace = true
+ethereum_ssz_derive.workspace = true
 libp2p.workspace = true
 tokio.workspace = true
 tracing.workspace = true
--- a/crates/networking/manager/src/p2p_sender.rs
+++ b/crates/networking/manager/src/p2p_sender.rs
@@ -1,7 +1,7 @@
 use anyhow::anyhow;
 use libp2p::{PeerId, swarm::ConnectionId};
 use ream_p2p::{
-    channel::{P2PMessage, P2PResponse},
+    channel::{GossipMessage, P2PMessage, P2PResponse},
     req_resp::{error::ReqRespError, handler::RespMessage, messages::ResponseMessage},
 };
 use tokio::sync::mpsc;
@@ -10,6 +10,12 @@ use tracing::warn;
 pub struct P2PSender(pub mpsc::UnboundedSender<P2PMessage>);
 
 impl P2PSender {
+    pub fn send_gossip(&self, message: GossipMessage) {
+        if let Err(err) = self.0.send(P2PMessage::Gossip(message)) {
+            warn!("Failed to send gossip message: {err}");
+        }
+    }
+
     pub fn send_response(
         &self,
         peer_id: PeerId,
--- a/crates/networking/p2p/Cargo.toml
+++ b/crates/networking/p2p/Cargo.toml
@@ -43,4 +43,5 @@ ream-discv5.workspace = true
 ream-executor.workspace = true
 ream-light-client.workspace = true
 ream-network-spec.workspace = true
+ream-storage.workspace = true
 ream-validator.workspace = true
--- a/crates/networking/p2p/src/channel.rs
+++ b/crates/networking/p2p/src/channel.rs
@@ -1,9 +1,12 @@
 use libp2p::{PeerId, swarm::ConnectionId};
 use tokio::sync::mpsc;
 
-use crate::req_resp::{
-    handler::RespMessage,
-    messages::{ResponseMessage, status::Status},
+use crate::{
+    gossipsub::topics::GossipTopic,
+    req_resp::{
+        handler::RespMessage,
+        messages::{ResponseMessage, status::Status},
+    },
 };
 
 pub enum P2PCallbackResponse {
@@ -14,6 +17,7 @@ pub enum P2PCallbackResponse {
 pub enum P2PMessage {
     Request(P2PRequest),
     Response(P2PResponse),
+    Gossip(GossipMessage),
 }
 
 pub enum P2PRequest {
@@ -34,4 +38,10 @@ pub struct P2PResponse {
     pub connection_id: ConnectionId,
     pub stream_id: u64,
     pub message: RespMessage,
+}
+
+#[derive(Debug, Clone)]
+pub struct GossipMessage {
+    pub topic: GossipTopic,
+    pub data: Vec<u8>,
 }
--- a/crates/networking/p2p/src/gossipsub/error.rs
+++ b/crates/networking/p2p/src/gossipsub/error.rs
@@ -4,6 +4,8 @@ pub enum GossipsubError {
     InvalidData(String),
     #[error("Gossipsub invalid topic {0}")]
     InvalidTopic(String),
+    #[error("Gossipsub validation failed {0:?}")]
+    ValidationFailed(String),
 }
 
 impl From<ssz::DecodeError> for GossipsubError {
--- a/crates/networking/p2p/src/gossipsub/mod.rs
+++ b/crates/networking/p2p/src/gossipsub/mod.rs
@@ -5,6 +5,7 @@ pub mod error;
 pub mod message;
 pub mod snappy;
 pub mod topics;
+pub mod validation;
 
 use libp2p::gossipsub::{AllowAllSubscriptionFilter, Behaviour};
 use snappy::SnappyTransform;
--- a/crates/networking/p2p/src/gossipsub/topics.rs
+++ b/crates/networking/p2p/src/gossipsub/topics.rs
@@ -1,10 +1,12 @@
+use GossipTopicKind::*;
 use alloy_primitives::{
     aliases::B32,
     hex::{FromHex, ToHexExt},
 };
 use libp2p::gossipsub::{IdentTopic as Topic, TopicHash};
+use ream_storage::db::ReamDB;
 
-use super::error::GossipsubError;
+use super::{error::GossipsubError, validation::validate_beacon_block};
 
 pub const TOPIC_PREFIX: &str = "eth2";
 pub const ENCODING_POSTFIX: &str = "ssz_snappy";
@@ -84,6 +86,30 @@ impl GossipTopic {
 
         Ok(GossipTopic { fork, kind })
     }
+
+    pub fn validate_gossip_message(
+        &self,
+        ssz_object: &Vec<u8>,
+        db: &ReamDB,
+    ) -> Result<(), GossipsubError> {
+        match self.kind {
+            GossipTopicKind::BeaconBlock => validate_beacon_block(ssz_object, db)
+                .map_err(|err| GossipsubError::ValidationFailed(err.to_string()))?,
+            GossipTopicKind::AggregateAndProof => todo!(),
+            GossipTopicKind::VoluntaryExit => todo!(),
+            GossipTopicKind::ProposerSlashing => todo!(),
+            GossipTopicKind::AttesterSlashing => todo!(),
+            GossipTopicKind::BeaconAttestation(_) => todo!(),
+            GossipTopicKind::SyncCommittee(_) => todo!(),
+            GossipTopicKind::SyncCommitteeContributionAndProof => todo!(),
+            GossipTopicKind::BlsToExecutionChange => todo!(),
+            GossipTopicKind::LightClientFinalityUpdate => todo!(),
+            GossipTopicKind::LightClientOptimisticUpdate => todo!(),
+            GossipTopicKind::BlobSidecar(_) => todo!(),
+        }
+
+        Ok(())
+    }
 }
 
 impl std::fmt::Display for GossipTopic {
@@ -108,6 +134,60 @@ impl From<GossipTopic> for Topic {
 impl From<GossipTopic> for String {
     fn from(topic: GossipTopic) -> Self {
         topic.to_string()
+    }
+}
+
+impl Into<TopicHash> for GossipTopic {
+    fn into(self) -> TopicHash {
+        let kind_str = match &self.kind {
+            BeaconBlock => BEACON_BLOCK_TOPIC,
+            AggregateAndProof => BEACON_AGGREGATE_AND_PROOF_TOPIC,
+            VoluntaryExit => VOLUNTARY_EXIT_TOPIC,
+            ProposerSlashing => PROPOSER_SLASHING_TOPIC,
+            AttesterSlashing => ATTESTER_SLASHING_TOPIC,
+            SyncCommitteeContributionAndProof => SYNC_COMMITTEE_CONTRIBUTION_AND_PROOF_TOPIC,
+            BlsToExecutionChange => BLS_TO_EXECUTION_CHANGE_TOPIC,
+            LightClientFinalityUpdate => LIGHT_CLIENT_FINALITY_UPDATE_TOPIC,
+            LightClientOptimisticUpdate => LIGHT_CLIENT_OPTIMISTIC_UPDATE_TOPIC,
+            BeaconAttestation(index) => {
+                return TopicHash::from_raw(format!(
+                    "/{}/{}/{}{}{}",
+                    TOPIC_PREFIX,
+                    self.fork.encode_hex(),
+                    BEACON_ATTESTATION_PREFIX,
+                    index,
+                    ENCODING_POSTFIX,
+                ));
+            }
+            SyncCommittee(index) => {
+                return TopicHash::from_raw(format!(
+                    "/{}/{}/{}{}{}",
+                    TOPIC_PREFIX,
+                    self.fork.encode_hex(),
+                    SYNC_COMMITTEE_PREFIX_TOPIC,
+                    index,
+                    ENCODING_POSTFIX,
+                ));
+            }
+            BlobSidecar(index) => {
+                return TopicHash::from_raw(format!(
+                    "/{}/{}/{}{}{}",
+                    TOPIC_PREFIX,
+                    self.fork.encode_hex(),
+                    BLOB_SIDECAR_PREFIX_TOPIC,
+                    index,
+                    ENCODING_POSTFIX,
+                ));
+            }
+        };
+
+        TopicHash::from_raw(format!(
+            "/{}/{}/{}{}",
+            TOPIC_PREFIX,
+            self.fork.encode_hex(),
+            kind_str,
+            ENCODING_POSTFIX
+        ))
     }
 }
 
--- a/crates/networking/p2p/src/gossipsub/validation.rs
+++ b/crates/networking/p2p/src/gossipsub/validation.rs
@@ -1,0 +1,30 @@
+use anyhow::{anyhow, ensure};
+use ream_consensus::{electra::beacon_block::SignedBeaconBlock, misc::compute_start_slot_at_epoch};
+use ream_storage::{db::ReamDB, tables::Field};
+use ssz::Decode;
+
+pub fn validate_beacon_block(block: &Vec<u8>, db: &ReamDB) -> anyhow::Result<()> {
+    let block =
+        SignedBeaconBlock::from_ssz_bytes(&block).map_err(|err| anyhow!(format!("{err:?}")))?;
+
+    let latest_block_in_db = db
+        .get_latest_block()
+        .map_err(|err| anyhow!(err.to_string()))?;
+
+    ensure!(
+        block.message.slot > latest_block_in_db.message.slot,
+        "Block slot must be greater than latest block slot in db"
+    );
+
+    let start_slot_at_epoch =
+        compute_start_slot_at_epoch(db.finalized_checkpoint_provider().get()?.epoch);
+    ensure!(
+        block.message.slot >= start_slot_at_epoch,
+        "Block slot must be greater than start slot at epoch"
+    );
+
+    // todo:The block is the first block with valid signature received for the proposer for the
+    // slot, signed_beacon_block.message.slot.
+
+    Ok(())
+}
--- a/crates/networking/p2p/src/network.rs
+++ b/crates/networking/p2p/src/network.rs
@@ -350,6 +350,11 @@ impl Network {
                         P2PMessage::Response(P2PResponse {peer_id, connection_id, stream_id, message}) => {
                             self.swarm.behaviour_mut().req_resp.send_response(peer_id, connection_id, stream_id, message)
                         },
+                        P2PMessage::Gossip(message) => {
+                            if let Err(err) = self.swarm.behaviour_mut().gossipsub.publish(message.topic, message.data) {
+                                warn!("Failed to publish gossip message: {err}");
+                            }
+                        }
                     }
                 }
                 Some(Ok(peer_id)) = self.peers_to_ping.next() => {


crates/common/beacon_chain/src/beacon_chain.rs
```diff
@@ -112,4 +121,112 @@ impl BeaconChain {
             head_slot,
         })
     }
+
+    pub async fn validate_beacon_block(&self, block: &SignedBeaconBlock) -> anyhow::Result<()> {
```
unnawut: I think this `validate_beacon_block()` function only makes sense in the context of gossipsub message validation. So I'd suggest moving this to p2p crate.

crates/common/beacon_chain/src/beacon_chain.rs
```diff
@@ -112,4 +121,112 @@ impl BeaconChain {
...
+        let store = self.store.lock().await;
+
+        let latest_block_in_db = store.db.get_latest_block()?;
+        let latest_state_in_db = store.db.get_latest_state()?;
+
+        ensure!(
+            block.message.slot > latest_block_in_db.message.slot,
+            "Block slot must be greater than latest block slot in db"
+        );
```
unnawut: The specs differentiate between `[IGNORE]` and `[REJECT]`, where IGNORE just drops the message but REJECT should also penalize the sender peer.

I'm not sure if we have peer scoring in place yet, but at least this logic should return the status accordingly.

crates/networking/manager/src/gossipsub.rs
```diff
@@ -78,9 +88,18 @@ pub async fn handle_gossipsub_message(message: Message, beacon_chain: &BeaconCha
                     signed_block.message.block_root()
                 );
 
-                if let Err(err) = beacon_chain.process_block(*signed_block).await {
+                if let Err(err) = beacon_chain.validate_beacon_block(&signed_block).await {
+                    error!("Failed to validate gossipsub beacon block: {err}");
+                }
+
+                if let Err(err) = beacon_chain.process_block(*signed_block.clone()).await {
```
unnawut: It looks like `process_block()` gets called regardless of `validate_beacon_block()` result...

crates/networking/manager/src/gossipsub.rs
```diff
@@ -78,9 +88,18 @@ pub async fn handle_gossipsub_message(message: Message, beacon_chain: &BeaconCha
...
                     error!("Failed to process gossipsub beacon block: {err}");
                 }
+                p2psender.send_gossip(GossipMessage {
+                    topic: GossipTopic::from_topic_hash(&message.topic)
+                        .expect("invalid topic hash"),
+                    data: signed_block.as_ssz_bytes(),
+                });
```
unnawut: Forwarding is also regardless of the validation result

crates/networking/manager/src/service.rs
```diff
@@ -152,7 +152,7 @@ impl NetworkManagerService {
                     match event {
                         // Handles Gossipsub messages from other peers.
                         ReamNetworkEvent::GossipsubMessage { message } =>
-                            handle_gossipsub_message(message, &beacon_chain).await,
+                            handle_gossipsub_message(message, &beacon_chain,&p2p_sender).await,
```
unnawut: nitpick
```suggestion
                            handle_gossipsub_message(message, &beacon_chain, &p2p_sender).await,
```

crates/common/beacon_chain/src/beacon_chain.rs
varun-doshi: This was initially in the p2p crate but i remember i encountered circular dependency errors and there's no way to mitigate that in rust hence I thought this was the best place since it's a direct impl for BeaconChain struct

crates/common/beacon_chain/src/beacon_chain.rs
varun-doshi: I'll be taking up peer scoring after the set of these issues are over... would be best to accomodate ignore/reject there I think
Let me know what you think though

crates/networking/manager/src/gossipsub.rs
varun-doshi: Ahh yes...good catch

crates/common/beacon_chain/src/beacon_chain.rs
varun-doshi: Created for the same https://github.com/ReamLabs/ream/issues/622

crates/common/beacon_chain/src/beacon_chain.rs
varun-doshi: I'll still recheck this again

crates/common/beacon_chain/src/beacon_chain.rs
unnawut: I agree that peer scoring should go in a separate issue. 

But I think we should still distinguish IGNORE vs REJECT in this PR though as it's part of the specs this PR is tackling.

varun-doshi: Understood...I'll make the necessary changes

KolbyML: > I agree that peer scoring should go in a separate issue.

> But I think we should still distinguish IGNORE vs REJECT in this PR though as it's part of the specs this PR is tackling.

^ I agree with this Peer scoring is pretty broad, I don't expect it to be done in 1 PR, more so something we slowly iterate on overtime over many issues.

crates/common/beacon_chain/src/beacon_chain.rs
unnawut: Ah I just had another look at the p2p crate, it looks quite purely networking stuff, so I think it makes more sense to keep it here actually.

But I think having it as `ream_beacon_chain::BeaconChain::validate_beacon_block()` is quite misleading, it's only validating in the sense of gossipsub forwarding rule. Could we at least move it to something like `ream_beacon_chain::gossipsub::validate_beacon_block()`?

varun-doshi: I think I should be able to move it to the network manager crate without any dep errors. What do you think @unnawut

KolbyML: > I think I should be able to move it to the network manager crate without any dep errors. What do you think @unnawut

I think it should probably be in the network manager crate. The Beacon Chain crate is moreso for managing the chain of blocks, re-orgs, fork choice, and on_* event's and such

unnawut: Agree let’s move to network manager!

crates/common/beacon_chain/src/beacon_chain.rs
```diff
@@ -17,13 +19,15 @@ use ream_storage::{
...
-use tokio::sync::Mutex;
+use tokio::sync::{Mutex, RwLock};
 use tracing::warn;
 
 /// BeaconChain is the main struct which manages the nodes local beacon chain.
 pub struct BeaconChain {
     pub store: Mutex<Store>,
     pub execution_engine: Option<ExecutionEngine>,
+    pub cached_proposer_signature: RwLock<HashMap<(PublicKey, u64), BLSSignature>>,
+    pub cached_bls_to_execution_signature: RwLock<HashMap<(PublicKey, u64), BLSToExecutionChange>>,
```
KolbyML: shouldn't this be moved out of the BeaconChain struct? as they aren't used here

crates/networking/manager/src/validate.rs
```diff
@@ -0,0 +1,123 @@
+use anyhow::{Ok, anyhow};
```
KolbyML:
```suggestion
use anyhow::{anyhow};
```

we shouldn't be importing OK

crates/networking/manager/src/validate.rs
```diff
@@ -0,0 +1,123 @@
...
+    constants::MAX_BLOBS_PER_BLOCK_ELECTRA, electra::beacon_block::SignedBeaconBlock,
+    misc::compute_start_slot_at_epoch,
+};
+use ream_storage::tables::{Field, Table};
+
+#[derive(Debug)]
+pub enum ValidationResult {
+    Accept,
+    Ignore,
+    Reject,
```
KolbyML: can you make gossipsub.rs into a folder then move this file their, but then make a folder called validate, and make a file called beacon_block and put `validate_beacon_block` there

crates/networking/manager/src/validate.rs
```diff
@@ -0,0 +1,123 @@
...
+pub async fn validate_beacon_block(
+    beacon_chain: &BeaconChain,
+    block: &SignedBeaconBlock,
+) -> anyhow::Result<ValidationResult> {
+    let store = beacon_chain.store.lock().await;
+
+    let latest_block_in_db = store.db.get_latest_block()?;
+    let latest_state_in_db = store.db.get_latest_state()?;
+
+    if block.message.slot < latest_block_in_db.message.slot {
```
KolbyML: > [IGNORE] The block is not from a future slot (with a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance) -- i.e. validate that signed_beacon_block.message.slot <= current_slot (a client MAY queue future blocks for processing at the appropriate slot).


Are you sure that it is the latest slot we have or it it the actual latest slot? because you can do math to calculate what the current_slot is as there is 1 slot every 12 seconds. Because it mentions MAXIMUM_GOSSIP_CLOCK_DISPARITY, I am assuming we are calcualting the actual current slot, but you should double check

crates/networking/manager/src/validate.rs
```diff
@@ -0,0 +1,123 @@
...
+        return Ok(ValidationResult::Ignore);
+    }
+
+    let start_slot_at_epoch =
+        compute_start_slot_at_epoch(store.db.finalized_checkpoint_provider().get()?.epoch);
+
+    if block.message.slot < start_slot_at_epoch {
+        return Ok(ValidationResult::Ignore);
+    }
```
KolbyML: we might as well inline this, as the variable isn't reused

crates/common/beacon_chain/src/beacon_chain.rs
KolbyML: where is the logic which clears this cache shouldn't we use an LRU here?

crates/networking/manager/src/gossipsub.rs
```diff
@@ -78,8 +91,27 @@ pub async fn handle_gossipsub_message(message: Message, beacon_chain: &BeaconCha
...
+                                topic: GossipTopic::from_topic_hash(&message.topic)
+                                    .expect("invalid topic hash"),
+                                data: signed_block.as_ssz_bytes(),
+                            });
+                        }
+                        ValidationResult::Ignore => warn!("Ignoring gossipsub beacon block"),
+                        ValidationResult::Reject => {
+                            warn!("Rejecting gossipsub beacon block. Peer should be penalized")
+                        }
+                    }
```
KolbyML:
```suggestion
                let Ok(validation_result) = validate_beacon_block(beacon_chain, &signed_block).await else {
                    return;
                }
                match validation_result {
                    ValidationResult::Accept => {
                        if let Err(err) =
                            beacon_chain.process_block(*signed_block.clone()).await
                        {
                            error!("Failed to process gossipsub beacon block: {err}");
                        }
                        p2psender.send_gossip(GossipMessage {
                            topic: GossipTopic::from_topic_hash(&message.topic)
                                .expect("invalid topic hash"),
                            data: signed_block.as_ssz_bytes(),
                        });
                    }
                    ValidationResult::Ignore => warn!("Ignoring gossipsub beacon block"),
                    ValidationResult::Reject => {
                        warn!("Rejecting gossipsub beacon block. Peer should be penalized")
                    }
                }
```

crates/networking/manager/src/validate.rs
```diff
@@ -0,0 +1,123 @@
...
+pub enum ValidationResult {
+    Accept,
+    Ignore,
+    Reject,
+}
+
+pub async fn validate_beacon_block(
+    beacon_chain: &BeaconChain,
+    block: &SignedBeaconBlock,
+) -> anyhow::Result<ValidationResult> {
```
KolbyML: if we are defining ValidationResult, shouldn't we have Error as a part of the result, instead of having to unwrap 2 results

KolbyML (commented):
I need to read validate_beacon_block more deeply, but here is a quick review to get started before I go to bed

crates/storage/src/db.rs
```diff
@@ -195,6 +197,33 @@ impl ReamDB {
             _ => false,
         }
     }
+    pub fn get_latest_block(&self) -> anyhow::Result<SignedBeaconBlock> {
+        let highest_root = self
```
Kayden-ML: add a new line here

crates/networking/manager/src/validate.rs
varun-doshi: actually reutrning a result in `validate_beacon_block` is helpful since there are many calls to other functions that return a Result. 
if `validate_beacon_block` only returns `ValidationResult`, this will mean there are many if..let or match operations in this function.

I dont think its an issue if we are checking once for error and then once again

crates/networking/manager/src/validate.rs
varun-doshi: Hmm good question

Here's my thought process on this:
if a node has latest block =10
actual latest block = 15

the node can never now that the actual latest block number(or by extension, slot) is 15...i dont think there's a way to calculate. Of course slot=time / 12. But in the context of the node, the time is that of block=10. i.e. it does not the actual chain has progresses till 15.  Please correct me if im wrong or the explanation is not clear.

Now,
- case 1:  node receives block from gossip network for block=11.
it passes all validation and the node is updated.

- case 2: node receives block = 13
currently it will ignore it because we are not queuing it. I think this should be done in a separate issue since there could possibly require discussions on how to implement it and then use the queued block later at the correct slot. Also, wrapping this PR will help others working on other validate and forward gossip issues to make progress.

varun-doshi: Will create an issue for this

crates/networking/manager/src/service.rs
```diff
@@ -152,7 +157,7 @@ impl NetworkManagerService {
                     match event {
                         // Handles Gossipsub messages from other peers.
                         ReamNetworkEvent::GossipsubMessage { message } =>
-                            handle_gossipsub_message(message, &beacon_chain).await,
+                            handle_gossipsub_message(message, &beacon_chain,&cached_db, &p2p_sender).await,
```
KolbyML:
```suggestion
                            handle_gossipsub_message(message, &beacon_chain, &cached_db, &p2p_sender).await,
```

crates/storage/src/cache.rs
```diff
@@ -0,0 +1,32 @@
...
+        Self {
+            cached_proposer_signature: LruCache::new(unsafe { NonZeroUsize::new_unchecked(64) })
+                .into(),
+            cached_bls_to_execution_signature: LruCache::new(unsafe {
+                NonZeroUsize::new_unchecked(64)
+            })
+            .into(),
+        }
+    }
+}
```
KolbyML: please use expect and not unsafe. Could you define u64 as a constant

crates/storage/src/cache.rs
```diff
@@ -0,0 +1,32 @@
...
+use lru::LruCache;
+use ream_bls::{BLSSignature, PublicKey};
+use ream_consensus::bls_to_execution_change::BLSToExecutionChange;
+use tokio::sync::RwLock;
+
+/// In-memory LRU cache.
+#[derive(Debug)]
+pub struct CachedDB {
+    pub cached_proposer_signature: RwLock<LruCache<(PublicKey, u64), BLSSignature>>,
+    pub cached_bls_to_execution_signature: RwLock<LruCache<(PublicKey, u64), BLSToExecutionChange>>,
```
KolbyML: could you make a struct for (PublicKey, u64)

KolbyML (commented):
Leaving feedback quick, I need to find some time to more thoroughly review the validate beacon block function

crates/networking/manager/src/gossipsub/handle_gossipsub.rs
```diff
@@ -78,8 +93,27 @@ pub async fn handle_gossipsub_message(message: Message, beacon_chain: &BeaconCha
                     signed_block.message.block_root()
                 );
 
-                if let Err(err) = beacon_chain.process_block(*signed_block).await {
-                    error!("Failed to process gossipsub beacon block: {err}");
+                let Ok(validation_result) =
+                    validate_beacon_block(beacon_chain, cached_db, &signed_block).await
+                else {
+                    return;
```
unnawut: Does this simply ignore `anyhow::Error` and `Err` from `validate_beacon_block()` e.g. below?

https://github.com/ReamLabs/ream/pull/618/files#diff-3d72847d6e5e47646ecfb45566a8d1b7bf32dca639af69659d29f24b31a57c76R36-R39
```rs
    let validator = latest_state_in_db
        .validators
        .get(block.message.proposer_index as usize)
        .ok_or(anyhow!("Invalid proposer index"))?;
```

https://github.com/ReamLabs/ream/pull/618/files#diff-3d72847d6e5e47646ecfb45566a8d1b7bf32dca639af69659d29f24b31a57c76R54-R56
```rs
    if !latest_state_in_db.verify_block_signature(block)? {
        return Ok(ValidationResult::Reject);
    }
```

https://github.com/ReamLabs/ream/pull/618/files#diff-3d72847d6e5e47646ecfb45566a8d1b7bf32dca639af69659d29f24b31a57c76R59-R62
```rs
    if let Some(parent) = store
        .db
        .beacon_block_provider()
        .get(block.message.parent_root)?
```

etc. etc.

I think we should at least log that the message could not be validated into ACCEPT/IGNORE/REJECT otherwise it'll be very hard to debug dropped messages during runtime.

crates/networking/manager/src/gossipsub/handle_gossipsub.rs
```diff
@@ -78,8 +93,27 @@ pub async fn handle_gossipsub_message(message: Message, beacon_chain: &BeaconCha
...
+                            error!("Failed to process gossipsub beacon block: {err}");
+                        }
+                        p2psender.send_gossip(GossipMessage {
+                            topic: GossipTopic::from_topic_hash(&message.topic)
+                                .expect("invalid topic hash"),
+                            data: signed_block.as_ssz_bytes(),
+                        });
+                    }
+                    ValidationResult::Ignore => warn!("Ignoring gossipsub beacon block"),
+                    ValidationResult::Reject => {
```
unnawut: Also would be nice to log the ignore/reject reasons because it'll be very hard to debug in runtime.

crates/networking/manager/src/validate.rs
unnawut: > the node can never now that the actual latest block number(or by extension, slot) is 15...i dont think there's a way to calculate.

I think it's this one: https://ethereum.github.io/consensus-specs/specs/phase0/fork-choice/#get_current_slot. So:

1. I think we can base off the global slot number. Another reason why I think we should base on the global slot number is because if our node is stalling (stuck processing a block), `latest_block_in_db.message.slot` will get stuck. Then, more and more messages will not be ignored by `block.message.slot < latest_block_in_db.message.slot` and spamming the gossip network. A `block.message.slot < store.get_current_slot()` will solve this.

2. But the function above also doesn't take into account `MAXIMUM_GOSSIP_CLOCK_DISPARITY` so I agree we can take this `MAXIMUM_GOSSIP_CLOCK_DISPARITY` part in another PR so please create a new issue!

varun-doshi: looks like i missed this comment...will push a fix

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,164 @@
...
+use super::result::ValidationResult;
+
+pub async fn validate_beacon_block(
+    beacon_chain: &BeaconChain,
+    cached_db: &CachedDB,
+    block: &SignedBeaconBlock,
+) -> anyhow::Result<ValidationResult> {
+    let store = beacon_chain.store.lock().await;
+
+    let latest_global_slot = store.db.time_provider().get()?;
```
unnawut: Please recheck that this is actually global slot

crates/storage/src/db.rs
```diff
@@ -195,6 +197,34 @@ impl ReamDB {
...
+            .get_highest_root()?
+            .expect("No highest root found");
+
+        let latest_block = self
+            .beacon_block_provider()
+            .get(highest_root)?
+            .ok_or_else(|| anyhow!("Unable to fetch latest block"))?;
+
+        Ok(latest_block)
+    }
```
unnawut: Can we remove unused functions? We can always re-introduce when it's needed again

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
unnawut: Also suggesting naming close to the spec if not exact

```suggestion
    let current_global_slot = store.db.time_provider().get()?;
```

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    // [IGNORE] The block is from a slot greater than the latest finalized slot.
+    if block.message.slot
+        < compute_start_slot_at_epoch(store.db.finalized_checkpoint_provider().get()?.epoch)
+    {
+        return Ok(ValidationResult::Ignore(
+            "Block is from a slot greater than the latest finalized slot".to_string(),
+        ));
+    }
```
unnawut: I have a broad question around the validation criteria impl e.g. 

Below we validate based _directly_ on the specs definition:

```rs
    // Specs: validate that `signed_beacon_block.message.slot <= current_slot`

    if block.message.slot <= latest_global_slot {
        return Ok(ValidationResult::Ignore(
            "Block is not from a future slot".to_string(),
        ));
    }
```

Below we validate by _partial-negating_ the specs definition (partial because the code is excluding equality):
```rs
    // Specs: validate that `signed_beacon_block.message.slot > compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)`
    
    if block.message.slot
        < compute_start_slot_at_epoch(store.db.finalized_checkpoint_provider().get()?.epoch)
    {
        return Ok(ValidationResult::Ignore(
            "Block is from a slot greater than the latest finalized slot".to_string(),
        ));
    }
```

Notice how in some validations, the comparators are different from the specs and some are the same. Can you recheck for all? Sometimes we're ignoring/rejecting when validation rules are met, sometimes passing when they are met.

unnawut: Ideally unit tests would help here but I know our stance was to get a minimally working, quick & dirty beacon client impl. So I'll take this (the unit test question) separately in a team meeting.

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,164 @@
...
+    if !latest_state_in_db.get_beacon_proposer_index(Some(block.message.slot))?
+        == block.message.proposer_index
+    {
+        return Ok(ValidationResult::Reject(
+            "Proposer index is incorrect".to_string(),
+        ));
+    }
+
+    // [REJECT] The block's execution payload timestamp is correct with respect to the slot.
+    if !block.message.body.execution_payload.timestamp
```
unnawut: Is this the intended way to compare timestamps (`u64`) using the bitwise negation? For example:

```rs
println!("{:?}", !1u64 == 1u64); // false
println!("{:?}", !1u64 == 2u64); // false
```

Also applies to other similar uses with other return types as well

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
KolbyML: Tests are good, if something piece of code is complex I think tests are justified or could even be required it depends on context

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
KolbyML: time_provider gives an epoch timestamp not a slot, but also we should probably just calculate the timestamp ourselves no?

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,164 @@
...
+    cached_db: &CachedDB,
+    block: &SignedBeaconBlock,
+) -> anyhow::Result<ValidationResult> {
+    let store = beacon_chain.store.lock().await;
+
+    let latest_global_slot = store.db.time_provider().get()?;
+    let latest_state_in_db = store.db.get_latest_state()?;
+
+    // [IGNORE] The block is not from a future slot.
+    if block.message.slot <= latest_global_slot {
```
KolbyML: `latest_global_slot` isn't a slot it is a epoch timestamp

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,164 @@
...
+        return Ok(ValidationResult::Ignore(
+            "Block is not from a future slot".to_string(),
+        ));
+    }
+
+    // [IGNORE] The block is from a slot greater than the latest finalized slot.
+    if block.message.slot
+        < compute_start_slot_at_epoch(store.db.finalized_checkpoint_provider().get()?.epoch)
+    {
```
KolbyML:
```suggestion
    // [IGNORE] The block is from a slot greater than the latest finalized slot.
    if block.message.slot
        <= compute_start_slot_at_epoch(store.db.finalized_checkpoint_provider().get()?.epoch)
    {
```

Isn't a = missing ?

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,164 @@
...
+        if block.message.slot > parent.message.slot + 1 {
+            return Ok(ValidationResult::Reject(
+                "Block is from a higher slot than expected".to_string(),
+            ));
+        }
+    } else {
+        return Ok(ValidationResult::Ignore(
+            "Parent block not found".to_string(),
+        ));
+    }
```
KolbyML:
```suggestion
    match store
        .db
        .beacon_block_provider()
        .get(block.message.parent_root)? {
        Some(parent_block) => {
                // [REJECT] The block is from a higher slot than its parent.
                if block.message.slot > parent.message.slot + 1 {
                    return Ok(ValidationResult::Reject(
                        "Block is from a higher slot than expected".to_string(),
                    ));
                }
        }    
        None => {
            // [IGNORE] The block's parent (defined by block.parent_root) has been seen.
            return Ok(ValidationResult::Ignore(
                "Parent block not found".to_string(),
            ));
        }
    }
```

I think a match case would be a lot more readable here, also moving `` // [IGNORE] The block's parent (defined by block.parent_root) has been seen.`` down

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    let finalized_checkpoint = store.db.finalized_checkpoint_provider().get()?;
```
KolbyML: Aren't we missing the 

``[REJECT] The block's parent (defined by block.parent_root) passes validation.`` check? I can't see it

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,164 @@
...
+    // [REJECT] The current finalized_checkpoint is an ancestor of block.
+    if !store.get_checkpoint_block(block.message.parent_root, finalized_checkpoint.epoch)?
+        == finalized_checkpoint.root
```
KolbyML:
```suggestion
    // [REJECT] The current finalized_checkpoint is an ancestor of block.
    if store.get_checkpoint_block(block.message.parent_root, finalized_checkpoint.epoch)?
        != finalized_checkpoint.root
```

shouldn't it be like this?

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,164 @@
...
+    {
+        return Ok(ValidationResult::Reject(
+            "Finalized checkpoint is not an ancestor".to_string(),
+        ));
+    }
+
+    // [REJECT] The block is proposed by the expected proposer_index for the block's slot.
+    if !latest_state_in_db.get_beacon_proposer_index(Some(block.message.slot))?
+        == block.message.proposer_index
```
KolbyML:
```suggestion
    // [REJECT] The block is proposed by the expected proposer_index for the block's slot.
    if latest_state_in_db.get_beacon_proposer_index(Some(block.message.slot))?
        != block.message.proposer_index
```

same comment, having ! at the start looks weird without (), but also it isn't needed

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    // [REJECT] The block's execution payload timestamp is correct with respect to the slot.
+    if !block.message.body.execution_payload.timestamp
+        == latest_state_in_db.compute_timestamp_at_slot(block.message.slot)
+    {
+        return Ok(ValidationResult::Reject(
+            "Execution payload timestamp is incorrect".to_string(),
+        ));
+    }
+
```
KolbyML, line 208: image


I think we are missing the highlighted cases

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,164 @@
...
+        .contains(&AddressSlotIdentifier {
+            address: validator.public_key.clone(),
+            slot: block.message.slot,
+        })
+    {
+        return Ok(ValidationResult::Ignore(
+            "Signature already received".to_string(),
+        ));
+    }
+
```
KolbyML, line 224: [IGNORE] current_epoch >= CAPELLA_FORK_EPOCH, where current_epoch is defined by the current wall-clock time.
[IGNORE] The signed_bls_to_execution_change is the first valid signed bls to execution change received for the validator with index signed_bls_to_execution_change.message.validator_index.
[REJECT] All of the conditions within process_bls_to_execution_change pass validation.


aren't we missing the last case here?

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
varun-doshi: Ahh yes good catch

KolbyML (commented):
Here is some feedback

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
varun-doshi: Yes this was fixed in latest commit...just saw your review now will implement the suggestions next

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
unnawut: Actually I suggested using `get_current_slot()` earlier at https://github.com/ReamLabs/ream/pull/618#discussion_r2185054552 I think it's still applicable.

It derives the slot from `time_provider()` but I think it's updated per slot here: https://github.com/ReamLabs/ream/blob/master/crates/common/fork_choice/src/store.rs#L582-L583

I guess we can calculate ourselves as well, but since `get_current_slot()` is part of the specs I'm thinking it's more compatible across the codebase.

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
varun-doshi: Agreed...will cover immediately after this here https://github.com/ReamLabs/ream/issues/635

crates/networking/manager/src/validate.rs
varun-doshi: Also on rechecking this comparison was wrong;
it should be
```
if block.message.slot > current_global_slot {
        return Ok(ValidationResult::Ignore(
            "Block is from a future slot".to_string(),
        ));
    }
```

Taking the example above:
if our node has latest block =10
actual latest block = 15

Cases:
- incoming block=11: this will pass since `11>15` is not true
- incoming block=17: this will ignore since `17>15`. 
Also for case 2, this is where queuing will come in: https://github.com/ReamLabs/ream/issues/636

varun-doshi:
Latest changes does the following things:
- Makes the assertions in `process_bls_to_execution_change` into a separate function called `validate_bls_to_execution_change` since we need to use it in gossip block validation and the former function executes a state change(which should not happen in the validation pipeline)
- creates a new function `validate_parent_beacon_block` since we need to check the parent block and to avoid recursively calling the same function till genesis
- Adds other missing checks

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,275 @@
...
+    // [REJECT] The proposer signature, signed_beacon_block.signature, is valid with respect to the
+    // proposer_index pubkey.
+    if !state.verify_block_signature(block)? {
```
unnawut: From [`verify_block_signature()` code](https://github.com/ReamLabs/ream/blob/master/crates/common/consensus/src/electra/beacon_state.rs#L2409-L2420):

```rs
    pub fn verify_block_signature(&self, signed_block: &SignedBeaconBlock) -> anyhow::Result<bool> {
        let proposer = &self.validators[signed_block.message.proposer_index as usize];
        let signing_root = compute_signing_root(
            signed_block.message.clone(),
            self.get_domain(DOMAIN_BEACON_PROPOSER, None),
        );

        signed_block
            .signature
            .verify(&proposer.public_key, signing_root.as_ref())
            .map_err(|err| anyhow!("Invalid block signature: {err}"))
    }
```

By using `?`, isn't an `Err` of invalid block signature early-returned out of `validate_beacon_block()` immediately (which is behaves closer to an ignore) instead of returning `ValidationResult::Reject`?

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,275 @@
...
+                    parent_beacon_block_root: block.message.parent_root,
+                    execution_requests: block.message.body.execution_requests.clone(),
+                })
+                .await?;
+
+            match payload_verification_status {
+                // If execution_payload verification of block's parent by an execution node is not
+                // complete: [REJECT] The block's parent passes all validation (excluding
+                // execution node verification of the block.body.execution_payload)
+                PayloadStatus::Valid | PayloadStatus::Accepted | PayloadStatus::Syncing => {
```
unnawut, line 264: Should `PayloadStatus::Syncing` be considered as a completed verification?

unnawut (changes requested):

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
varun-doshi: here's my understanding from the specs:
```
If execution_payload verification of block's parent by an execution node is not complete:
[REJECT] The block's parent (defined by block.parent_root) passes all validation (excluding execution node verification of the block.body.execution_payload).
```

if it is syncing, then the parent block must only pass the local validation(we dont care about the exec node validation)
In this case, all the validation happens before we get to the execution layer validation so if it still syncing, it means we have already passed validation.

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
varun-doshi: Yup, the error case is being handled explicitly in the latest commit

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
unnawut: ok noted!

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
unnawut: Hmm but now with https://github.com/ReamLabs/ream/pull/618/commits/2a369c5676f5b2f17a4ee62b551993df5c34a3c2, wouldn't all other uses of `?` result in a reject?

E.g. `store.db.finalized_checkpoint_provider().get()?`, `get(block.message.parent_root)?`, etc. where they are internal errors, while previously it would have gone into this logic: https://github.com/ReamLabs/ream/pull/618/files#diff-5311d8512b40fed4001b0c59e6cc2f0c1369832d53984d0fe4a41eee4e34d7bdR104-R107

```rs
let validation_result = match validate_gossip_beacon_block(
    beacon_chain,
    cached_db,
    &signed_block,
)
.await
{
    Ok(result) => result,
    Err(err) => { // <-- This ---
        warn!("Failed to validate gossipsub beacon block: {err}");
        return;
    }
};
```

varun-doshi: I see your point...updating this to specifically handling the two cases where an error should mean a REJECT
The rest of them would be handled as normal internal errors.
Let me know what you think @unnawut

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,285 @@
...
+        return Ok(ValidationResult::Ignore(
+            "Signature already received".to_string(),
+        ));
+    }
+
+    // [REJECT] The proposer signature, signed_beacon_block.signature, is valid with respect to the
+    // proposer_index pubkey.
+    if !state.verify_block_signature(block).is_ok_and(|valid| valid) {
+        return Ok(ValidationResult::Reject("Invalid signature".to_string()));
+    }
```
unnawut: I think this is going in the right direction but as a general rule I think we shouldn't drop error information midway. 

For example in this case, if we dig into how `verify_block_signature()` behave, you'll find that it's returning `Result<bool, BLSError>`.

The current code handles the `Ok(true/false)` just fine, but it's dropping `BLSError` cases. If you look into [`bls::errors::BLSError`](https://github.com/ReamLabs/ream/blob/master/crates/crypto/bls/src/errors.rs) you'll see:

```rs
pub enum BLSError {
    #[cfg(feature = "supranational")]
    #[error("blst error: {0}")]
    BlstError(#[from] BlstError),
    #[error("invalid byte length")]
    InvalidByteLength,
    #[error("invalid private key")]
    InvalidPrivateKey,
    #[error("invalid public key")]
    InvalidPublicKey,
    #[error("invalid signature")]
    InvalidSignature,
    #[error("invalid hex string")]
    InvalidHexString,
}
```

We're basically dropping the error messages like "invalid byte length", "invalid public key", "invalid signature", "invalid hex string", etc. and return just `ValidationResult::Reject("Invalid signature".to_string())` for all above cases. We're losing a lot of crucial info for both UX and our own debugging.

I'd suggest handling these `BLSError` so that the returned `ValidationResult::Reject` still contains the error information.

unnawut (changes requested):

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
unnawut: I'll resolve this one but I have one follow up suggestion: https://github.com/ReamLabs/ream/pull/618#pullrequestreview-3000516187

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
varun-doshi: I did it this way to reduce verbose match statements but I think your points are valid here. Latest commit will fix this

unnawut:
lgtm now. Please recheck formatting & @KolbyML's 👍 before merging

crates/networking/manager/src/gossipsub/handle_gossipsub.rs
KolbyML, line 1: ``gossipsub/handle_gossipsub.rs``

this should just be ``gossipsub/handle.rs`` then, having gossipsub written twice is redundent

crates/networking/manager/src/gossipsub/handle_gossipsub.rs
```diff
@@ -78,8 +93,37 @@ pub async fn handle_gossipsub_message(message: Message, beacon_chain: &BeaconCha
...
+
+                match validation_result {
+                    ValidationResult::Accept => {
+                        if let Err(err) = beacon_chain.process_block(*signed_block.clone()).await {
+                            error!("Failed to process gossipsub beacon block: {err}");
+                        }
+                        p2psender.send_gossip(GossipMessage {
+                            topic: GossipTopic::from_topic_hash(&message.topic)
+                                .expect("invalid topic hash"),
+                            data: signed_block.as_ssz_bytes(),
```
KolbyML:
```suggestion
                    ValidationResult::Accept => {
                    let signed_block_bytes = signed_block.as_ssz_bytes();
                        if let Err(err) = beacon_chain.process_block(signed_block).await {
                            error!("Failed to process gossipsub beacon block: {err}");
                        }
                        p2psender.send_gossip(GossipMessage {
                            topic: GossipTopic::from_topic_hash(&message.topic)
                                .expect("invalid topic hash"),
                            data: signed_block_bytes,
```

we can avoid the clone right?

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,289 @@
...
+    match validate_beacon_block(beacon_chain, cached_db, &parent_block, &parent_state, true).await?
+    {
+        ValidationResult::Accept => {}
+        ValidationResult::Ignore(reason) => {
+            return Ok(ValidationResult::Ignore(reason));
+        }
+        ValidationResult::Reject(reason) => {
+            return Ok(ValidationResult::Reject(reason));
+        }
+    };
```
KolbyML:
```suggestion
    // Validate incoming block
    match validate_beacon_block(beacon_chain, cached_db, block, &latest_state_in_db, false).await? {
        ValidationResult::Accept => {}
        ValidationResult::Ignore(reason) => {
            return Ok(ValidationResult::Ignore(reason));
        }
        ValidationResult::Reject(reason) => {
            return Ok(ValidationResult::Reject(reason));
        }
    }
    
        let Some(parent_block) = store
        .db
        .beacon_block_provider()
        .get(block.message.parent_root)?
    else {
        bail!("failed to get parent block"));
    };

    let Some(parent_state) = store
        .db
        .beacon_state_provider()
        .get(block.message.parent_root)?
    else {
        bail!("failed to get parent state"));
    };

    
    // Validate parent block [block.message.parent_root]
    match validate_beacon_block(beacon_chain, cached_db, &parent_block, &parent_state, true).await?
    {
        ValidationResult::Accept => {}
        ValidationResult::Ignore(reason) => {
            return Ok(ValidationResult::Ignore(reason));
        }
        ValidationResult::Reject(reason) => {
            return Ok(ValidationResult::Reject(reason));
        }
    };
```

we should probably move those 2 getter calls down as it isn't needed above

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,289 @@
...
+        },
+        block.signature.clone(),
+    );
+
+    let signed_proposer_bls_execution_change = block
+        .message
+        .body
+        .bls_to_execution_changes
+        .get(block.message.proposer_index as usize)
+        .ok_or(anyhow!("Invalid index for signed bls to execution change"))?;
```
KolbyML: I think you should choose one style and be consistent either do let some or ok_or in this function, but not both

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,289 @@
...
+}
+
+pub async fn validate_beacon_block(
+    beacon_chain: &BeaconChain,
+    cached_db: &CachedDB,
+    block: &SignedBeaconBlock,
+    state: &BeaconState,
+    is_parent: bool,
+) -> anyhow::Result<ValidationResult> {
+    let store = beacon_chain.store.lock().await;
```
KolbyML, line 107: This code will not work, as it will deadlock no?

We call ``let store = beacon_chain.store.lock().await;`` in the function above. Have you tested if this code works?

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    let current_global_slot = store.get_current_slot()?;
+
+    // [IGNORE] The block is not from a future slot.
+    if block.message.slot > current_global_slot {
```
KolbyML:
```suggestion
    // [IGNORE] The block is not from a future slot.
    if block.message.slot > store.get_current_slot()? {
```

we might as well inline this, as we don't reuse the variable

KolbyML (commented):
The PR is looking better, but a I have a few concerns that should be resolved

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
varun-doshi: You're right about the deadlock here...will be fixed in latest

As discussed with O, testing this will be top priority after this PR. Issue here: https://github.com/ReamLabs/ream/issues/635

KolbyML: You shouldn't be submitting broken code, you need to make sure it works. If that expectation can't be met maybe tests should be done in this PR. 

Even if tests aren't added in a PR you should test if the code works.

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,300 @@
...
+pub async fn validate_gossip_beacon_block(
+    beacon_chain: &BeaconChain,
+    cached_db: &CachedDB,
+    block: &SignedBeaconBlock,
+) -> anyhow::Result<ValidationResult> {
+    let latest_state_in_db = {
+        let store = beacon_chain.store.lock().await;
+
+        store.db.get_latest_state()?
+    };
```
KolbyML:
```suggestion
    let latest_state = beacon_chain.store.lock().await.db.get_latest_state()?;
```

this can be inlined

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
KolbyML: Of course edge cases can be missed and that is fine, it is just the deadlock problem could be avoided fairly easily, also deadlocks could cost the team a week or two, depending on when we find them.

KolbyML:
> Tests added for block validation. To run:
> 
> ```
> cargo test --package beacon-api-tests --test validate_block -- tests::test_validate_beacon_block --exact --show-output
> ```
> 
> There's 1 catch here regarding the test though. Specifically the following validation rule: `[REJECT] The current finalized_checkpoint is an ancestor of block -- i.e. get_checkpoint_block(store, block.parent_root, store.finalized_checkpoint.epoch) == store.finalized_checkpoint.root`

we shouldn't be uncommenting validation cases, maybe mark this section with a cfg to not compile it in test code

> The issue is this goes back recursively backwords to the last finalized checkpoint so which means we need atleast the last 64 blocks (which means we need to save 64 blocks as json). Now I've tested it going backwards 3 blocks and it words as expected. This is why you will see this specific validation commented out. Even on general case, this validation should not always run. I think it should only run when we know there is sufficient backfill. Which would mean there needs to be a flag indicating backfill status and execute this validation only then.
> 
> Thoughts @unnawut @KolbyML

I need to read more into the case before I can fully comment on this. I left some early feedback above

KolbyML:
@varun-doshi can you rebase?

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
KolbyML: This feedback wasn't addressed

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,291 @@
...
+            "Block is from a slot greater than the latest finalized slot".to_string(),
+        ));
+    }
+
+    let validator =
+        if let Some(validator) = state.validators.get(block.message.proposer_index as usize) {
+            validator
+        } else {
+            return Ok(ValidationResult::Reject("Validator not found".to_string()));
+        };
```
KolbyML:
```suggestion
    let Some(validator) = state.validators.get(block.message.proposer_index as usize) else {
            return Ok(ValidationResult::Reject("Validator not found".to_string()));
        };
```

^ we can simplify the code to this

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,291 @@
...
+    };
+
+    let validator = if let Some(validator) = latest_state_in_db
+        .validators
+        .get(block.message.proposer_index as usize)
+    {
+        validator
+    } else {
+        return Ok(ValidationResult::Reject("Validator not found".to_string()));
+    };
```
KolbyML: here as well

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,291 @@
...
+        .db
+        .beacon_block_provider()
+        .get(block.message.parent_root)?
+    {
+        Some(parent_block) => {
+            // [REJECT] The block is from a higher slot than its parent.
+            if block.message.slot > parent_block.message.slot + 1 {
+                return Ok(ValidationResult::Reject(
+                    "Block is from a higher slot than expected".to_string(),
+                ));
```
KolbyML:
```suggestion
    match store
        .db
        .beacon_block_provider()
        .get(block.message.parent_root)?
    {
        Some(parent_block) => {
            // [REJECT] The block is from a higher slot than its parent.
            if block.message.slot <= parent_block.message.slot {
                return Ok(ValidationResult::Reject(
                    "Block is from a higher slot than expected".to_string(),
                ));
```

^ shouldn't the condition be this?

crates/networking/manager/src/gossipsub/validate/beacon_block.rs
```diff
@@ -0,0 +1,291 @@
...
+        let finalized_checkpoint = store.db.finalized_checkpoint_provider().get()?;
+        // [REJECT] The current finalized_checkpoint is an ancestor of block.
+        if store.get_checkpoint_block(block.message.parent_root, finalized_checkpoint.epoch)?
+            != finalized_checkpoint.root
+        {
+            return Ok(ValidationResult::Reject(
+                "Finalized checkpoint is not an ancestor".to_string(),
+            ));
+        }
+    }
```
KolbyML: can you negate this so it is on by default, then disable it just for the test

testing/beacon-api/tests/validate_block.rs
KolbyML, line 1: the PR is pretty much ready to merge, could you move these tests to the crate they are for, as this test isn't related to the beacon-api

KolbyML: or make a new testing crate

KolbyML (approved):
:shipit: looks good. I will merge the PR now as I want it in :triumph:

Issue #221, opened 2025-03-25
Implement /eth/v1/debug/fork_choice


KolbyML:
https://ethereum.github.io/beacon-APIs/#/Debug/getDebugForkChoice

BornPsych:
Hi @KolbyML ! I'm interested in working on this. can you please assign it to me

syjn99:
It seems this issue is stale, so I can take this up
cc. @varun-doshi

PR #624, opened 2025-06-28
feat: implement `/eth/v1/debug/fork_choice` endpoint

syjn99:

Ref #221 

- Refactored handlers for debug endpoints to be placed in the same file.
- `fork_choice/store.rs` has a notable change: As we should track the map between block(node) and its justified/finalized epoch, function signatures of `filter_block_tree` and `get_filter_block_tree` have changed to maintain a tuple as a value.

### To-Do

 
- [xI have read [CONTRIBUTING.md](https://github.com/ReamLabs/ream/blob/master/CONTRIBUTING.md).

### DIFF
--- a/crates/common/fork_choice/src/store.rs
+++ b/crates/common/fork_choice/src/store.rs
@@ -88,7 +88,7 @@ impl Store {
     pub fn filter_block_tree(
         &self,
         block_root: B256,
-        blocks: &mut HashMap<B256, BeaconBlock>,
+        blocks: &mut HashMap<B256, (BeaconBlock, u64, u64)>,
     ) -> anyhow::Result<bool> {
         let Some(block) = self.db.beacon_block_provider().get(block_root)? else {
             bail!("failed to get block");
@@ -109,7 +109,18 @@ impl Store {
                 .collect::<anyhow::Result<Vec<_>>>()?;
 
             if filter_results.iter().any(|&result| result) {
-                blocks.insert(block_root, block.message.clone());
+                let voting_source = self.get_voting_source(block_root)?;
+                let finalized_epoch = self.db.finalized_checkpoint_provider().get()?.epoch;
+
+                blocks.insert(
+                    block_root,
+                    (
+                        block.message.clone(),
+                        // NOTE: Use the node's own `voting_source.epoch` as its `justified_epoch`.
+                        voting_source.epoch,
+                        finalized_epoch,
+                    ),
+                );
                 return Ok(true);
             }
             return Ok(false);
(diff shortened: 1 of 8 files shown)


syjn99:
Not yet tested, WiFi in this train is super slow :) Will test after check-in the hotel

crates/rpc/src/handlers/debug.rs
```diff
@@ -0,0 +1,120 @@
...
+use crate::handlers::state::get_state_from_id;
+
+#[get("/debug/beacon/states/{state_id}")]
+pub async fn get_debug_beacon_state(
+    db: Data<ReamDB>,
+    state_id: Path<ID>,
+) -> Result<impl Responder, ApiError> {
+    let state = get_state_from_id(state_id.into_inner(), &db).await?;
+
+    Ok(HttpResponse::Ok().json(BeaconResponse::new(state)))
```
KolbyML:
```suggestion
#[get("/debug/beacon/states/{state_id}")]
pub async fn get_debug_beacon_state(
    db: Data<ReamDB>,
    state_id: Path<ID>,
) -> Result<impl Responder, ApiError> {
    Ok(HttpResponse::Ok().json(BeaconResponse::new(get_state_from_id(state_id.into_inner(), &db).await?)))
```

we might as well inline this

crates/common/fork_choice/src/store.rs
```diff
@@ -109,7 +109,18 @@ impl Store {
...
-                blocks.insert(block_root, block.message.clone());
+                let voting_source = self.get_voting_source(block_root)?;
+                let finalized_epoch = self.db.finalized_checkpoint_provider().get()?.epoch;
+
+                blocks.insert(
+                    block_root,
+                    (
+                        block.message.clone(),
+                        // NOTE: Use the node's own `voting_source.epoch` as its `justified_epoch`.
+                        voting_source.epoch,
```
KolbyML: Can you add a justification/reasoning in the comment, I can read what is happening etc etc. same below

crates/common/fork_choice/src/store.rs
```diff
@@ -145,7 +164,12 @@ impl Store {
 
     /// Retrieve a filtered block tree from ``store``, only returning branches
     /// whose leaf state's justified/finalized info agrees with that in ``store``.
-    pub fn get_filtered_block_tree(&self) -> anyhow::Result<HashMap<B256, BeaconBlock>> {
+    ///
+    /// NOTE: ``blocks`` must contain justified/finalized epoch information of its node, so tuple of
+    /// `(BeaconBlock, justified_epoch, finalized_epoch)` should be the value of the map.
+    pub fn get_filtered_block_tree(
+        &self,
+    ) -> anyhow::Result<HashMap<B256, (BeaconBlock, u64, u64)>> {
```
KolbyML: could you just make a struct for (BeaconBlock, u64, u64)? as it is a little non-intuitive

KolbyML (approved):
:shipit: PR looks good, if you just apply the feedback feel free to merge

KolbyML:
In the PR's description it says ``  "extra_data": null``, but the BeaconAPI specification says ``"extra_data": {}`` is this field use in other nodes?, I am just asking cause it doesn't seem like we are matching the spec rn

syjn99:
> I am just asking cause it doesn't seem like we are matching the spec rn

@KolbyML Yeah you're correct, I will fix this

crates/common/fork_choice/src/store.rs
syjn99: Sure!

KolbyML (approved):
:shipit: looks good

Issue #617, opened 2025-06-24
Add voluntary exit command to Validator CLI


shariqnaiyer:
### Describe the feature

Once #610 is merged, we can implement the flow to voluntary exit a validator in practice. It takes a minimum of 5 epochs after initiating a voluntary exit for a validator to exit. We must monitor the validator status for this time period and continue performing the Validator's duties as failing to do so will incur penalties.

PR #625, opened 2025-07-03
feat: add voluntary exit command to cli

shariqnaiyer:

Fixes: #617 

I first created the function to process the voluntary exit. This function currently takes in a validator index for a validator for which to exit, however, I am finding it to be common practice to exit all validators in the provided keystore. I am going to do more research into this and decide if we want the ability to exit individual validators. 

I then created the CLI components for this and updated the docs.

This is what the voluntary exit command will look like:
```cmd
./ream voluntary_exit \
    --beacon-api-endpoint http://localhost:5052 \
    --import-keystores ./keystores \
    --password-file ./password.txt \
    --validator-index 12345 \
    --epoch 150000 \
    --wait
```

I find (existing implementations)(https://www.coincashew.com/coins/overview-eth/guide-or-how-to-setup-a-validator-on-eth2-mainnet/part-iii-tips/voluntary-exiting-a-validator) to be follow a similar structure for exits.

### To-Do

 
### DIFF
--- a/bin/ream/src/cli/voluntary_exit.rs
+++ b/bin/ream/src/cli/voluntary_exit.rs
@@ -1,0 +1,57 @@
+use std::{path::PathBuf, sync::Arc, time::Duration};
+
+use clap::Parser;
+use ream_network_spec::{cli::network_parser, networks::NetworkSpec};
+use url::Url;
+
+use crate::cli::{
+    constants::{DEFAULT_BEACON_API_ENDPOINT, DEFAULT_NETWORK, DEFAULT_REQUEST_TIMEOUT},
+    validator_node::duration_parser,
+};
+
+#[derive(Debug, Parser)]
+pub struct VoluntaryExitConfig {
+    /// Verbosity level
+    #[arg(short, long, default_value_t = 3)]
+    pub verbosity: u8,
+
+    #[arg(long, help = "Set HTTP url of the beacon api endpoint", default_value = DEFAULT_BEACON_API_ENDPOINT)]
+    pub beacon_api_endpoint: Url,
+
+    #[arg(long, help = "Set HTTP request timeout for beacon api calls", default_value = DEFAULT_REQUEST_TIMEOUT, value_parser = duration_parser)]
+    pub request_timeout: Duration,
+
+    #[arg(
+        long,
+        help = "Choose mainnet, holesky, sepolia, hoodi, dev or provide a path to a YAML config file",
+        default_value = DEFAULT_NETWORK,
+        value_parser = network_parser
+    )]
+    pub network: Arc<NetworkSpec>,
+
+    #[arg(long, help = "The directory for importing keystores")]
+    pub import_keystores: PathBuf,
+
+    #[arg(
+        long,
+        group = "password_source",
+        help = "The plaintext password file to use for keystores"
+    )]
+    pub password_file: Option<PathBuf>,
+
+    #[arg(
+        long,
+        group = "password_source",
+        help = "The password to use for keystores. It's recommended to use password-file over this in order to prevent your keystore password from appearing in the shell history"
+    )]
+    pub password: Option<String>,
+
+    #[arg(long, help = "The validator index to exit")]
+    pub validator_index: u64,
+
+    #[arg(long, help = "The epoch for the voluntary exit")]
+    pub epoch: u64,
+
+    #[arg(long, help = "Wait until the validator has fully exited")]
+    pub wait: bool,
+}
--- a/bin/ream/src/main.rs
+++ b/bin/ream/src/main.rs
@@ -7,6 +7,7 @@ use ream::cli::{
     beacon_node::BeaconNodeConfig,
     import_keystores::{load_keystore_directory, load_password_file, process_password},
     validator_node::ValidatorNodeConfig,
+    voluntary_exit::VoluntaryExitConfig,
 };
 use ream_checkpoint_sync::initialize_db_from_checkpoint;
 use ream_consensus::constants::set_genesis_validator_root;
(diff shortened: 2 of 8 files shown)


bin/ream/src/main.rs
```diff
@@ -228,3 +232,54 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
...
+    match ream_validator::voluntary_exit::process_voluntary_exit(
+        validator_service,
+        config.validator_index,
+        config.epoch,
+        config.wait,
+    )
+    .await
+    {
+        Ok(()) => info!("Voluntary exit completed successfully"),
+        Err(err) => error!("Voluntary exit failed: {}", err),
```
KolbyML: inline the err

Patchoulis: Eh, ngl, if we're gonna be reusing the same code for password setting as what we have in the validator, shouldn't we perhaps make a new function or method for this? Perhaps we could define a trait for it and make it so that both the validator config and voluntary exit code uses it..

crates/common/validator/src/voluntary_exit.rs
```diff
@@ -34,3 +43,54 @@ pub fn sign_voluntary_exit(
...
+        return Err(anyhow!(
+            "Cannot process voluntary exit while node is syncing"
+        ));
+    }
+
+    validator_service
+        .submit_voluntary_exit(validator_index, epoch)
+        .await?;
+
+    if wait_till_exit {
```
Patchoulis: My worry is that if you were to say, make a few voluntary exit requests, wouldn't it use up a thread for each? I'd imagine the better method would be to do it sort of like how key manager services are done, where we could call an api for this, and it would run on the same thread. 

I'd imagine if we were to try and exit a bunch of validators at once, it might be problematic?

Patchoulis: Plus, we also do fetch the head beacon block info though in the validator service. Aren't we sending unnecessary requests though like this?

Patchoulis (changes requested):

crates/common/validator/beacon/src/voluntary_exit.rs
```diff
@@ -34,3 +43,52 @@ pub fn sign_voluntary_exit(
...
+    if wait_till_exit {
+        loop {
+            match beacon_api_client
+                .get_state_validator(ID::Head, ValidatorID::Index(validator_index))
+                .await?
+                .data
+                .status
+            {
+                ValidatorStatus::ActiveExiting => {
+                    println!(
```
KolbyML: don't use print in production code use tracing

KolbyML:
@Patchoulis we need your review to continue

bin/ream/src/main.rs
```diff
@@ -258,3 +266,60 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
...
+
+    let password = process_password({
+        if let Some(ref password_file) = config.password_file {
+            load_password_file(password_file).expect("Failed to read password from password file")
+        } else if let Some(password_str) = config.password {
+            password_str
+        } else {
+            panic!("Expected either password or password-file to be set")
+        }
+    });
```
Patchoulis: It's used in a multiple places. Would be nice to refactor it.

bin/ream/src/cli/voluntary_exit.rs
```diff
@@ -0,0 +1,57 @@
...
+        group = "password_source",
+        help = "The password to use for keystores. It's recommended to use password-file over this in order to prevent your keystore password from appearing in the shell history"
+    )]
+    pub password: Option<String>,
+
+    #[arg(long, help = "The validator index to exit")]
+    pub validator_index: u64,
+
+    #[arg(long, help = "The epoch for the voluntary exit")]
+    pub epoch: u64,
```
Patchoulis: I don't believe we're supposed to be getting the epoch from the CLI. I guess you could modify it for minimum epoch though.

crates/common/validator/beacon/src/voluntary_exit.rs
```diff
@@ -34,3 +43,52 @@ pub fn sign_voluntary_exit(
...
+    private_key: &PrivateKey,
+    wait_till_exit: bool,
+) -> anyhow::Result<()> {
+    let sync_status = beacon_api_client.get_node_syncing_status().await?;
+
+    if sync_status.data.is_syncing {
+        return Err(anyhow!(
+            "Cannot process voluntary exit while node is syncing"
+        ));
+    }
```
Patchoulis: If you're planning on doing minimum epoch. You should probably add the necessary validations for it. For example, checking that it is indeed a future epoch. Another would be to make sure that the validator is present. At the minimum, it'd be nice to at least check if there was an error message of some sort for when you submit the voluntary exit.

Patchoulis (changes requested):

crates/common/validator/beacon/src/voluntary_exit.rs
```diff
@@ -34,3 +43,52 @@ pub fn sign_voluntary_exit(
...
+                    );
+                    sleep(Duration::from_secs(network_spec().seconds_per_slot)).await;
+                }
+                ValidatorStatus::ExitedSlashed | ValidatorStatus::ExitedUnslashed => {
+                    println!("Validator has successfully exited");
+                    break;
+                }
+                _ => {
+                    println!("Voluntary exit has not yet been published to beacon chain.");
+                    sleep(Duration::from_secs(network_spec().seconds_per_slot)).await;
```
Patchoulis: This line is duplicated. I'd say just move it before the match statement.

Patchoulis (changes requested):

crates/common/validator/beacon/src/voluntary_exit.rs
syjn99: Nice note, indeed we need a basic validation for receiving epoch data from CLI! But I think specifying minimum epoch is an advanced feature, so in this phase it's enough to implement very basic feature set.

syjn99 (approved):
Reviewed all the diff and looks good to me.

Patchoulis (approved):

KolbyML (approved):
:shipit: looks good

PR #628, opened 2025-07-03
feat: Implement /eth/v1/node/identity

### LINKED ISSUE #177: Implement /eth/v1/node/identity

### LINKED ISSUE #627: Expose local node info to the NetworkManagerService
### Describe the feature

The local node info is not available to any service in the node. This refers to the peer id, enr and other details. Some of it is available in the `Network` structure but we lose handle on it when we call `init` in the `main.rs` file.

### Additional context

_No response_

varun-doshi:

Closes #627 
Update: Also closes #177 

To run endpoint `node/identity`:
```
http://127.0.0.1:5052/eth/v1/node/identity
```

- Add additional fields to the `NetworkState`: peer id, enr, socket_address, socket_port, discovery_port
- implements `node/identity` endpoint: https://ethereum.github.io/beacon-APIs/#/Node/getNetworkIdentity


KolbyML (changes requested):
What is the usecases of this? we already know the config information, and we have the network_state for a accessing this data, is there a technical feature or reasoning for this PR?

Currently the code being added isn't used for anything, and it's usecase isn't clear to me

varun-doshi:
> What is the usecases of this? we already know the config information, and we have the network_state for a accessing this data, is there a technical feature or reasoning for this PR?

Does network_state really have this data? metadata is but the other stuff isnt

```
pub struct NetworkState {
    pub peer_table: RwLock<HashMap<PeerId, CachedPeer>>,
    pub meta_data: RwLock<GetMetaDataV2>,
    pub status: RwLock<Status>,
    pub data_dir: PathBuf,
}
```

crates/networking/manager/src/service.rs
```diff
@@ -106,14 +108,26 @@ impl NetworkManagerService {
...
+        let meta_data = network.network_state().meta_data.read().clone();
+
+        let local_node = Arc::new(NodeInfo::from(
+            network.peer_id(),
+            network.enr(),
+            config.discovery_port,
+            config.socket_address,
+            config.socket_port,
+            meta_data,
+        ));
```
unnawut: If it's only for #177, can we build the result from `Network`, `ManagerConfig` and `NetworkState` at the endpoint handler level than building a long lasting `NodeInfo` for the use case?

Otherwise `NetworkManagerService` is not maintaining a [separation of concerns](https://www.geeksforgeeks.org/software-engineering/separation-of-concerns-soc/) by specifically preparing a data structure for a specific endpoint.

My opinion might be different if there are other use cases than #177 however...

crates/networking/manager/src/service.rs
```diff
@@ -106,14 +108,26 @@ impl NetworkManagerService {
         let status = beacon_chain.build_status_request().await?;
 
         let network = Network::init(executor.clone(), &network_config, status).await?;
+        let meta_data = network.network_state().meta_data.read().clone();
```
unnawut: Also here maybe just my lack of Rust knowledge, but I'm not sure if cloning the `RwLock` directly actually maintains the original reference.

Because if it's not maintaining the reference, we're risking `NodeInfo` data being stale from ip address changes and metadata (attnets, syncnets) changes during runtime.

The [documentation for `Arc`](https://doc.rust-lang.org/std/sync/struct.Arc.html#cloning-references) does specifically mention that `clone()` points to the same reference, but [not mentioned for `RwLock`](https://doc.rust-lang.org/std/sync/struct.RwLock.html).

Would be nice if this is confirmed by a unit test if not by documentation.

KolbyML: By default rust objects are non-mutable, RwLock makes the Arc mutable, although yes what you mentioned above about the data becoming stale is true, and should be dealt with in this PR, I didn't bring it up yet because my primary concern is the ENR/IP's should just be put into the NetworkState, we don't need another Arc for NodeInfo as NetworkState can already serve the needed role and we can add fields to it if needed. Adding another is redundant

KolbyML:
> > What is the usecases of this? we already know the config information, and we have the network_state for a accessing this data, is there a technical feature or reasoning for this PR?
> 
> Does network_state really have this data? metadata is but the other stuff isnt
> 
> ```
> pub struct NetworkState {
>     pub peer_table: RwLock<HashMap<PeerId, CachedPeer>>,
>     pub meta_data: RwLock<GetMetaDataV2>,
>     pub status: RwLock<Status>,
>     pub data_dir: PathBuf,
> }
> ```

NetworkState is already exposed to the BeaconAPI ``            .app_data(Data::new(network_state.clone()))``. We can just add the additional fields here if required, as I mentioned here https://github.com/ReamLabs/ream/pull/628#discussion_r2185142856

crates/networking/p2p/src/network.rs
```diff
@@ -173,6 +173,7 @@ impl Network {
             identify::Behaviour::new(identify_config)
         };
 
+        let local_enr = discovery.local_enr().clone();
```
unnawut: If it's only used by `NetworkState` instantiation on line 210, maybe inline it with the instantiation?

unnawut: Also by cloning, `NetworkState`'s knowledge of the ENR will stale when the ENR is updated. Can we reference the original value?

crates/networking/p2p/src/network.rs
```diff
@@ -204,7 +205,13 @@ impl Network {
                 .build()
         };
 
+        let local_peer_id = PeerId::from_public_key(&PublicKey::from(local_key.public().clone()));
```
unnawut: On the other hand cloning here is probably fine because the public key should not change, unlike ENR.

(No changes suggested here, just want to point out the difference)

crates/networking/p2p/src/network.rs
```diff
@@ -845,8 +852,6 @@ mod tests {
         .build();
 
         let config = NetworkConfig {
-            socket_address,
-            socket_port,
```
unnawut, line 849: Good that we're deduplicating these fields 👍

crates/networking/p2p/src/req_resp/messages/meta_data.rs
```diff
@@ -1,9 +1,10 @@
+use serde::Serialize;
 use ssz_derive::{Decode, Encode};
 use ssz_types::BitVector;
 
 use crate::req_resp::configurations::{AttestationSubnetCount, SyncCommitteeSubnetCount};
 
-#[derive(Debug, Default, Clone, PartialEq, Eq, Encode, Decode)]
+#[derive(Debug, Default, Clone, PartialEq, Eq, Encode, Decode, Serialize)]
```
unnawut: Is adding `Serialize` needed for the scope of this PR?

If it's for endpoint response, I'd suggest including as part of the subsequent PR.

crates/networking/p2p/src/network.rs
varun-doshi: When can we expect the ENR to change? Correct me if Im wrong, but i dont see it changing anywhere whilst a node is running. If so, the only way an ENR can change is when a node starts?

I'm asking so that i understand if we need shared mutable ref or immutable shared state for ENR

KolbyML: The ENR  can change various ways during runtime

- The Discv5 service will update the ENR with the latest IP address, and send us updates of that to our event stream, so when this happens we will want to update the version of the ENR in the NetworkState struct to the new one.
- When we update attnets and the other fields

So there is 2 different places the ENR could be updated during runtime

```
pub struct Discovery {
    discv5: Discv5,
    local_enr: Enr,
    event_stream: EventStream,
```

If the ENR is in the Network state then, it shouldn't be in the  Discovery as well, ideally we are only tracking this in one location to reduce complexityy.

So your reference would need to be mutable

KolbyML: https://docs.rs/discv5/latest/discv5/enum.Event.html#variant.SocketUpdated

Here is documentation for the event that notifies if Discv5 changed the ENR on us

varun-doshi: Thank you for clarifying...yes I agree ideally we only want this in 1 struct. Looking into this

crates/rpc/src/handlers/identity.rs
```diff
@@ -0,0 +1,51 @@
...
+        discovery_port: u16,
+        socket_addr: IpAddr,
+        socket_port: u16,
+        metadata: GetMetaDataV2,
+    ) -> Self {
+        Self {
+            peer_id: peer_id.to_string(),
+            enr: enr.to_base64(),
+            p2p_address: vec![format!("/ip4/{socket_addr}/tcp/{socket_port}/{peer_id}")],
+            discovery_address: vec![format!("/ip4/7.7.7.7/udp/{discovery_port}/p2p/{peer_id}")],
```
KolbyML: we shouldn't be hardcoding these values they should be derived on the fly, the node might have ip4  enabled and only ipv6 enabled,

also the IP/PORT would be derivable from the ENR, so it shouldn't be included in the NetworkSpec

crates/networking/discv5/src/discovery.rs
```diff
@@ -115,18 +114,16 @@ impl Discovery {
...
+                event_stream,
+                discovery_queries: FuturesUnordered::new(),
+                find_peer_active: false,
+                started: !config.disable_discovery,
+            },
+            enr,
+        ))
     }
 
     pub fn discover_peers(&mut self, query: QueryType, target_peers: usize) {
```
KolbyML, line 131: In `   fn poll(` you should check if the UpdateAddress event is sent and if it is, you should update the ENR in NetworkSpec, to the Enr stored in Discv5

KolbyML:
Also please rebase the PR

crates/rpc/src/handlers/identity.rs
```diff
@@ -0,0 +1,65 @@
...
+                    "/ip6/{}/udp/{}/p2p/{}",
+                    enr.tcp6().expect("tcp6 not set"),
+                    enr.udp6().expect("udp6 not set"),
+                    peer_id
+                ),
+            }],
+            metadata,
+        }
+    }
+}
```
unnawut: Looks good to me that we're now formatting the response data at the RPC level 👍

crates/rpc/src/handlers/identity.rs
```diff
@@ -0,0 +1,65 @@
...
+/// Called by `eth/v1/node/identity` to get the Node Identity.
+#[get("/node/identity")]
+pub async fn get_identity(
+    network_state: Data<Arc<NetworkState>>,
+) -> Result<impl Responder, ApiError> {
+    Ok(HttpResponse::Ok().json(DataResponse::new(Identity::new(
+        network_state.local_peer_id,
+        network_state.local_enr.read().clone(),
+        network_state.meta_data.read().clone(),
+    ))))
```
unnawut: Very clean I like this

unnawut (approved):
lgtm but please wait for @KolbyML's 👍 before merging since he also did extensive review on this PR.

Also suggest renaming the PR to "Implement /eth/v1/node/identity"

crates/networking/discv5/src/discovery.rs
```diff
@@ -329,7 +334,22 @@ impl NetworkBehaviour for Discovery {
                     }
                 }
             }
-            EventStream::Present(_receiver) => {}
+            EventStream::Present(receiver) => match receiver.try_recv() {
+                Ok(event) => {
+                    if let Event::SocketUpdated(socket_addr) = event {
+                        self.discv5.update_local_enr_socket(socket_addr, true);
```
KolbyML:
```suggestion
                    if let Event::SocketUpdated(socket_addr) = event {
```
I don't think we woul.d need to call this, as Discv5 already updated the ENR on their end

crates/networking/discv5/src/discovery.rs
```diff
@@ -329,7 +334,22 @@ impl NetworkBehaviour for Discovery {
...
+                        if let Some(enr_sender) = &self.enr_update_sender {
+                            if let Err(e) = enr_sender.send(self.discv5.local_enr()) {
+                                warn!("Failed to send socket address update: {:?}", e);
+                            }
+                        }
+                    }
+                }
+                Err(error) => {
+                    error!("Failed to start discovery event stream: {:?}", error);
+                    self.event_stream = EventStream::Inactive;
```
KolbyML: We should stay consistent with our error variable names, could you use `err` instead of `error` and `e` and inline them {err:?}

crates/networking/discv5/src/discovery.rs
```diff
@@ -61,15 +61,19 @@ struct QueryResult {
...
     pub started: bool,
+    pub enr_update_sender: Option<mpsc::UnboundedSender<Enr>>,
 }
 
 impl Discovery {
-    pub async fn new(local_key: Keypair, config: &DiscoveryConfig) -> anyhow::Result<Self> {
+    pub async fn new(
+        local_key: Keypair,
+        config: &DiscoveryConfig,
+        enr_update_sender: Option<mpsc::UnboundedSender<Enr>>,
```
KolbyML: could we update/refactor the code so we can just update NetworkState directly instead of using a channel

crates/networking/p2p/src/network_state.rs
```diff
@@ -13,6 +13,8 @@ use crate::{
 };
 
 pub struct NetworkState {
+    pub local_peer_id: PeerId,
+    pub local_enr: RwLock<Enr>,
```
KolbyML: The PeerId can also be derived from the ENR so we don't need to store it here

crates/rpc/src/handlers/identity.rs
```diff
@@ -0,0 +1,65 @@
...
+                    enr.udp4().expect("udp4 not set"),
+                    peer_id
+                ),
+                None => format!(
+                    "/ip6/{}/udp/{}/p2p/{}",
+                    enr.tcp6().expect("tcp6 not set"),
+                    enr.udp6().expect("udp6 not set"),
+                    peer_id
+                ),
+            }],
```
KolbyML: Nodes can have both ipv4 and ipv6 addressees at the same time

crates/networking/discv5/src/discovery.rs
varun-doshi: We cant directly update NetworkState from `fn poll` because it is a trait function so we cant call external containers. the only other way is to store network_state inside Discovery struct as well which does not make much sense to me and just increasing complexity since we have to maintain shared mutable state.
Also for my own understanding, what is the issue with using channels here? I thought it was a perfect usecase since it is an event triggered change.

Please let me know if there is any another way you can think of

KolbyML: > Please let me know if there is any another way you can think of

``store network_state inside Discovery struct`` this is what I was saying to do yeah. I am not looking for another way, I am directly looking for this. Instead of ``local_enr: Enr,`` replace it with NetworkState. Also just remove the Discv5 crate and make it a module in the p2p crate

> increasing complexity since we have to maintain shared mutable state.

I don't think this increases complexity I think it decreases it

> Also for my own understanding, what is the issue with using channels here? I thought it was a perfect usecase since it is an event triggered change.

You are currently using a tangling tokio task, I don't think there is a need for it

KolbyML: Oh there is 1 other way to accomplish this, you can just pass up the event from the poll back to the Swarm and handle it in the network.rs code. This wouldn't require the channel, and would also resolve your concerns

KolbyML: You can do

return Poll::Ready(ToSwarm::GenerateEvent(UpdatedEnr { enr: self.discv5.local_enr() }));

^ I think we should do this method, instead of what I initially suggested, and the channel solution

varun-doshi: This method sounds good

crates/rpc/src/handlers/identity.rs
```diff
@@ -0,0 +1,72 @@
...
+                let mut addresses = Vec::new();
+
+                if let Some(ip4) = enr.ip4() {
+                    if let Some(udp4) = enr.udp4() {
+                        addresses.push(format!("/ip4/{}/udp/{}/p2p/{}", ip4, udp4, peer_id));
+                    }
+                }
+                if let Some(ip6) = enr.ip6() {
+                    if let Some(udp6) = enr.udp6() {
+                        addresses.push(format!("/ip6/{}/udp/{}/p2p/{}", ip6, udp6, peer_id));
```
KolbyML, line 52: the /p2p/ thing should only be applied once right?

crates/networking/discv5/src/discovery.rs
```diff
@@ -39,6 +39,17 @@ pub struct DiscoveredPeers {
...
+#[derive(Debug)]
+pub struct UpdatedEnr {
+    pub enr: Enr,
+}
+
+#[derive(Debug)]
+pub enum DiscoveryOutEvent {
+    DiscoveredPeers(DiscoveredPeers),
+    UpdatedEnr(UpdatedEnr),
+}
```
KolbyML:
```suggestion
#[derive(Debug)]
pub enum DiscoveryOutEvent {
    DiscoveredPeers { peers: HashMap<Enr, Option<Instant>> },
    UpdatedEnr { enr: Enr },
}
```

What do you think of inlining these? Having structs for these fields, just seems a little overkill right now

crates/networking/discv5/src/discovery.rs
```diff
@@ -61,15 +72,14 @@ struct QueryResult {
...
-    local_enr: Enr,
     event_stream: EventStream,
     discovery_queries: FuturesUnordered<Pin<Box<dyn Future<Output = QueryResult> + Send>>>,
     find_peer_active: bool,
     pub started: bool,
 }
 
 impl Discovery {
-    pub async fn new(local_key: Keypair, config: &DiscoveryConfig) -> anyhow::Result<Self> {
+    pub async fn new(local_key: Keypair, config: &DiscoveryConfig) -> anyhow::Result<(Self, Enr)> {
```
KolbyML:
```suggestion
    pub async fn new(local_key: Keypair, config: &DiscoveryConfig) -> anyhow::Result<Self> {
```

instead of returning the ENR from new we should just
let discovery = Discovery::new(x,y,z);
let enr = discovery.local_enr();

local_enr() would fetch it from discv5

crates/networking/p2p/src/network.rs
```diff
@@ -459,10 +462,18 @@ impl Network {
...
+                )) => {
                     self.handle_discovered_peers(peers);
                     None
                 }
+                ReamBehaviourEvent::Discovery(DiscoveryOutEvent::UpdatedEnr(UpdatedEnr {
+                    enr,
+                })) => {
+                    *self.network_state.local_enr.write() = enr;
+                    None
+                }
```
KolbyML: Can you just do a match case instead of duplicating ``ReamBehaviourEvent::Discovery`` twice

crates/networking/discv5/src/discovery.rs
varun-doshi: yup good idea

PR #629, opened 2025-07-03
perf: save 3 seconds by reusing the calculated state root in the checkpoint sync crate

syjn99:

 

While a node is syncing from the checkpoint, the node calculates the state root of the checkpoint state repeatedly, which leads to bad UX/DX. In my current environment (M4 Macbook Air with WiFi connection of 200 mbps, syncing from Holesky), it takes about ~5 minutes for every state root calculation. 

This PR caches the calculation result so it is used in the very near future.

### To-Do

 
### DIFF
--- a/crates/common/fork_choice/src/store.rs
+++ b/crates/common/fork_choice/src/store.rs
@@ -775,8 +775,11 @@ pub fn get_forkchoice_store(
     anchor_state: BeaconState,
     anchor_block: BeaconBlock,
     db: ReamDB,
+    anchor_state_root: Option<B256>,
 ) -> anyhow::Result<Store> {
-    ensure!(anchor_block.state_root == anchor_state.tree_hash_root());
+    let anchor_state_root = anchor_state_root.unwrap_or_else(|| anchor_state.state_root());
+    ensure!(anchor_block.state_root == anchor_state_root);
+
     let anchor_root = anchor_block.tree_hash_root();
     let anchor_epoch = anchor_state.get_current_epoch();
     let justified_checkpoint = Checkpoint {
@@ -814,7 +817,7 @@ pub fn get_forkchoice_store(
     db.beacon_state_provider()
         .insert(anchor_root, anchor_state.clone())?;
     db.state_root_index_provider()
-        .insert(anchor_state.tree_hash_root(), anchor_root)?;
+        .insert(anchor_state_root, anchor_root)?;
     db.slot_index_provider()
         .insert(anchor_state.slot, anchor_root)?;
     db.checkpoint_states_provider()
--- a/crates/common/checkpoint_sync/src/lib.rs
+++ b/crates/common/checkpoint_sync/src/lib.rs
@@ -97,16 +97,13 @@ pub async fn initialize_db_from_checkpoint(
 
     info!("Fetching initial state...");
     let state = get_state(&checkpoint_sync_url, slot).await?;
-    info!(
-        "Downloaded state with root: {}. Slot: {}",
-        state.state_root(),
-        slot
-    );
+    let state_root = state.state_root();
+    info!("Downloaded state with root: {}. Slot: {}", state_root, slot);
 
     ensure!(block.message.slot == state.slot, "Slot mismatch");
 
-    ensure!(block.message.state_root == state.state_root());
-    let mut store = get_forkchoice_store(state.clone(), block.message, db)?;
+    ensure!(block.message.state_root == state_root);
+    let mut store = get_forkchoice_store(state.clone(), block.message, db, Some(state_root))?;
 
     let time = network_spec().min_genesis_time + SECONDS_PER_SLOT * (slot + 1);
     on_tick(&mut store, time)?;
--- a/testing/ef-tests/src/macros/fork_choice.rs
+++ b/testing/ef-tests/src/macros/fork_choice.rs
@@ -127,7 +127,7 @@ macro_rules! test_fork_choice {
 
                         let ream_dir = setup_data_dir("ream", None, true).expect("Failed to create data dir");
                         let reamdb = ReamDB::new(ream_dir).expect("count not find reabdb");
-                        let mut store = get_forkchoice_store(anchor_state, anchor_block, reamdb)
+                        let mut store = get_forkchoice_store(anchor_state, anchor_block, reamdb, None)
                             .expect("get_forkchoice_store failed");
 
                         for step in steps {


KolbyML (commented):
```
#[cfg(test)]
mod tests {
    use super::*;
    use ream_network_spec::networks::{MAINNET, set_network_spec};
    use tree_hash::TreeHash;

    #[tokio::test]
    async fn test_initialize_db_from_checkpoint() -> anyhow::Result<()> {
        set_network_spec(MAINNET.clone());
        let checkpoint_sync_url = get_checkpoint_sync_sources(None).remove(0);
        info!("Initiating checkpoint sync");

        info!("Fetching finalized block...");
        let block = fetch_finalized_block(&checkpoint_sync_url).await?;
        info!(
            "Downloaded block: {} with root: {}. Slot: {}",
            block.message.body.execution_payload.block_number,
            block.message.block_root(),
            block.message.slot
        );
        let slot = block.message.slot;

        info!("Fetching initial state...");
        let state = get_state(&checkpoint_sync_url, slot).await?;
        info!(
            "Downloaded state with root: {}. Slot: {}",
            state.state_root(),
            slot
        );

        // time computing the tree hash root for the state

        let start = std::time::Instant::now();
        let root = state.tree_hash_root();
        println!(
            "Computed tree hash root {root} for state in {:?}",
            start.elapsed()
        );

        ensure!(block.message.slot == state.slot, "Slot mismatch");

        ensure!(block.message.state_root == state.state_root());

        Ok(())
    }
}

```

Computed tree hash root 0x726cd92bf1c20fc6b783ecd94156a3bb2cb76a4d374788e25603adf56065b636 for state in 6.118892625s

I am not sure where you are getting the 5 minute number from. Mainnet's state only takes 6 seconds

Computed tree hash root 0x47598e1273923da17b1c44befee94c92a990c4792bc33df50c46fc612b54f4c2 for state in 8.497829792s 

^ this is holesky

Maybe you are running a debug build?

syjn99:
@KolbyML Yes, it was a debug profile. I forgot to add real numbers, so I brought some logs:

1. `cargo run beacon_node --network holesky --purge-db`

```
2025-07-03T18:21:07.136907Z  INFO ream_checkpoint_sync: Fetching initial state...
2025-07-03T18:22:28.169046Z  INFO ream_checkpoint_sync: Downloaded state for slot: 4638624 in 81.03255475s, now calculating the root
Computed tree hash root 0x2818c3e999acd567251ec33885a2facf282e459b16c200f7da43eedb03bea650 for state in 343.559310625s
2025-07-03T18:28:11.713311Z  INFO ream_checkpoint_sync: Downloaded state with root: 0x2818c3e999acd567251ec33885a2facf282e459b16c200f7da43eedb03bea650. Slot: 4638624
```

2.  `cargo run --release beacon_node --network holesky --purge-db`

```
2025-07-03T18:31:18.610351Z  INFO ream_checkpoint_sync: Fetching initial state...
2025-07-03T18:32:27.871648Z  INFO ream_checkpoint_sync: Downloaded state for slot: 4638656 in 69.261920625s, now calculating the root
Computed tree hash root 0xa0e51d764cedb10384fbd896338cae4bdbb7560a1352ae9b70fd1dbe8b90de9e for state in 14.966644542s
2025-07-03T18:32:42.838244Z  INFO ream_checkpoint_sync: Downloaded state with root: 0xa0e51d764cedb10384fbd896338cae4bdbb7560a1352ae9b70fd1dbe8b90de9e. Slot: 4638656
```

In release mode, it takes 14 seconds which seems quite a lot from my perspective

crates/common/fork_choice/src/store.rs
```diff
@@ -775,8 +775,11 @@ pub fn get_forkchoice_store(
     anchor_state: BeaconState,
     anchor_block: BeaconBlock,
     db: ReamDB,
+    anchor_state_root: Option<B256>,
```
KolbyML, line 778: I don't think passing the state root here is justified for something that only happens once initially. This seems like a premature optimization

crates/common/fork_choice/src/store.rs
```diff
@@ -775,8 +775,11 @@ pub fn get_forkchoice_store(
...
 ) -> anyhow::Result<Store> {
-    ensure!(anchor_block.state_root == anchor_state.tree_hash_root());
+    let anchor_state_root = anchor_state_root.unwrap_or_else(|| anchor_state.state_root());
+    ensure!(anchor_block.state_root == anchor_state_root);
```
KolbyML, line 781:
```suggestion
    let anchor_state_root = anchor_state.state_root();
    ensure!(anchor_block.state_root == anchor_state_root);
```

I agree it is good to reuse the root calculation, within this function

KolbyML:
> @KolbyML Yes, it was a debug profile. I forgot to add real numbers, so I brought some logs:
> 
> 1. `cargo run beacon_node --network holesky --purge-db`
> 
> ```
> 2025-07-03T18:21:07.136907Z  INFO ream_checkpoint_sync: Fetching initial state...
> 2025-07-03T18:22:28.169046Z  INFO ream_checkpoint_sync: Downloaded state for slot: 4638624 in 81.03255475s, now calculating the root
> Computed tree hash root 0x2818c3e999acd567251ec33885a2facf282e459b16c200f7da43eedb03bea650 for state in 343.559310625s
> 2025-07-03T18:28:11.713311Z  INFO ream_checkpoint_sync: Downloaded state with root: 0x2818c3e999acd567251ec33885a2facf282e459b16c200f7da43eedb03bea650. Slot: 4638624
> ```
> 
> 2. `cargo run --release beacon_node --network holesky --purge-db`
> 
> ```
> 2025-07-03T18:31:18.610351Z  INFO ream_checkpoint_sync: Fetching initial state...
> 2025-07-03T18:32:27.871648Z  INFO ream_checkpoint_sync: Downloaded state for slot: 4638656 in 69.261920625s, now calculating the root
> Computed tree hash root 0xa0e51d764cedb10384fbd896338cae4bdbb7560a1352ae9b70fd1dbe8b90de9e for state in 14.966644542s
> 2025-07-03T18:32:42.838244Z  INFO ream_checkpoint_sync: Downloaded state with root: 0xa0e51d764cedb10384fbd896338cae4bdbb7560a1352ae9b70fd1dbe8b90de9e. Slot: 4638656
> ```
> 
> In release mode, it takes 14 seconds which seems quite a lot from my perspective

I have an M3 air, so I am a bit confused how you are getting 14 seconds. 

This is only done once during initialization, and the state root calculation isn't even the bottleneck, downloading the state is the bottleneck, so this PR as a whole (other then the PR's description having bad numbers) seems like a premature optimization

KolbyML:
> @KolbyML Personally, it was a bad experience of waiting my node, as the logs doesn't show anything about the state root calculation. (I was curious about what's going on right now) I'm not also sure why my laptop needs more than seconds calculating the state root, but IMO I think this optimization is quite reasonable since it will reduce about 10> seconds of running initialize_db_from_checkpoint.

If calculating the state root takes 14 seconds you wouldn't be able to keep up with mainnet at all.

There are 2 major bottlenecks for syncing
- downloading the initial state
- syncing the node from the checkpoint to the head

If tree_hash_root()/calculating the state root is so slow then the root cause is most likely switching ethereum_hashing to use sha2, and since as you stated above you wouldn't be able to sync mainnet regardless of this PR, I am having trouble taking this PR seriously.

> @KolbyML Personally, it was a bad experience of waiting my node

If ^ this is your concern this PR does nothing, as 99.9% of the time will be spent sinking


IMO this PR isn't an efficient use of the teams resources, as we are trying to improve performance by 0.0001% compared to 99.999% of time that it takes to sync to HEAD. Performance/UX changes should be done once we have a working client at the head of the chain. The only exception to this would be if we are blocked by performance, which should be an extremely rare occurrence. 

>. it takes about ~5 minutes for every state root calculation.
^ If this was the case there would be no working clients on mainnet

By removing our patch for ethereum_hashing I get

Computed tree hash root 0xae63e5b351565d47ca641c339f6ae97a6f337f4872627642c3d9a19c1a44a477 for state in 2.859012s

^ Which implies we should make our patch of ethereum_hashing only use sha2 when being compiled to a riscv elf for zkVM's, using sha2 in general as it is very slow.

> but IMO I think this optimization is quite reasonable since it will reduce about 10> seconds of running initialize_db_from_checkpoint.

With everything I covered above this isn't what we should be focusing on. This PR only improves the first boot at that, even then downloading the state requires the vast majority of the time spent, but that doesn't matter as for the reasons above.

KolbyML (changes requested):
I was fine with us merging this if my suggestions were applied, but I think this PR is kind of missing the mark. 

> it takes about ~5 minutes for every state root calculation

^ the PR wasn't benchmarked properly or at all. If there is a slot every 12 seconds, how could calculating the state root take 5 minutes etc. Performance related PR's should have benchmarks of before and after, or how do we know the PR even improved performance or made it worse?

And for all the reasons I listed above, performance and UX just aren't the priority at this stage. The priority is getting a working product, then after that is done we can clean it up. For all we know this code will be deleted by the time we have a working product, so I am not sure why we are optimizing code we might not even keep.

syjn99:
Okay okay, I see some of your points are valid. To rephrase them for clarity:

- If we need 10+ seconds for just calculating a state root, it means this will not work properly in the "real world" use cases.
- Patching `ethereum_hashing` crate is critical for performance. I agree. (because it removes `ring` crate which is not written in pure Rust.)

But it found it quite dismissive when the PR title was intentionally changed to include "3 seconds" in quotes. I want to clarify my perspective:

1. This PR didn't waste any team resources, since it took me about 10 minutes to submit this PR. 
2. This PR doesn't mess up the codebase or introduce any technical debt, in my perspective.
3. I shared my logs in my environment (so we did get to know each other that some environment with debug profile can take over 5 minutes just for calculating a state root).
4. We all know UX is important, no? Even it is not our priority to dive into the performance and UX issues rn, I think the improvement can be valid in terms of our **contributor-friendly** vision. Excerpt from [John's message](https://t.me/ReamLabs/783) in our TG chat: 

> Hi I'm trying to start working on issues tagged with “Beacon API”. 
I’m running command
> cargo run beacon_node --network holesky --http-port 5052 --checkpoint-sync-url https://checkpoint-sync.holesky.ethpandaops.io -e
> Do I need to wait till checkpoint sync finished to access Beacon API endpoints? How long does it take normally? 
I'm trying:
> curl -X GET "http://localhost:5052/eth/v1/node/version" -H  "accept:application/json"
> but http://127.0.0.1:5052 is unreachable.

^ which implies this contributor also ran in a debug profile, and founds sth weird on our implementation. He would probably misunderstand that the ream node is stuck at:

```
2025-07-03T18:21:07.136907Z  INFO ream_checkpoint_sync: Fetching initial state...
```

I may suggest adding more logs before and after can be helpful in this context but have no idea of what's better.

---

I'm not trying to hinder our progress and journey. I just found it will be helpful to our codebase, in case of the other contributors trying to sync with mainnet or testnet.

Given the **strong** feedback like "missing the mark" and "not an efficient use of the teams resources", I will go ahead and close this PR.

syjn99:
Yeah thanks for your detailed and kind comment @KolbyML! I think I misunderstood your point at the first glimpse, but it seems much clear rn.

> It is clear our PR template needs improvement.

Yup it was my bad for writing the PR description as clear as it can be. I will definitely take care of this next time.

> Anyways let me know if you want to make PR's for any of these possible improvements, if not we can make issues.

I think we can keep our discussion up at the new GH issue.

Issue #626, opened 2025-07-03
Query the same peers on the network spec instead of peers on any network


Kayden-ML:
### Describe the feature

Querry the same peers on the network spec instead of peers on any network

PR #632, opened 2025-07-06
feat: only query peers on the same network, increases connected rate 6x

Kayden-ML:

Fixes: https://github.com/ReamLabs/ream/issues/626

It was implemented by only querying with the predicate of eth2

image


KolbyML (approved):
:shipit: looks good

PR #639, opened 2025-07-12
feat: add forward syncer

KolbyML:
### What was wrong?

We can't sync forward from the initial checkpoint

### How was it fixed?

Implement a forward syncer which takes our connected peers, downloads blocks and blobs, and executes them till we get to the finalized point.

### What I will do in follow up PR's

Currently syncing blobs will eventually freeze due to peers never responding and we don't implement timing out requests yet, so I need to implement timing out P2P requests.

### DIFF
--- a/crates/networking/syncer/src/block_range/block_cache.rs
+++ b/crates/networking/syncer/src/block_range/block_cache.rs
@@ -1,0 +1,327 @@
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Display,
+};
+
+use alloy_primitives::B256;
+use anyhow::{bail, ensure};
+use ream_consensus::{
+    blob_sidecar::{BlobIdentifier, BlobSidecar},
+    electra::beacon_block::SignedBeaconBlock,
+};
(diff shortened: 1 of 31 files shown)


crates/networking/syncer/src/block_range/block_cache.rs
```diff
@@ -0,0 +1,328 @@
...
+    electra::beacon_block::SignedBeaconBlock,
+};
+use ream_network_spec::networks::network_spec;
+use ssz::Encode;
+use tree_hash::TreeHash;
+
+use super::{MAX_BLOCKS_PER_REQUEST, peer_range_downloader::Range};
+
+/// The size of blobs is 1 MB (1_048_576 bytes)
+const BLOB_SIZE: u64 = 1_048_576;
```
jihoonsong: Isn't blob size 128 KB?

KolbyML: Yeah my bad, I will update that

Kayden-ML (approved):
Looks good nice work 👍

Issue #623, opened 2025-06-27
Implement waiting in submit attestation function


shariqnaiyer:
### Describe the feature

> 2. Wait for new BeaconBlock for the assigned slot (either stream updates or poll)
    - Max wait: `SECONDS_PER_SLOT / 3` seconds into the assigned slot

As laid out in the [validator flow](https://ethereum.github.io/beacon-APIs/validator-flow.md) doc, we need to wait for a new BeaconBlock to submit an attestation. 

A premature attestation will likely be incorrect and as "[attestations are penalised for being missing, late, or incorrect](https://eth2book.info/latest/part2/incentives/penalties/#attestation-penalties)", may incur penalties. 

The implementation will include a function to wait for either a valid BeaconBlock for the slot or for the first interval (1/3rd slot) to pass before making an attestation.

PR #640, opened 2025-07-14
Feature/shariqnaiyer/attestation wait

shariqnaiyer:
### What was wrong?

Fixes: #623 

### How was it fixed?

We wanted to implement a wait before making attestations which waits for either:
1. A valid block to be proposed to the Beacon node or
2. Passing of 1/3rd of the slot 

(2) is easy as we just implement a tokio::time::sleep()
For (1) we had to first subscribe to the Beacon API events stream on the 'head' topic, we then process each head topic by extracting the slot and keeping track of the highest slot for which a valid block has been seen. We also relay this to the waiting function using a channel.

### To-Do

 
### DIFF
--- a/crates/common/validator/src/validator.rs
+++ b/crates/common/validator/src/validator.rs
@@ -7,8 +7,9 @@ use std::{
 };
 
 use alloy_primitives::Address;
+use alloy_rpc_types_beacon::events::HeadEvent;
 use anyhow::{anyhow, bail};
-use futures::future::try_join_all;
+use futures::{StreamExt, future::try_join_all};
 use ream_beacon_api_types::{
     block::{BroadcastValidation, ProduceBlockData},
     duties::{AttesterDuty, ProposerDuty, SyncCommitteeDuty},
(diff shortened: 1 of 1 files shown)


KolbyML:
@Patchoulis do you think we need stream for this? I don't like the way channels are being used here, but I need your answer first before diving in

crates/common/validator/beacon/src/validator.rs
```diff
@@ -131,8 +143,26 @@ impl ValidatorService {
         };
         interval.set_missed_tick_behavior(MissedTickBehavior::Burst);
 
+        let mut event_stream = match self
```
Patchoulis, line 146: Seeing as the current logic is using system time waits, does it not make sense to do it like that?

crates/common/validator/beacon/src/validator.rs
```diff
@@ -498,12 +538,33 @@ impl ValidatorService {
...
+
+        let mut slot_receiver = self.slot_sender.subscribe();
+
+        tokio::select! {
+            Ok(received_slot) = slot_receiver.recv() => {
+                if received_slot >= slot {
+                    info!("Received slot {received_slot}, proceeding with attestation");
+                }
+            }
+            _ = tokio::time::sleep(Duration::from_secs(network_spec().seconds_per_slot / 3)) => {
```
Patchoulis, line 554: This one I need to confirm with @KolbyML , but I remember he wanted us to use Interval if I'm not mistaken?

Also, the 3 is a network spec constant. Intervals per slot I believe? I'd say replace the 3 with the actual constant.

Patchoulis (changes requested):

PR #642, opened 2025-07-17
fix: wrong index was used in process_attestation

KolbyML:
### What was wrong?

data.index was used instead of committee_index

which caused

```
        ensure!(
            attestation.aggregation_bits.len() == committee_offset,
            "Aggregation bits length must match committee size"
            "Aggregation bits length must match committee size {} != {}",
            attestation.aggregation_bits.len(),
            committee_offset
        );
```
To fail when I was executing blocks

### How was it fixed?

Use the correct index


syjn99 (approved):
Good catch!

PR #643, opened 2025-07-18
perf: don't recompute base_reward_per_increment for every participant

KolbyML:
### What was wrong?

I am working on fixing our networking code so the forward syncer works. I got the forward syncer working reliable, but now the problem is it takes multiple hours? possibly days to execute a block.

I debugged the problem and figure out it was because we were calling ``self.get_base_reward_per_increment()`` every time we calculated the base_reward

```rust
    pub fn get_base_reward_per_increment(&self) -> u64 {
        EFFECTIVE_BALANCE_INCREMENT * BASE_REWARD_FACTOR
            / integer_squareroot(self.get_total_active_balance())
    }
    
        pub fn get_total_active_balance(&self) -> u64 {
        self.get_total_balance(
            self.get_active_validator_indices(self.get_current_epoch())
                .into_iter()
                .collect::<HashSet<_>>(),
        )
    }
    
    pub fn get_active_validator_indices(&self, epoch: u64) -> Vec<u64> {
        self.validators
            .iter()
            .enumerate()
            .filter_map(|(i, validator)| {
                if validator.is_active_validator(epoch) {
                    Some(i as u64)
                } else {
                    None
                }
            })
            .collect()
    }
```    

^ So basically we had an absolutely horrible time complexity, O(n ^ n) or something dumb everytime we called ``process_attestation``.

### How was it fixed?

Calculate `base_reward_per_increment` once, instead of recomputing it everytime, which makes `get_base_reward` run O(1) time complexity.

### DIFF
--- a/crates/common/consensus/src/electra/beacon_state.rs
+++ b/crates/common/consensus/src/electra/beacon_state.rs
@@ -748,14 +748,15 @@ impl BeaconState {
 
     /// Return the base reward for the validator defined by ``index`` with respect to the current
     /// ``state``.
-    pub fn get_base_reward(&self, index: u64) -> u64 {
+    pub fn get_base_reward(&self, index: u64, base_reward_per_increment: u64) -> u64 {
         let increments =
             self.validators[index as usize].effective_balance / EFFECTIVE_BALANCE_INCREMENT;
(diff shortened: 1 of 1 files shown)


Kayden-ML (approved):
Looks good 👍

PR #646, opened 2025-07-19
perf: don't recalculate unslashed_participating_indices redundantly

KolbyML:
### What was wrong?

Processing an epoch takes forever
### How was it fixed?

don't calculate get_unslashed_participating_indices every loop interation


Kayden-ML (approved):
Looks good 👍

PR #647, opened 2025-07-20
feat: implement /eth/v1/validator/attestation_data

jveer634:
### What was wrong?

 Missing attestation_data validator endpoint for the electra hardfork.

### How was it fixed?

- I have added some helper functions on the unrealized_checkpoint_table and justified_checkpoint_table to get the checkpoint details.
- Created a new request type for the API endpoint.
- Fetched data from the checkpoints table and highest block root from the slot index table. And returned the data object

### To-Do
- Logic Verification
 
### DIFF
--- a/crates/common/fork_choice/src/store.rs
+++ b/crates/common/fork_choice/src/store.rs
@@ -802,6 +802,22 @@ impl Store {
 
         Ok(())
     }
+
+    pub fn is_syncing(&self) -> anyhow::Result<bool> {
+        let head = self.get_head()?;
+
+        let head_slot = match self.db.beacon_block_provider().get(head) {
+            Ok(Some(block)) => block.message.slot,
+            err => {
+                return Err(anyhow!("Failed to get head slot, error: {err:?}"));
+            }
+        };
+
+        // calculate sync_distance
+        let sync_distance = self.get_current_slot()?.saturating_sub(head_slot);
+
+        Ok(sync_distance > 1)
+    }
 }
 
 pub fn get_forkchoice_store(
--- a/crates/common/beacon_api_types/src/error.rs
+++ b/crates/common/beacon_api_types/src/error.rs
@@ -23,6 +23,9 @@ pub enum ApiError {
 
     #[error("Too many validator IDs in request")]
     TooManyValidatorsIds,
+
+    #[error("Beacon node is currently syncing and not serving request on that endpoint")]
+    UnderSyncing,
 }
 
 impl ResponseError for ApiError {
@@ -39,6 +42,7 @@ impl ResponseError for ApiError {
             ApiError::InvalidParameter(_) => StatusCode::BAD_REQUEST,
             ApiError::ValidatorNotFound(_) => StatusCode::NOT_FOUND,
             ApiError::TooManyValidatorsIds => StatusCode::URI_TOO_LONG,
+            ApiError::UnderSyncing => StatusCode::SERVICE_UNAVAILABLE,
         }
     }
 }
--- a/crates/common/beacon_api_types/src/query.rs
+++ b/crates/common/beacon_api_types/src/query.rs
@@ -44,6 +44,12 @@ pub struct StatusQuery {
     pub status: Option<Vec<ValidatorStatus>>,
 }
 
+#[derive(Default, Debug, Deserialize)]
+pub struct AttestationQuery {
+    pub slot: u64,
+    pub committee_index: u64,
+}
+
 impl StatusQuery {
     pub fn has_status(&self) -> bool {
         match &self.status {
--- a/crates/rpc/beacon/src/handlers/validator.rs
+++ b/crates/rpc/beacon/src/handlers/validator.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::{collections::HashSet, sync::Arc};
 
 use actix_web::{
     HttpResponse, Responder, get, post,
@@ -7,19 +7,25 @@ use actix_web::{
 use ream_beacon_api_types::{
     error::ApiError,
     id::{ID, ValidatorID},
-    query::{IdQuery, StatusQuery},
+    query::{AttestationQuery, IdQuery, StatusQuery},
     request::ValidatorsPostRequest,
-    responses::BeaconResponse,
+    responses::{BeaconResponse, DataResponse},
     validator::{ValidatorBalance, ValidatorData, ValidatorStatus},
 };
 use ream_bls::PublicKey;
 use ream_consensus_beacon::electra::beacon_state::BeaconState;
-use ream_consensus_misc::{constants::SLOTS_PER_EPOCH, validator::Validator};
-use ream_storage::db::ReamDB;
+use ream_consensus_misc::{
+    attestation_data::AttestationData, constants::SLOTS_PER_EPOCH, validator::Validator,
+};
+use ream_fork_choice::store::Store;
+use ream_operation_pool::OperationPool;
+use ream_storage::{db::ReamDB, tables::Field};
 use serde::Serialize;
 
 use super::state::get_state_from_id;
 
+///  For slots in Electra and later, this AttestationData must have a committee_index of 0.
+const ELECTRA_COMMITTEE_INDEX: u64 = 0;
 const MAX_VALIDATOR_COUNT: usize = 100;
 
 fn build_validator_balances(
@@ -422,4 +428,60 @@ fn check_validator_participation(
     } else {
         Ok(validator.is_active_validator(epoch))
     }
+}
+#[get("/validator/attestation_data")]
+pub async fn get_attestation_data(
+    db: Data<ReamDB>,
+    opertation_pool: Data<Arc<OperationPool>>,
+    query: Query<AttestationQuery>,
+) -> Result<impl Responder, ApiError> {
+    let store = Store {
+        db: db.get_ref().clone(),
+        operation_pool: opertation_pool.get_ref().clone(),
+    };
+
+    if store.is_syncing().map_err(|err| {
+        ApiError::InternalError(format!("Failed to check syncing status, err: {err:?}"))
+    })? {
+        return Err(ApiError::UnderSyncing);
+    }
+
+    let slot = query.slot;
+
+    let current_slot = store
+        .get_current_slot()
+        .map_err(|err| ApiError::InternalError(format!("Failed to slot_index, error: {err:?}")))?;
+
+    if slot > current_slot + 1 {
+        return Err(ApiError::InvalidParameter(format!(
+            "Slot {slot:?} is too far ahead of the current slot {current_slot:?}"
+        )));
+    }
+
+    let beacon_block_root = db
+        .slot_index_provider()
+        .get_highest_root()
+        .map_err(|err| ApiError::InternalError(format!("Failed to slot_index, error: {err:?}")))?
+        .ok_or(ApiError::NotFound(
+            "Failed to find highest block root".to_string(),
+        ))?;
+
+    let source_checkpoint = db.justified_checkpoint_provider().get().map_err(|err| {
+        ApiError::InternalError(format!("Failed to get source checkpoint, error: {err:?}"))
+    })?;
+
+    let target_checkpoint = db
+        .unrealized_justified_checkpoint_provider()
+        .get()
+        .map_err(|err| {
+            ApiError::InternalError(format!("Failed to target checkpoint, error: {err:?}"))
+        })?;
+
+    Ok(HttpResponse::Ok().json(DataResponse::new(AttestationData {
+        slot,
+        index: ELECTRA_COMMITTEE_INDEX,
+        beacon_block_root,
+        source: source_checkpoint,
+        target: target_checkpoint,
+    })))
 }
--- a/crates/rpc/beacon/src/routes/validator.rs
+++ b/crates/rpc/beacon/src/routes/validator.rs
@@ -3,10 +3,12 @@ use actix_web::web::ServiceConfig;
 use crate::handlers::{
     duties::{get_attester_duties, get_proposer_duties},
     prepare_beacon_proposer::prepare_beacon_proposer,
+    validator::get_attestation_data,
 };
 
 pub fn register_validator_routes(config: &mut ServiceConfig) {
     config.service(get_proposer_duties);
     config.service(get_attester_duties);
     config.service(prepare_beacon_proposer);
+    config.service(get_attestation_data);
 }


crates/rpc/src/handlers/validator.rs
```diff
@@ -424,3 +425,40 @@ fn check_validator_participation(
...
+
+    let data = AttestationData {
+        slot: query.slot,
+        index: query.committee_index,
+        beacon_block_root,
+        source,
+        target,
+    };
+
+    Ok(HttpResponse::Ok().json(BeaconResponse::new(data)))
```
syjn99:
```suggestion
    Ok(HttpResponse::Ok().json(DataResponse::new(data)))
```

I think we should use `DataResponse` to be consistent with the spec

crates/rpc/src/handlers/validator.rs
```diff
@@ -424,3 +425,40 @@ fn check_validator_participation(
...
+    let target = db
+        .unrealized_justified_checkpoint_provider()
+        .get_unrealized_checkpoint()
+        .map_err(|err| {
+            ApiError::InternalError(format!("Failed to target checkpoint, error: {err:?}"))
+        })?;
+
+    let data = AttestationData {
+        slot: query.slot,
+        index: query.committee_index,
```
syjn99: >  For slots in Electra and later, this AttestationData must have a committee_index of 0.

We also need to fix this, and I think we don't have to consider pre-electra case.

crates/storage/src/tables/justified_checkpoint.rs
```diff
@@ -41,3 +41,9 @@ impl Field for JustifiedCheckpointField {
         Ok(())
     }
 }
+
+impl JustifiedCheckpointField {
+    pub fn get_justified_checkpoint(&self) -> Result<Checkpoint, StoreError> {
```
syjn99: I wonder why this function is needed, can we just use `.get()`?

syjn99 (changes requested):
first look! thank you for your work 😁

crates/storage/src/tables/justified_checkpoint.rs
jveer634: The get function is not public to use outside the crate, so, had to write a public wrapper around it.

syjn99: You may import `ream_storage::tables::Field` trait to handle it

syjn99:
Also I think @shariqnaiyer or @Patchoulis should take a look at this, as the validator client will call this HTTP API to build an attestation.

crates/rpc/src/handlers/validator.rs
```diff
@@ -424,3 +425,40 @@ fn check_validator_participation(
...
+    let target = db
+        .unrealized_justified_checkpoint_provider()
+        .get_unrealized_checkpoint()
+        .map_err(|err| {
+            ApiError::InternalError(format!("Failed to target checkpoint, error: {err:?}"))
+        })?;
+
+    let data = AttestationData {
+        slot: query.slot,
+        index: 0,
```
syjn99: Would you mind adding a note comment for this?

crates/rpc/src/handlers/validator.rs
```diff
@@ -424,3 +425,40 @@ fn check_validator_participation(
...
+
+    let target = db
+        .unrealized_justified_checkpoint_provider()
+        .get_unrealized_checkpoint()
+        .map_err(|err| {
+            ApiError::InternalError(format!("Failed to target checkpoint, error: {err:?}"))
+        })?;
+
+    let data = AttestationData {
+        slot: query.slot,
```
syjn99: Also, I think `slot` should also be validated. 

Simplest validation will check whether requested slot is timely or not. When a node propogates an attestation, it should not be future slot. Also, it is useless to propagate a stale attestation.

- Upper bound: Now + `MAXIMUM_GOSSIP_CLOCK_DISPARITY` (= 500ms) 
(Ref: see [description](https://github.com/ethereum/consensus-specs/blob/dev/specs/phase0/p2p-interface.md#why-is-there-maximum_gossip_clock_disparity-when-validating-slot-ranges-of-messages-in-gossip-subnets) in phase0 spec.)
- Lower bound: the first slot of the previous epoch. (From [EIP-7045](https://eips.ethereum.org/EIPS/eip-7045))

I'm not sure where this validation takes place, but as beacon node has more information about the chain, I believe this should be done at the beacon node level. Any thoughts? @shariqnaiyer

jnalam-ith: I think since we are directly dealing with slot numbers, this could be a good validation 
```
    if slot < current_slot -1 :
        block_root = block_roots[slot]
    else if slot == current_slot or slot == current_slot + 1:
        block_root = latest_block_root
    else:
        throw Error
```

I believe this is the case, because in the API, we are just providing the data rather than providing attestation by the validator.

I think it makes sense to validate the slot on publishing attestation rather than getting attestation data

jveer634: I think since we are directly dealing with slot numbers, this could be a good validation

    if slot < current_slot -1 :
        block_root = block_roots[slot]
    else if slot == current_slot or slot == current_slot + 1:
        block_root = latest_block_root
    else:
        throw Error
I believe this is the case, because in the API, we are just providing the data rather than providing attestation by the validator.

I think it makes sense to validate the slot on publishing attestation rather than getting attestation data

crates/rpc/src/handlers/validator.rs
jveer634: As in why we we have 0 ?

crates/rpc/src/handlers/validator.rs
jveer634: If the above logic works, then update on source and target checkpoints is also needed

crates/rpc/src/handlers/validator.rs
syjn99: Yes please 🙏! It seems quite strange without context haha

jveer634: Will push with the remaining changes

crates/rpc/src/handlers/validator.rs
syjn99: @jveer634 Yeah, your validation logic makes sense. We should have one slot buffer in case of clock drift. 

We already have `get_block_root_at_slot` function, so you might use this instead of `block_roots[slot]`.

jveer634: @syjn99  Is this what you mean ?/

```    let source_epoch = compute_epoch_at_slot(slot);
    let target_epoch = source_epoch + 1;

    let epoch_start_slot = compute_start_slot_at_epoch(source_epoch);
    let epoch_end_slot = epoch_start_slot + SLOTS_PER_EPOCH;

    let state = db
        .get_latest_state()
        .map_err(|err| ApiError::InternalError(format!("Failed to get latest State, err {err}")))?;

    let source_root = state.get_block_root(source_epoch).map_err(|err| {
        ApiError::InternalError(format!("Failed to get source checkpoint, error: {err:?}"))
    })?;

    let target_root = state.get_block_root(target_epoch).map_err(|err| {
        ApiError::InternalError(format!("Failed to get target checkpoint, error: {err:?}"))
    })?;```

File: crates/rpc/src/handlers/validator.rs [hunk 6]
```diff
@@ -424,3 +434,109 @@ fn check_validator_participation(
...
+        operation_pool: operation_pool.get_ref().clone(),
+    };
+    let slot = query.slot;
+    let current_slot = store.get_current_slot().map_err(|err| {
+        ApiError::InternalError(format!("Failed to get current slot, error: {err:?}"))
+    })?;
+
+    if slot > current_slot + 1u64 {
+        return Err(ApiError::InvalidParameter(format!(
+            "slot too far in the future {slot:?}"
```
syjn99:
```suggestion
            "Slot {slot:?} is too far ahead of the current slot {current_slot:?}"
```
Just a nit

crates/rpc/src/handlers/validator.rs
```diff
@@ -424,3 +434,109 @@ fn check_validator_participation(
...
+        let data = AttestationData {
+            slot,
+            index: ELECTRA_COMMITTEE_INDEX,
+            beacon_block_root,
+            source,
+            target,
+        };
+
+        return Ok(HttpResponse::Ok().json(DataResponse::new(data)));
+    }
```
syjn99: Thinking again, I think filtering `slot > current_slot + 1u64` is enough for this API. Below this line, you are trying to "mock" the attestation at the past which is not right. (e.g., the difference between source and target doesn't have to be an epoch)

So, I'd like to suggest like following:

```rust

    if slot > current_slot + 1u64 {
        return Err(ApiError::InvalidParameter(format!(
            "Slot {slot:?} is too far ahead of the current slot {current_slot:?}"
        )));
    }
    
    // Skip for brevity...

    let data = AttestationData {
        slot,
        index: ELECTRA_COMMITTEE_INDEX,
        beacon_block_root,
        source,
        target,
    };

    return Ok(HttpResponse::Ok().json(DataResponse::new(data)));
```

syjn99 (commented):
We are almost there! Added one more concern about rebuilding an attestation that might not be right. It's enough for `beacon_node` to share its own view at that time to the validator client.

crates/rpc/src/handlers/validator.rs
syjn99: Let's keep discussing at the [comment](https://github.com/ReamLabs/ream/pull/647#discussion_r2222290254) below. Actually, we don't have to fetch the root from `block_root`, my bad.

crates/rpc/src/handlers/validator.rs
jveer634: At first, I though of the same to just provide the latest info. But later I though of, when we have the record to provide the attestation of a previous record, why not??

What is people would come up with a use case for that

syjn99: @jveer634 

Reconstructing attestation for the past slot is **dangerous**. It might cause the validator slashed for double voting or surround voting.

So what you're trying to build an attestation data is based on the naive assumption:

> The source and the target checkpoints do not need to be consecutive; it is permissible to jump checkpoints.

Excerpt from [eth2book](https://eth2book.info/capella/part2/consensus/casper_ffg/#sources-and-targets-links-and-conflicts). We cannot guarantee the distance between source and target checkpoint.

So I think this API should only share the current view of the node to the validator client with the minimum slot validation, which is already implemented well!

jveer634: Which means the first iteration was good enough?

And also I did miss the syncing state reponse though, need to add that

syjn99: > Which means the first iteration was good enough?

Yup!

> And also I did miss the syncing state reponse though, need to add that

Sounds cool!

crates/storage/src/tables/justified_checkpoint.rs
```diff
@@ -41,3 +41,9 @@ impl Field for JustifiedCheckpointField {
...
+        self.get()
+    }
+}
```
syjn99: We can delete this one and other one if not used.

crates/common/fork_choice/src/store.rs
```diff
@@ -801,6 +801,24 @@ impl Store {
 
         Ok(())
     }
+
+    pub fn is_syncing(&self) -> anyhow::Result<bool> {
```
syjn99, line 806: I think this function can be cleaner but it shares almost similar structure with the function `get_syncing_status`. It seems fine in this stage, will take this as next task.

syjn99 (approved):
Looks good to me. @shariqnaiyer can check out this PR once!

shariqnaiyer:
@jveer634 I will be reviewing and testing this today. Please make the final changes requested by @syjn99 so we can merge it after I test it.

crates/common/fork_choice/src/store.rs
jveer634: We will be needing this function in some more endpoints too. So making it as a helper function makes more sense than use another API requestion.

jveer634:
@shariqnaiyer I will include them after you review.  Or do you want delete the dead code before review ??

shariqnaiyer:
> @shariqnaiyer I will include them after you review. Or do you want delete the dead code before review ??

Resolve the conflicts and should be good to review!

crates/common/fork_choice/src/store.rs
jveer634: @shariqnaiyer We will be requiring this helper function in many places. So will be leaving this out. What do you say ?

crates/common/fork_choice/src/store.rs
```diff
@@ -801,6 +801,24 @@ impl Store {
...
+            Ok(Some(block)) => block.message.slot,
+            err => {
+                return Err(anyhow!("Failed to get head slot, error: {err:?}"));
+            }
+        };
+
+        // calculate sync_distance
+        let current_slot = self.get_current_slot()?;
+
+        let sync_distance = current_slot.saturating_sub(head_slot);
```
KolbyML:
```suggestion
        let head_slot = match self.db.beacon_block_provider().get(self.get_head()?) {
            Ok(Some(block)) => block.message.slot,
            err => {
                return Err(anyhow!("Failed to get head slot, error: {err:?}"));
            }
        };

        // calculate sync_distance
        let sync_distance = self.get_current_slot()?.saturating_sub(head_slot);
```

we might as well inline these right values as making variables for them doesn't add much value

crates/rpc/src/handlers/validator.rs
```diff
@@ -7,22 +7,28 @@ use actix_web::{
...
+use ream_operation_pool::OperationPool;
+use ream_storage::{db::ReamDB, tables::Field};
 use serde::Serialize;
 
 use super::state::get_state_from_id;
 
 const MAX_VALIDATOR_COUNT: usize = 100;
 
+//  For slots in Electra and later, this AttestationData must have a committee_index of 0.
+const ELECTRA_COMMITTEE_INDEX: u64 = 0;
```
KolbyML:
```suggestion
///  For slots in Electra and later, this AttestationData must have a committee_index of 0.
const ELECTRA_COMMITTEE_INDEX: u64 = 0;
const MAX_VALIDATOR_COUNT: usize = 100;
```

the comment should be a doc comment and sorted

crates/rpc/src/handlers/validator.rs
```diff
@@ -424,3 +430,61 @@ fn check_validator_participation(
...
+
+    let current_slot = store
+        .get_current_slot()
+        .map_err(|err| ApiError::InternalError(format!("Failed to slot_index, error: {err:?}")))?;
+
+    if slot > current_slot + 1u64 {
+        return Err(ApiError::InvalidParameter(format!(
+            "Slot {slot:?} is too far ahead of the current slot {current_slot:?}"
+        )));
+    }
```
KolbyML:
```suggestion
    if slot > current_slot + 1 {
        return Err(ApiError::InvalidParameter(format!(
            "Slot {slot:?} is too far ahead of the current slot {current_slot:?}"
        )));
    }
```

the type declaration here isn't needed

crates/rpc/src/handlers/validator.rs
```diff
@@ -424,3 +430,61 @@ fn check_validator_participation(
...
+        .map_err(|err| ApiError::InternalError(format!("Failed to slot_index, error: {err:?}")))?
+        .ok_or(ApiError::NotFound(
+            "Failed to find highest block root".to_string(),
+        ))?;
+
+    let source = db.justified_checkpoint_provider().get().map_err(|err| {
+        ApiError::InternalError(format!("Failed to get source checkpoint, error: {err:?}"))
+    })?;
+
+    let target = db
```
KolbyML: can you write more descriptive variable names

crates/rpc/src/handlers/validator.rs
```diff
@@ -424,3 +430,61 @@ fn check_validator_participation(
...
+
+    let data = AttestationData {
+        slot,
+        index: ELECTRA_COMMITTEE_INDEX,
+        beacon_block_root,
+        source,
+        target,
+    };
+
+    Ok(HttpResponse::Ok().json(DataResponse::new(data)))
```
KolbyML:
```suggestion
    Ok(HttpResponse::Ok().json(DataResponse::new(AttestationData {
        slot,
        index: ELECTRA_COMMITTEE_INDEX,
        beacon_block_root,
        source,
        target,
    })))
```

this should just be inlined, as the `data` variable isn't used, so there isn't much point in declaring a left value for it

crates/storage/src/tables/unrealized_justified_checkpoint.rs
```diff
@@ -41,3 +41,9 @@ impl Field for UnrealizedJustifiedCheckpointField {
         Ok(())
     }
 }
+
+impl UnrealizedJustifiedCheckpointField {
+    pub fn get_unrealized_checkpoint(&self) -> Result<Checkpoint, StoreError> {
+        self.get()
+    }
+}
```
KolbyML: please remove this and call `get()` directly

KolbyML (commented):
Here is some feedback

crates/storage/src/tables/justified_checkpoint.rs
KolbyML: @jveer634 What @syjn99 said is the correct path forward, I am not sure why this was marked as resolved

jnalam-ith: Sorry I made the changes on the handler, but forget to delete the function

jveer634: Sorry. I made the changes on the handler, but forgot to delete the dead code

crates/rpc/src/handlers/validator.rs
jveer634: Can we include rules for these in clippy ? I have made changes though

crates/rpc/src/handlers/validator.rs
jveer634: please check now

PR #648, opened 2025-07-20
fix: up network code to get forward syncer more reliable

KolbyML:
### What was wrong?

The network code wasn't robust enough to reliably forward sync

### How was it fixed?

- Add a few connection/callback timeouts
- reduce noisy logs


Kayden-ML (approved):
Looks good nice work 👍

PR #652, opened 2025-07-23
chore: bump rust version to 1.88 & make linter happy

syjn99:
### What was wrong?

 

Rust `1.88.0` is now stable. ([changelog](https://releases.rs/docs/1.88.0/)) I wanted to use let-chain because it makes the code cleaner.

### How was it fixed?

 

Bump Rust version to `1.88.0` and fix linting issue when I run `make lint`.

Also I forgot to update our book 😂, this PR also introduces a new command `lean_node` in our book.

### To-Do

 
### DIFF
--- a/crates/common/checkpoint_sync/src/lib.rs
+++ b/crates/common/checkpoint_sync/src/lib.rs
@@ -35,6 +35,7 @@ use weak_subjectivity::{WeakSubjectivityState, verify_state_from_weak_subjectivi
 ///     "data" : T
 /// })
 /// }
+#[allow(dead_code)]
 #[derive(Debug, Serialize, Deserialize)]
 struct OptionalBeaconVersionedResponse<T> {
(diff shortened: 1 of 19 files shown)


crates/common/checkpoint_sync/src/lib.rs
```diff
@@ -35,6 +35,7 @@ use weak_subjectivity::{WeakSubjectivityState, verify_state_from_weak_subjectivi
 ///     "data" : T
 /// })
 /// }
+#[allow(dead_code)]
```
varun-doshi: Can we remove this struct if it is not used anywhere?

varun-doshi (commented):
small nit

crates/common/checkpoint_sync/src/lib.rs
syjn99: sure

PR #664, opened 2025-07-27
chore: add lean network spec

### LINKED ISSUE #663: Add NetworkSpec for Lean
### Describe the feature

Although the specification is not yet finalized (See ongoing discussion in this [PR](https://github.com/leanEthereum/pm/pull/4)), we must initialize different network spec from the one of Beacon chain.

I think `SECONDS_PER_SLOT` and `GENESIS_TIME` are two must-contained values, so we can start with minimum specification.

### Additional context

_No response_

syjn99:
### What was wrong?

 

Closes #663 

### How was it fixed?

 

~~In this PR I Introduced a trait `SingletonSpec`. Using `SingletonSpec`, it is now available to use `set_network_spec` and `network_spec` more generic way. (Feedbacks are welcomed, especially for the point that this change *might* seem to be over-engineered.)~~

I just renamed current `NetworkSpec` into `BeaconNetworkSpec` and add `LeanNetworkSpec` after the [convo](https://github.com/ReamLabs/ream/pull/664#issuecomment-3124788686) with Kolby.

You can now run `lean_node` with custom specification by running the command:

```bash
cargo run --release lean_node --network fancy_spec.yml
```

where `fancy_spec.yml` looks like:

```yaml
SECONDS_PER_SLOT: 12
GENESIS_TIME: 0
```

### To-Do

 

bin/ream/assets/lean/sample_spec.yml
```diff
@@ -0,0 +1,2 @@
+SECONDS_PER_SLOT: 12
+GENESIS_TIME: 0
```
Kayden-ML: add a new line at the end of the file

KolbyML:
Can we add the LeanNetworkSpec without generics, we will likely want to run both the Beacon and Lean client at the same time during the transition, so making it so you can only set one wouldn't be desired. Also generics are a little overkill here and make the code less readable, so I don't think it is justified here.

syjn99:
@KolbyML yeah, thank you for the feedback!

> so making it so you can only set one wouldn't be desired

Actually there are two `OnceLock` instances here for Beacon and Lean respectively, so it would be no problem in this context. You can just call `set_network_specs` with Lean spec.

> Also generics are a little overkill here and make the code less readable

Yeah, indeed, I was also worried about this point. Is it clear to rename `NetworkSpec` -> `BeaconNetworkSpec` / `network_specs()` -> `beacon_network_specs()`? Or do you want the spec for Beacon remained intact and add just `LeanNetworkSpec`?

KolbyML:
> @KolbyML yeah, thank you for the feedback!
> 
> > so making it so you can only set one wouldn't be desired
> 
> Actually there are two `OnceLock` instances here for Beacon and Lean respectively, so it would be no problem in this context. You can just call `set_network_specs` with Lean spec.

Oh nice, I missed that detail when looking the PR over quick initially, thanks for pointing that out

> > Also generics are a little overkill here and make the code less readable
> 
> Yeah, indeed, I was also worried about this point. Is it clear to rename `NetworkSpec` -> `BeaconNetworkSpec` / `network_specs()` -> `beacon_network_specs()`? Or do you want the spec for Beacon remained intact and add just `LeanNetworkSpec`?

`NetworkSpec` -> `BeaconNetworkSpec` / `network_specs()` -> `beacon_network_specs()`

^ I think renaming these would be beneficial :+1:

syjn99:
@KolbyML @Kayden-ML Ready for review.

PR #666, opened 2025-07-28
feat: use simple 1/3 slot wait

shariqnaiyer:
### What was wrong?

Fixes: #623 

### How was it fixed?

We wanted to implement a wait before making attestations which waits for either:
1. A valid block to be proposed to the Beacon node or
2. Passing of 1/3rd of the slot 

For now we only wait 1/3rd of the slot. We will later optimize it to use the events stream to get the head event and wait for valid block as in #640. However setting up the event stream properly will take time so we are moving forward with a simple sleep function that will work.

### To-Do

 

KolbyML (approved):
:shipit: looks good

PR #668, opened 2025-07-28
feat: Initial implementation of post quantum crypto + account_manager changes

ch4r10t33r:
[… to account_manager](feat: Initial implementation of post quantum crypto + account_manager changes)

 
- [X] This PR uses [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5f7b0a21988c1bf877cf4759ef5ddaac04c1c9fe808c9142ecb78ba97d97a28a"
 dependencies = [
- "bitflags 2.9.0",
+ "bitflags 2.9.1",
  "bytes",
  "futures-core",
  "futures-sink",
@@ -21,16 +21,16 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.10.0"
+version = "3.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0fa882656b67966045e4152c634051e70346939fced7117d5f0b52146a7c74c9"
+checksum = "44dfe5c9e0004c623edc65391dfd51daa201e7e30ebd9c9bedf873048ec32bc2"
 dependencies = [
  "actix-codec",
  "actix-rt",
  "actix-service",
  "actix-utils",
  "base64 0.22.1",
- "bitflags 2.9.0",
+ "bitflags 2.9.1",
  "brotli",
  "bytes",
  "bytestring",
@@ -39,7 +39,7 @@ dependencies = [
  "flate2",
  "foldhash",
  "futures-core",
- "h2 0.3.26",
+ "h2 0.3.27",
  "http 0.2.12",
  "httparse",
  "httpdate",
@@ -49,7 +49,7 @@ dependencies = [
  "mime",
  "percent-encoding",
  "pin-project-lite",
- "rand 0.9.1",
+ "rand 0.9.2",
  "sha1",
  "smallvec",
  "tokio",
@@ -65,7 +65,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.101",
+ "syn 2.0.104",
 ]
 
 [[package]]
@@ -95,9 +95,9 @@ dependencies = [
 
 [[package]]
 name = "actix-server"
-version = "2.5.1"
+version = "2.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6398974fd4284f4768af07965701efbbb5fdc0616bff20cade1bb14b77675e24"
+checksum = "a65064ea4a457eaf07f2fba30b4c695bf43b721790e9530d26cb6f9019ff7502"
 dependencies = [
  "actix-rt",
  "actix-service",
@@ -105,7 +105,7 @@ dependencies = [
  "futures-core",
  "futures-util",
  "mio",
- "socket2",
+ "socket2 0.5.10",
  "tokio",
  "tracing",
 ]
@@ -132,9 +132,9 @@ dependencies = [
 
 [[package]]
 name = "actix-web"
-version = "4.10.2"
+version = "4.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f2e3b15b3dc6c6ed996e4032389e9849d4ab002b1e92fbfe85b5f307d1479b4d"
+checksum = "a597b77b5c6d6a1e1097fddde329a83665e25c5437c696a3a9a4aa514a614dea"
 dependencies = [
  "actix-codec",
  "actix-http",
@@ -167,7 +167,7 @@ dependencies = [
  "serde_json",
  "serde_urlencoded",
  "smallvec",
- "socket2",
+ "socket2 0.5.10",
  "time",
  "tracing",
  "url",
@@ -182,14 +182,14 @@ dependencies = [
  "actix-router",
  "proc-macro2",
  "quote",
- "syn 2.0.101",
+ "syn 2.0.104",
 ]
 
 [[package]]
 name = "actix-web-lab"
-version = "0.24.1"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a33034dd88446a5deb20e42156dbfe43d07e0499345db3ae65b3f51854190531"
+checksum = "781b41742426e73ab2fe531a7d204b9ffa0815a63e358bcd24f2c3c69cda644a"
 dependencies = [
  "actix-http",
  "actix-router",
@@ -230,7 +230,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9dd80fa0bd6217e482112d9d87a05af8e0f8dec9e3aa51f34816f761c5cf7da7"
 dependencies = [
  "quote",
- "syn 2.0.101",
+ "syn 2.0.104",
 ]
 
 [[package]]
@@ -244,9 +244,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aead"
@@ -285,15 +285,15 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.11"
+version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
+checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
- "getrandom 0.2.16",
+ "getrandom 0.3.3",
  "once_cell",
  "version_check",
- "zerocopy 0.7.35",
+ "zerocopy",
 ]
 
 [[package]]
@@ -328,15 +328,16 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alloy-consensus"
-version = "1.0.6"
+version = "1.0.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b2123d190c823301be54e14a12ef10c00823d90047a140aa41650c26668d5b1"
+checksum = "1b6093bc69509849435a2d68237a2e9fea79d27390c8e62f1e4012c460aabad8"
 dependencies = [
  "alloy-eips",
  "alloy-primitives",
  "alloy-rlp",
  "alloy-serde",
  "alloy-trie",
+ "alloy-tx-macros",
  "auto_impl",
  "c-kzg",
  "derive_more",
@@ -344,7 +345,7 @@ dependencies = [
  "k256",
  "once_cell",
  "rand 0.8.5",
- "secp256k1 0.30.0",
+ "secp256k1",
  "serde",
  "serde_with",
  "thiserror 2.0.12",
@@ -352,9 +353,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus-any"
-version = "1.0.7"
+version = "1.0.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f31b286aeef04a32720c10defd21c3aa6c626154ac442b55f6d472caeb1c6741"
+checksum = "8d1cfed4fefd13b5620cb81cdb6ba397866ff0de514c1b24806e6e79cdff5570"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -402,9 +403,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-eips"
-version = "1.0.6"
+version = "1.0.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f7d2126e39e1557f0e661e0a5a36748a0bf280490ca2c0a1d149a55d2c2c8675"
+checksum = "5937e2d544e9b71000942d875cbc57965b32859a666ea543cc57aae5a06d602d"
 dependencies = [
  "alloy-eip2124",
  "alloy-eip2930",
@@ -417,14 +418,14 @@ dependencies = [
  "derive_more",
  "either",
  "serde",
- "sha2 0.10.8",
+ "sha2",
 ]
 
 [[package]]
 name = "alloy-json-abi"
-version = "1.1.0"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0068ae277f5ee3153a95eaea8ff10e188ed8ccde9b7f9926305415a2c0ab2442"
+checksum = "459f98c6843f208856f338bfb25e65325467f7aff35dfeb0484d0a76e059134b"
 dependencies = [
  "alloy-primitives",
  "alloy-sol-type-parser",
@@ -434,9 +435,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-network-primitives"
-version = "1.0.7"
+version = "1.0.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f05aa52713c376f797b3c7077708585f22a5c3053a7b1b2b355ea98edeb2052d"
+checksum = "793df1e3457573877fbde8872e4906638fde565ee2d3bd16d04aad17d43dbf0e"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -447,9 +448,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-primitives"
-version = "1.1.0"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a12fe11d0b8118e551c29e1a67ccb6d01cc07ef08086df30f07487146de6fa1"
+checksum = "3cfebde8c581a5d37b678d0a48a32decb51efd7a63a08ce2517ddec26db705c8"
 dependencies = [
  "alloy-rlp",
  "bytes",
@@ -457,14 +458,14 @@ dependencies = [
  "const-hex",
  "derive_more",
  "foldhash",
- "hashbrown 0.15.3",
- "indexmap 2.9.0",
+ "hashbrown 0.15.4",
+ "indexmap 2.10.0",
  "itoa",
  "k256",
  "keccak-asm",
  "paste",
  "proptest",
- "rand 0.9.1",
+ "rand 0.9.2",
  "ruint",
  "rustc-hash",
  "serde",
@@ -474,9 +475,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-rlp"
-version = "0.3.11"
+version = "0.3.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3d6c1d995bff8d011f7cd6c81820d51825e6e06d6db73914c1630ecf544d83d6"
+checksum = "5f70d83b765fdc080dbcd4f4db70d8d23fe4761f2f02ebfa9146b833900634b4"
 dependencies = [
  "alloy-rlp-derive",
  "arrayvec",
@@ -485,20 +486,20 @@ dependencies = [
 
 [[package]]
 name = "alloy-rlp-derive"
-version = "0.3.11"
+version = "0.3.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a40e1ef334153322fd878d07e86af7a529bcb86b2439525920a88eba87bcf943"
+checksum = "64b728d511962dda67c1bc7ea7c03736ec275ed2cf4c35d9585298ac9ccf3b73"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.101",
+ "syn 2.0.104",
 ]
 
 [[package]]
 name = "alloy-rpc-types-beacon"
-version = "1.0.9"
+version = "1.0.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "241aba7808bddc3ad1c6228e296a831f326f89118b1017012090709782a13334"
+checksum = "9196cbbf4b82a3cc0c471a8e68ccb30102170d930948ac940d2bceadc1b1346b"
 dependencies = [
  "alloy-eips",
  "alloy-primitives",
@@ -510,9 +511,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-rpc-types-engine"
-version = "1.0.9"
+version = "1.0.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cab52691970553d84879d777419fa7b6a2e92e9fe8641f9324cc071008c2f656"
+checksum = "f2f9cbf5f781b9ee39cfdddea078fdef6015424f4c8282ef0e5416d15ca352c4"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
@@ -527,9 +528,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-rpc-types-eth"
-version = "1.0.7"
+version = "1.0.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "89d9b4293dfd4721781d33ee40de060376932d4a55d421cf6618ad66ff97cc52"
+checksum = "46586ec3c278639fc0e129f0eb73dbfa3d57f683c44b2ff5e066fab7ba63fa1f"
 dependencies = [
  "alloy-consensus",
  "alloy-consensus-any",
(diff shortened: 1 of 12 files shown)


crates/networking/p2p/src/network/misc.rs
```diff
@@ -26,7 +27,7 @@ impl libp2p::swarm::Executor for Executor {
         self.0.spawn(f);
     }
 }
-
+#[allow(deprecated)]
```
syjn99, line 30: It seems those kind of changes is not needed.

Cargo.toml
```diff
@@ -82,7 +86,9 @@ ethereum_ssz_derive = "0.9"
 eventsource-client = "0.15.0"
 futures = "0.3"
 hashbrown = "0.15.3"
-hashsig = { git = "https://github.com/b-wagn/hash-sig" }
+hashsig = { git = "https://github.com/b-wagn/hash-sig.git" }
```
syjn99, line 89:
```suggestion
hashsig = { git = "https://github.com/b-wagn/hash-sig.git", rev = "287517a763edba7e518b0c1ee5beb868f26f1f66" }
```

How about lock the version, as it is under development actively?

Cargo.lock
syjn99, line 1: It seems like there are bunch of minor version updates which are not that related with this PR. I couldn't find any issues rn. But personally if we want to bump the dependency versions, I think this PR is not a right place to do so as it brings lots of dependency changes itself.

crates/common/consensus/lean/src/block.rs
```diff
@@ -10,7 +10,7 @@ use crate::vote::Vote;
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
 pub struct SignedBlock {
     pub data: Block,
-    pub signature: PQSignature,
+    // pub signature: PQSignature,
```
syjn99, line 13: actually `PQSignature` is a stub struct that should be fixed after we finalized how we put signature in `SignedBlock`. I would suggest not changing `consensus/lean` crate, and remain the stub struct as well.

crates/common/consensus/lean/src/validator.rs
syjn99, line 1: This file isn't necessary anymore, seems like you have been working on the outdated codebase. So please remove it.

syjn99 (changes requested):
Few cosmetics reviews. I will dive deeper into the actual implementation. I think I need to spend a couple of hours of `hash-sig` library to fully understand what this PR deals with.

bin/ream/src/main.rs
```diff
@@ -345,7 +345,10 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
     );
 
     let seed_phrase = config.get_seed_phrase();
-    ream_account_manager::generate_keys(&seed_phrase);
+    let _ = ream_account_manager::generate_and_save_keys(
+        config.lifetime.try_into().unwrap(),
```
syjn99, line 349: How about changing the type of `lifetime` of `AccountManagerConfig` into `u32`, so that we don't have to `into` and `unwrap` it?

crates/account_manager/Cargo.toml
```diff
@@ -10,8 +10,17 @@ rust-version.workspace = true
...
+hex.workspace = true
+hmac.workspace = true
 rand.workspace = true
-rand_chacha.workspace = true
+ream-pqc = { path = "../crypto/pqc" }
+serde = { version = "1.0", features = ["derive"] }
+serde_json.workspace = true
 sha2.workspace = true
+thiserror = "2.0"
 tracing.workspace = true
```
syjn99, line 26:
```toml
[dependencies]
aes-gcm.workspace = true
anyhow.workspace = true
argon2.workspace = true
base64.workspace = true
hashsig.workspace = true
hex.workspace = true
hmac.workspace = true
rand.workspace = true
serde.workspace = true
serde_json.workspace = true
sha2.workspace = true
thiserror.workspace = true
tracing.workspace = true

# ream dependencies
ream-pqc.workspace = true
```

I think above code is more consistent with our current codebase.

crates/crypto/pqc/src/keystore.rs
syjn99, line 1: It seems the filename `keystore.rs` isn't intuitive, as it mostly handles the signature scheme itself.

crates/crypto/pqc/src/keystore.rs
```diff
@@ -0,0 +1,426 @@
...
+    symmetric::{
+        message_hash::poseidon::PoseidonMessageHash, prf::shake_to_field::ShakePRFtoF,
+        tweak_hash::poseidon::PoseidonTweakHash,
+    },
+};
+use rand::Rng;
+use serde::{Deserialize, Serialize};
+
+// Custom signature scheme with lifetime 2^8
+mod custom_lifetime_8 {
```
syjn99, line 21: I guess this is introduced due to the testing efficiency, right?

crates/crypto/pqc/src/keystore.rs
```diff
@@ -0,0 +1,426 @@
...
+}
+
+impl Error for XmssWrapperError {}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct KeyPair {
+    pub public_key: Vec<u8>,
+    pub secret_key: Vec<u8>,
+    pub lifetime: u32,
+}
```
syjn99, line 91: What's the benefit to define as a byte array for both `public_key` and `secret_key`, provided `GeneralizedXMSSPublicKey` and `GeneralizedXMSSSecretKey` are already serializable? Same comment for `XmssSignature.signature`.

syjn99: We don't have multiple backends for the hash-based signature (at this point), so I don't think this is justified.

crates/crypto/pqc/src/keystore.rs
```diff
@@ -0,0 +1,426 @@
...
+
+#[derive(Serialize, Deserialize)]
+pub struct XmssSignature {
+    pub signature: Vec<u8>,
+    pub tree_height: u32,
+    pub signature_index: u32, // TODO: Implement proper signature index tracking
+}
+
+pub struct XmssWrapper;
```
syjn99, line 100: I think this struct can be improved much better. Here's what I experimented with my local environment:

---

Matching by `height` isn't scalable, and also can't handle different signature schemes. There are so many schemes in `hash-sig` with different 1) incomparable encodings (e.g., Winternitz, Target Sum, Top-level), 2) chunk size and 3) lifetime.

Fortunately, I found `SignatureScheme` trait in `hash-sig` crate, which will make our lives easier. Let me call the new struct `XmssWrapper2`:

```rust
pub struct XmssWrapper2<T: SignatureScheme> {
    _phantom: std::marker::PhantomData<T>,
}
```

Then we can implement `generate_keys` with this trait, and note that if we don't use a byte array for `PublicKey`, most of the code below can also be removed. (Please refer to the previous comment that I left)

```rust
impl<T> XmssWrapper2<T>
where
    T: SignatureScheme,
    T::PublicKey: Serialize + for<'de> Deserialize<'de>,
    T::SecretKey: Serialize + for<'de> Deserialize<'de>,
    T::Signature: Serialize + for<'de> Deserialize<'de>,
{
    /// Generate XMSS key pair (serialized)
    pub fn generate_keys<R: Rng>(rng: &mut R) -> Result<KeyPair, XmssWrapperError> {
        let (public_key, secret_key) = T::key_gen(rng, 0, T::LIFETIME as usize);

        let public_key_bytes = serde_json::to_vec(&public_key).map_err(|e| {
            XmssWrapperError::Serialization(format!("Failed to serialize public key: {:?}", e))
        })?;
        let secret_key_bytes = serde_json::to_vec(&secret_key).map_err(|e| {
            XmssWrapperError::Serialization(format!("Failed to serialize secret key: {:?}", e))
        })?;

        Ok(KeyPair {
            public_key: public_key_bytes,
            secret_key: secret_key_bytes,
            lifetime: (T::LIFETIME as f64).log2() as u32,
        })
    }
}
```

And this can be tested like:

```rust
    #[test]
    fn test_key_generation2() {
        let mut rng = rng();

        type XmssWrapper8 = XmssWrapper2<custom_lifetime_8::CustomSIGWinternitzLifetime8W8>;

        let result = XmssWrapper8::generate_keys(&mut rng);
        assert!(result.is_ok());

        let key_pair = result.unwrap();
        assert_eq!(key_pair.lifetime, 8);
        assert!(!key_pair.public_key.is_empty());
        assert!(!key_pair.secret_key.is_empty());
    }
```

From this approach, I expect we can shrink the LOC of this file upto 40%.

crates/crypto/pqc/src/keystore.rs
```diff
@@ -0,0 +1,426 @@
...
+                    <SIGWinternitzLifetime18W8 as SignatureScheme>::SecretKey,
+                >(&secret_key_bytes_vec)
+                .map_err(|e| {
+                    XmssWrapperError::Signing(format!("Failed to reconstruct secret key: {:?}", e))
+                })?;
+
+                // Convert message to 32-byte array as required by SignatureScheme
+                let mut message_array = [0u8; 32];
+                let message_len = message.len().min(32);
+                message_array[..message_len].copy_from_slice(&message[..message_len]);
```
syjn99, line 175: We shouldn't sign only **part of the given message**. Is this code intended to get **hash value** of the message, as an argument?

crates/crypto/pqc/src/keystore.rs
```diff
@@ -0,0 +1,426 @@
...
+
+                #[allow(clippy::unnecessary_mut_passed)]
+                let signature =
+                    SIGWinternitzLifetime18W8::sign(rng, &mut secret_key, epoch, &message_array)
```
syjn99, line 179:
```suggestion
                    SIGWinternitzLifetime18W8::sign(rng, &secret_key, epoch, &message_array)
```

It is enough to pass only reference, not mutable reference. It also means that `secret_key_bytes` is never updated.

crates/crypto/pqc/src/keystore.rs
```diff
@@ -0,0 +1,426 @@
...
+
+        let result = XmssWrapper::generate_keys(8, &mut rng);
+        assert!(result.is_ok());
+
+        let key_pair = result.unwrap();
+        assert_eq!(key_pair.lifetime, 8);
+        assert!(!key_pair.public_key.is_empty());
+        assert!(!key_pair.secret_key.is_empty());
+
+        // Test with tree height 8 (256 signatures)
```
syjn99, line 374: Why do we have copy-pasted assertion for each tests?

crates/crypto/pqc/src/keystore.rs
syjn99: Also in my knowledge, we won't use Winternitz as our primary incomparable encoding scheme. 

Reference: https://eprint.iacr.org/2025/1332.pdf

crates/crypto/pqc/src/keystore.rs
syjn99: `message: &[u8; MESSAGE_LENGTH]`

^ This can be much better for parameter definition.

crates/common/consensus/lean/src/block.rs
ch4r10t33r: I had to comment it because I ran into issues while testing. I no longer see the issue, I'll revert my changes.

Cargo.toml
```diff
@@ -106,7 +113,7 @@ sha2 = "0.10"
 snap = "1.1"
 ssz_types = { git = "https://github.com/ReamLabs/ssz_types", branch = "removable-variable-list" }
 tempdir = "0.3.7"
-tempfile = "3.19"
+tempfile = "=3.19"
```
unnawut, line 116: just for my own learning, what required this pinning?

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+    let mut rng = rng();
+
+    // Generate keys using the pqc crate
+    let key_pair = XmssWrapper::generate_keys(height, &mut rng)
+        .map_err(|e| AccountManagerError::Serialization(e.to_string()))?;
+
+    // Save keys
+    save_secret_key(&key_pair, password)?;
+
+    println!("Poseidon2-XMSS key pair generated and saved successfully!");
```
unnawut, line 64: use [tracing](https://docs.rs/tracing/latest/tracing/)'s `info!` here

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+        max_signatures: 1u64 << key_pair.lifetime,
+        key_type: "XMSS_SECRET_ENCRYPTED".to_string(),
+        hash_function: "Poseidon2".to_string(),
+    };
+
+    // Save encrypted secret key to file
+    let encrypted_json = serde_json::to_string_pretty(&encrypted_secret)
+        .map_err(|e| AccountManagerError::Serialization(e.to_string()))?;
+
+    let mut secret_file = File::create("xmss_secret_key.json")?;
```
unnawut, line 153: Can we pass in the path as a function argument rather than hardcoding the file name here? Also applies to public key

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+        .map_err(|e| AccountManagerError::Serialization(e.to_string()))?;
+
+    let mut public_file = File::create("xmss_public_key.json")?;
+    public_file.write_all(public_json_string.as_bytes())?;
+
+    // Encrypt and save secret key file
+    save_encrypted_secret_key(key_pair, password)?;
+
+    Ok(())
+}
```
unnawut, line 98: Instead of having `save_secret_key()` that saves the public key and calls `save_encrypted_secret_key()`, can we instead put the public key saving into `save_public_key()`.

Then call `save_public_key()` and `save_encrypted_secret_key()` from `generate_and_save_keys()`?

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+    })?;
+
+    hex::decode(hex_key).map_err(|e| AccountManagerError::Serialization(e.to_string()))
+}
+
+/// Sign a message using stored secret key
+pub fn sign_message(
+    message: &[u8],
+    secret_key_path: &Path,
+    password: &str,
```
unnawut, line 297: I think it'll be quite impractical to recover the secret key from file each time we want to sign a message.

Can we wrap this into, say, an `AccountManager` struct that can load a secret key from file and keep the decrypted key in memory in the struct? then we call `account_manager.sign_message(...)` to sign a message?

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+
+    // Load the secret key
+    let secret_key_bytes = load_secret_key(secret_key_path, password)?;
+
+    // Get tree height from the secret key file metadata
+    let tree_height = get_tree_height_from_file(secret_key_path)?;
+
+    // Sign the message
+    let (signature, updated_secret_key_bytes) =
+        XmssWrapper::sign_message(message, &secret_key_bytes, tree_height, &mut rng, 0)
```
unnawut, line 311: Umm don't we need to keep the epoch argument (i.e. fixed to `0` here) inputtable?

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+            .map_err(|e| AccountManagerError::Serialization(e.to_string()))?;
+
+    // Update the secret key file with the new state
+    let updated_keypair = KeyPair {
```
unnawut, line 315: Does the keypair change after each signing? I thought the secret_key and lifetime doesn't change per signing.

Also by setting `public_key: vec![]` this is a destructive operation that removes public key from the keystore file.

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+    signature: &XmssSignature,
+    public_key_path: &Path,
+) -> Result<bool, AccountManagerError> {
+    use ream_pqc::keystore::XmssWrapper;
+
+    // Load the public key
+    let public_key_bytes = load_public_key(public_key_path)?;
+
+    // Verify the signature
+    let is_valid = XmssWrapper::verify_signature(message, signature, &public_key_bytes, 0)
```
unnawut, line 338: Similar comment to signing, doesn't the epoch need to be an input?

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+/// Helper function to get tree height from secret key file
+fn get_tree_height_from_file<P: AsRef<Path>>(path: P) -> Result<u32, AccountManagerError> {
+    let mut file = File::open(path)?;
+    let mut content = String::new();
+    file.read_to_string(&mut content)?;
+
+    let encrypted_secret: EncryptedSecretKey = serde_json::from_str(&content)
+        .map_err(|e| AccountManagerError::Serialization(e.to_string()))?;
+
+    Ok(encrypted_secret.tree_height)
```
unnawut, line 353: Do we need to load the file separately to get the tree height? Can we do this as part of `load_secret_key()`?

Cargo.toml
ch4r10t33r: The latest version of tempfile has deprecated several functions, which is causing our `make pr` command to fail.

crates/account_manager/src/lib.rs
ch4r10t33r: Yes, this entire crate has to be refactored. I'll be addressing all of them in the new PR. The recent discussions in post-quantum group will require further changes in our code base.

crates/account_manager/src/lib.rs
ch4r10t33r: Yes, I had them as separate functions for me to test out the serialisation. I'll be moving this over to `generate_and_save_keys()` with a pre-agreed path.

crates/account_manager/src/lib.rs
ch4r10t33r: Please ignore the acocunt_manager, this code is incomplete.

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+    #[error("Encryption error: {0}")]
+    Encryption(String),
+    #[error("Decryption error: {0}")]
+    Decryption(String),
+    #[error("Authentication error: {0}")]
+    Authentication(String),
+}
+
+/// Generate and save XMSS key pair with encryption
+pub fn generate_and_save_keys(height: u32, password: &str) -> Result<(), AccountManagerError> {
```
unnawut, line 52: Super nitpick and super unsure. I think we use anyhow::Result everywhere else instead of import it for disambiguity.

@KolbyML @syjn99 Can you help confirm?

syjn99: @unnawut Yes, that would be better in terms of consistency

crates/account_manager/Cargo.toml
```diff
@@ -10,8 +10,17 @@ rust-version.workspace = true
...
+aes-gcm.workspace = true
+anyhow.workspace = true
+argon2.workspace = true
+base64.workspace = true
 hashsig.workspace = true
+hex.workspace = true
+hmac.workspace = true
 rand.workspace = true
-rand_chacha.workspace = true
+ream-pqc = { path = "../crypto/pqc" }
```
KolbyML, line 21: defining path's should only be defined in the main Cargo.toml

bin/ream/src/cli/account_manager.rs
```diff
@@ -58,8 +57,10 @@ impl AccountManagerConfig {
         if let Some(phrase) = &self.seed_phrase {
             phrase.clone()
         } else {
-            let mnemonic = Mnemonic::random(OsRng, Default::default());
-            let phrase = mnemonic.phrase().to_string();
+            // Generate a random 12-word mnemonic using the correct bip39 v2.0 API
+            let entropy = rand::random::<[u8; 16]>(); // 128 bits for 12 words
+            let mnemonic = Mnemonic::from_entropy_in(bip39::Language::English, &entropy).unwrap();
```
KolbyML, line 62: unwrap shouldn't be used in production code, can this fail? under what conditions? depending on the case we should pass up an error or use expect instead of unwrap, it is probably safer to just pass up an error

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+    use ream_pqc::keystore::XmssWrapper;
+
+    let mut rng = rng();
+
+    // Generate keys using the pqc crate
+    let key_pair = XmssWrapper::generate_keys(height, &mut rng)
+        .map_err(|e| AccountManagerError::Serialization(e.to_string()))?;
```
KolbyML, line 59:
```suggestion
        .map_err(|err| AccountManagerError::Serialization(err.to_string()))?;
```}

please make this change in the 22 other spots in the PR

File: crates/account_manager/src/lib.rs (hunk 21 above)
Comment (KolbyML) [reply crates/account_manager/src/lib.rs:52]: He isn't using anyhow here.

We can do 2 things for errors
- anyhow
- this_error

context decides which we opt for

File: crates/account_manager/src/lib.rs [hunk 25]
```diff
@@ -1,25 +1,354 @@
...
+    // Save keys
+    save_secret_key(&key_pair, password)?;
+
+    println!("Poseidon2-XMSS key pair generated and saved successfully!");
+    println!("Public key saved to: xmss_public_key.json");
+    println!("Secret key saved to: xmss_secret_key.json (encrypted)");
+    println!(
+        "Tree height: {} (2^{} = {} signatures)",
+        height,
+        height,
```
KolbyML, line 70: these can be inlined, and as @unnawut state println!() shouldn't be used in production code

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+/// Derive encryption key from password using Argon2id
+fn derive_key_from_password(password: &str, salt: &[u8]) -> Result<[u8; 32], AccountManagerError> {
+    // Configure Argon2id with strong parameters
+    let params = Params::new(
+        19456,    // memory cost (19 MiB)
+        2,        // time cost (iterations)
+        1,        // parallelism
+        Some(32), // output length
+    )
+    .map_err(|e| AccountManagerError::Encryption(format!("Argon2 params error: {}", e)))?;
```
KolbyML, line 168: err and inline

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+
+    let argon2 = Argon2::new(Algorithm::Argon2id, Version::V0x13, params);
+
+    let mut key = [0u8; 32];
+    argon2
+        .hash_password_into(password.as_bytes(), salt, &mut key)
+        .map_err(|e| {
+            AccountManagerError::Encryption(format!("Argon2 key derivation failed: {}", e))
```
KolbyML, line 176: err and inline, apply this comment to anywhere relavent

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+) -> Result<(), AccountManagerError> {
+    let mut mac = <HmacSha256 as Mac>::new_from_slice(key)
+        .map_err(|e| AccountManagerError::Authentication(format!("HMAC key error: {}", e)))?;
+
+    mac.update(encrypted_data);
+    mac.update(nonce);
+    mac.update(salt);
+
+    mac.verify_slice(expected_hmac)
+        .map_err(|_| AccountManagerError::Authentication("HMAC verification failed".to_string()))?;
```
KolbyML, line 215: we shouldn't be hiding this error, instead we should pass it up

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+    let nonce = Nonce::from_slice(&nonce_bytes);
+    let cipher = Aes256Gcm::new(Key::<Aes256Gcm>::from_slice(&derived_key));
+    let decrypted_data = cipher
+        .decrypt(nonce, encrypted_data.as_ref())
+        .map_err(|e| {
+            AccountManagerError::Decryption(format!("AES-GCM decryption failed: {}", e))
+        })?;
+
+    // Parse decrypted JSON
+    let decrypted_json: serde_json::Value = serde_json::from_slice(&decrypted_data)
```
KolbyML, line 283:
```suggestion
    let decrypted_json: serde_json::Value = serde_json::from_slice(&decrypted_data)
```

can't we just import Value? instead of qualifying the path

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+    Ok(signature)
+}
+
+/// Verify a signature against a message using stored public key
+pub fn verify_signature(
+    message: &[u8],
+    signature: &XmssSignature,
+    public_key_path: &Path,
+) -> Result<bool, AccountManagerError> {
+    use ream_pqc::keystore::XmssWrapper;
```
KolbyML, line 332: imports should be at the top of the file, unless it is a test, which in that case the import just be at the top of the mod

crates/common/consensus/lean/src/block.rs
KolbyML: use `pub _signature:PQSignature` then or remove it, but commenting this out doesn't make sense

crates/common/consensus/lean/src/validator.rs
KolbyML: This comment was marked as resolved, but the problem wasn't resolved

crates/crypto/pqc/src/keystore.rs
```diff
@@ -0,0 +1,426 @@
...
+        message: &[u8],
+        signature: &XmssSignature,
+        public_key_bytes: &[u8],
+        epoch: u32,
+    ) -> Result<bool, XmssWrapperError> {
+        // Verify based on tree height
+        let result: Result<bool, XmssWrapperError> = match signature.tree_height {
+            18 => {
+                // Use the Winternitz encoding with chunk size w = 1 for height 18
+                use instantiations_poseidon::lifetime_2_to_the_18::winternitz::SIGWinternitzLifetime18W8;
```
KolbyML, line 274: this shouldn't be imported here

crates/networking/p2p/src/network/misc.rs
```diff
@@ -15,6 +15,7 @@ use libp2p::{
     yamux,
 };
 use libp2p_identity::{Keypair, PeerId, secp256k1::PublicKey as Secp256k1PublicKey};
+#[allow(deprecated)]
```
KolbyML, line 18: We shouldn't be using `#[allow(deprecated)]`

Either we update the crate in a separate PR or fix it here in this PR

crates/account_manager/src/lib.rs
```diff
@@ -1,25 +1,354 @@
...
+    hex::decode(hex_key).map_err(|e| AccountManagerError::Serialization(e.to_string()))
+}
+
+/// Sign a message using stored secret key
+pub fn sign_message(
+    message: &[u8],
+    secret_key_path: &Path,
+    password: &str,
+) -> Result<XmssSignature, AccountManagerError> {
+    use ream_pqc::keystore::XmssWrapper;
```
KolbyML, line 299: import etc

crates/account_manager/src/lib.rs
KolbyML: Overall this code feels unstructured. I agree most of these functions should be in structures organized by what makes sense.

crates/crypto/pqc/src/keystore.rs
```diff
@@ -0,0 +1,426 @@
...
+    type PRFw8 = ShakePRFtoF<HASH_LEN_FE>;
+    type IEw8 = WinternitzEncoding<MHw8, CHUNK_SIZE_W8, NUM_CHUNKS_CHECKSUM_W8>;
+
+    /// Custom instantiation with Lifetime 2^8, Winternitz encoding, chunk size w = 8
+    pub type CustomSIGWinternitzLifetime8W8 =
+        GeneralizedXMSSSignatureScheme<PRFw8, IEw8, THw8, LOG_LIFETIME>;
+}
+
+#[allow(unused_imports)]
+use custom_lifetime_8::CustomSIGWinternitzLifetime8W8;
```
KolbyML, line 61: if this is unused, why is it here?

crates/crypto/pqc/src/keystore.rs
```diff
@@ -0,0 +1,426 @@
...
+        }
+    }
+}
+
+impl Error for XmssWrapperError {}
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct KeyPair {
+    pub public_key: Vec<u8>,
+    pub secret_key: Vec<u8>,
```
KolbyML, line 89: aren't these defined lengths? we shouldn't use vector's if the lengths are defined (are they?)

ch4r10t33r:
The work will be split into multiple tasks which will be shared by @shariqnaiyer and myself. Hence, closing this pull request.

Issue #654, opened 2025-07-23
PQ Devnet-0 Meta Tracking Issue


syjn99:
### Describe the feature

Now we're on the track shipping our very first PQ devnet ([`pq-devnet-0`](https://github.com/leanEthereum/leanSpec/tree/main/docs/client)). I'd like to open this issue for task management regarding PQ devnet, and feedbacks are welcomed!

# Overview

#679 introduces a basic working network with a single local node. ([Architecture](https://hackmd.io/@reamlabs/B1V8VRxdxl)).

## Networking

- [x] MUST be able to connect peers given by a configuration file (static peers)
  - Related issue: #682 
- [x] MUST be able to gossip.
  - Related issue: #671 

## Validator

- [x] MUST determine its duty for current slot.
- [x] MUST be able to sign a empty block and propose the signed block to p2p.
- [x] MUST be able to attest (=vote).
- NICE TO HAVE ability for fresh PQC key generation.

## API

Related issue: #684 

## Observability

Related issue: #687 

## Cryptography

Related issue #655 and PR #732

PR #672, opened 2025-07-31
3SF for lean pqdevnet

unnawut:
### To-Do

 

crates/common/consensus/lean/src/vote.rs
```diff
@@ -0,0 +1,32 @@
...
+pub struct VoteData {
+    pub validator_id: usize,
+    pub slot: usize,
+    pub head: Hash,
+    pub head_slot: usize,
+    pub target: Hash,
+    pub target_slot: usize,
+    pub source: Hash,
+    pub source_slot: usize,
+}
```
syjn99:
```suggestion
#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
pub struct SignedVote {
    pub data: Vote,
    pub signature: PQSignature,
}

#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
pub struct Vote {
    pub validator_id: usize,
    pub slot: usize,
    pub head: Hash,
    pub head_slot: usize,
    pub target: Hash,
    pub target_slot: usize,
    pub source: Hash,
    pub source_slot: usize,
}
```

I think it'll be better to match struct name with the 3sf-mini specification. `SignedVote` will play similarly with `SignedBlock`.

crates/common/consensus/lean/src/block.rs
```diff
@@ -1,22 +1,36 @@
-use alloy_primitives::B256;
+use ethereum_hashing::hash;
 use ream_pqc::PQSignature;
 use serde::{Deserialize, Serialize};
 use ssz_derive::{Decode, Encode};
-use tree_hash_derive::TreeHash;
+use ssz_types::{
+    VariableList,
+    typenum::{
+        U16777216, // 2**24
```
syjn99: Is there any reference to choose the limit as `2**24`? Also some fields in `LeanState` has `2**12`  (sorry I couldn't find it)

crates/common/consensus/lean/src/block.rs
```diff
@@ -1,22 +1,36 @@
...
-#[derive(
-    Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash, Default,
-)]
+// TODO: Add back #[derive(TreeHash)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, Default)]
 pub struct Block {
-    pub slot: u64,
-    pub proposer_index: u64,
+    pub slot: usize,
+    pub proposer_index: usize,
```
syjn99: I think `proposer_index` should be deleted, as it is not included in the [Python code](https://github.com/ethereum/research/blob/d225a6775a9b184b5c1fd6c830cc58a375d9535f/3sf-mini/consensus.py#L38-L44).

crates/common/consensus/lean/src/state.rs
```diff
@@ -1,14 +1,37 @@
...
+    pub latest_finalized_slot: usize,
 
-#[derive(Debug, PartialEq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
-pub struct BeamState {
-    pub genesis_time: u64,
+    pub historical_block_hashes: VariableList<Option<Hash>, U4096>,
+    pub justified_slots: VariableList<bool, U4096>,
+
+    pub justifications: HashMap<Hash, Vec<bool>>,
+}
```
syjn99: Similarly, I think it'll be better to be aligned with Python code. (e.g., adding `LeanConfig` container)

Regarding `stakers` field, I guess the validator set of the devnet won't change, so each node can identify the signer of each block.

crates/common/consensus/lean/src/lib.rs
```diff
@@ -1,4 +1,213 @@
...
+    || (delta as f64).sqrt().fract() == 0.0 // any x^2
+    || (delta as f64 + 0.25).sqrt() % 1.0 == 0.5 // any x^2+x
+}
+
+// Given a state, output the new state after processing that block
+pub fn process_block(pre_state: &LeanState, block: &Block) -> LeanState {
+    let mut state = pre_state.clone();
+
+    // Track historical blocks in the state
+    // TODO: proper error handlings
```
syjn99: Yeah I guess this is the place where `anyhow` comes into play, but let's leave it as is.

crates/common/consensus/lean/src/vote.rs
syjn99: Upon that, I think it would be fine to let block has the list of `Vote` (without signature).

crates/common/consensus/lean/src/lib.rs
```diff
@@ -1,4 +1,213 @@
...
+
+        // Track attempts to justify new hashes
+        if !state.justifications.contains_key(&vote.data.target) {
+            let mut empty_justifications = Vec::<bool>::with_capacity(state.num_validators);
+            empty_justifications.resize(state.num_validators, false);
+
+            state
+                .justifications
+                .insert(vote.data.target, empty_justifications);
+        }
```
syjn99:
```suggestion
        // Track attempts to justify new hashes
        if !state.justifications.contains_key(&vote.data.target) {
            state
                .justifications
                .insert(vote.data.target, vec![false; state.num_validators]);
        }
```

This can be simplified.

crates/common/consensus/lean/src/lib.rs
```diff
@@ -1,4 +1,213 @@
...
+            // Finalization: if the target is the next valid justifiable
+            // hash after the source
+            let mut is_target_next_valid_justifiable_slot = true;
+
+            for slot in (vote.data.source_slot + 1)..vote.data.target_slot {
+                if is_justifiable_slot(&state.latest_finalized_slot, &slot) {
+                    is_target_next_valid_justifiable_slot = false;
+                    break;
+                }
+            }
```
syjn99:
```suggestion
            let is_target_next_valid_justifiable_slot = !((vote.data.source_slot + 1)
                ..vote.data.target_slot)
                .any(|slot| is_justifiable_slot(&state.latest_finalized_slot, &slot));
```

Just my personal preference!

crates/common/consensus/lean/src/lib.rs
```diff
@@ -1,4 +1,213 @@
...
+        root = blocks
+            .iter()
+            .min_by_key(|(_, block)| block.slot)
+            .map(|(hash, _)| *hash)
+            .unwrap();
+    }
+
+    // Sort votes by ascending slots to ensure that new votes are inserted last
+    let mut sorted_votes = votes.clone();
+    sorted_votes.sort_by_key(|vote| vote.data.slot);
```
syjn99: Seems this `sorted_votes` is not used.

crates/common/consensus/lean/src/lib.rs
```diff
@@ -1,4 +1,213 @@
...
+                state.latest_finalized_slot = vote.data.source_slot;
+            }
+        }
+    }
+
+    state
+}
+
+// Get the highest-slot justified block that we know about
+pub fn get_latest_justified_hash(post_states: &HashMap<Hash, LeanState>) -> Option<Hash> {
```
syjn99: I think this would return `Err` when `post_states.values().len() == 0`. We might use `.unwrap()` here right now, and use `anyhow` in later PR.

crates/common/consensus/lean/src/lib.rs
```diff
@@ -1,4 +1,213 @@
...
+
+    for (hash, block) in blocks {
+        if block.parent.is_some() && *vote_weights.get(hash).unwrap_or(&0) >= min_score {
+            match children_map.get_mut(&block.parent.unwrap()) {
+                Some(child_hashes) => {
+                    child_hashes.push(*hash);
+                }
+                None => {
+                    children_map.insert(block.parent.unwrap(), vec![*hash]);
+                }
```
syjn99:
```rust
            children_map
                .entry(block.parent.unwrap())
                .or_insert_with(Vec::new)
                .push(*hash);
```
we can replace like this

crates/common/consensus/lean/src/staker.rs
```diff
@@ -0,0 +1,322 @@
...
+            safe_target: genesis_hash,
+            head: genesis_hash,
+        }
+    }
+
+    pub fn latest_justified_hash(&self) -> Option<Hash> {
+        get_latest_justified_hash(&self.post_states)
+    }
+
+    pub fn latest_finalized_hash(&self) -> Option<Hash> {
```
syjn99: ~Same as `latest_justified_hash`, I think this could return `Hash` by using `unwrap()` at this moment. We can handle error later on~ -> I think `Option<Hash>` is fine. Resolved this comment

crates/common/consensus/lean/src/staker.rs
```diff
@@ -0,0 +1,322 @@
...
+    Hash, QueueItem, SLOT_DURATION,
+    block::Block,
+    get_fork_choice_head, get_latest_justified_hash, is_justifiable_slot, process_block,
+    state::LeanState,
+    vote::{Vote, VoteData},
+};
+
+// TODO: Add back #[derive(TreeHash)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub struct Staker {
```
syjn99: It seems that `Staker` struct doesn't have to be SSZ compatible, unless it is included in `LeanState` or propagated through P2P. If so, we don't need to implement `TreeHash` trait here.

Also, I think implementing `Staker` and deciding which fields to have is up to each client. In our case, the functionality that `Staker` has can be divided into `NetworkService` and `ValidatorService`. What do you think of this point?

crates/common/consensus/lean/src/block.rs
unnawut: Good point! No there isn't any reference, this is just my arbitrary picking for now. I'll bring these numbers up in the pqinterop discussion.

crates/common/consensus/lean/src/lib.rs
unnawut: Nice use of range and any. Thanks!

crates/common/consensus/lean/src/lib.rs
unnawut: Good point!

crates/common/consensus/lean/src/staker.rs
unnawut: Totally agree! I was thinking the same and already working on this on [another branch](https://github.com/unnawut/ream/commit/0ff7a54fd6beea317ca4a5d6e297e95c73987132) for the next PR (it's not optimal yet).

The only small difference from your suggestion is `stakers` is part of `State` so we'll need `Staker` serializable and standardized. 

But yes I think `Staker` can be really small and contains only `staker_index` and `public_key` which is serializable. The rest of the fields can go into client-dependent service like you mentioned.

crates/common/consensus/lean/src/lib.rs
unnawut: I'll criminally update all to use `unwrap()` for now, makes more sense than ignoring function returns.

Then anyhow in the next PR.

unnawut:
Converting back to draft to address Jun's comments when I wake up.

crates/common/consensus/lean/src/staker.rs
unnawut: Btw. I prefer to do this splitting in another PR because this will be diverging from the [3SF-mini ref impl](https://github.com/ethereum/research/tree/master/3sf-mini) if that's okay with you?

syjn99: > The only small difference from your suggestion is stakers is part of State so we'll need Staker serializable and standardized.

Actually I think we do not have to contain the list of `stakers` in the `State`. ([comment above](https://github.com/ReamLabs/ream/pull/672#discussion_r2245065216)) The network would have a fixed set of validators which means it isn't part of the state. Maybe we provide a YAML file of validators when we run `lean_node`.

crates/common/consensus/lean/src/block.rs
unnawut: I'm now tracking these items in #654 under "To be agreed in interop"

crates/common/consensus/lean/src/lib.rs
KolbyML: use expect() at least

crates/common/consensus/lean/src/state.rs
```diff
@@ -1,14 +1,37 @@
...
 
-    /// Up to 1 million validators
-    pub validators: VariableList<Validator, ssz_types::typenum::U1000000>,
+impl LeanState {
+    pub fn compute_hash(&self) -> Hash {
```
KolbyML:
```suggestion
    pub fn compute_hash(&self) -> B256 {
```

can we use B256 directly instead of a type alias

crates/common/consensus/lean/src/state.rs
```diff
@@ -1,14 +1,37 @@
...
+};
+
+use crate::{Hash, staker::Staker};
+
+// TODO: Add back #[derive(Encode, Decode, TreeHash)]
+#[derive(Debug, Eq, PartialEq, Clone, Serialize, Deserialize)]
+pub struct LeanState {
+    pub genesis_time: usize,
+    pub stakers: VariableList<Staker, U4096>,
+    pub num_validators: usize,
```
KolbyML: all these usize's should be u64's

crates/common/consensus/lean/src/staker.rs
```diff
@@ -0,0 +1,322 @@
...
+    Hash, QueueItem, SLOT_DURATION,
+    block::Block,
+    get_fork_choice_head, get_latest_justified_hash, is_justifiable_slot, process_block,
+    state::LeanState,
+    vote::{SignedVote, Vote},
+};
+
+// TODO: Split to Staker and Service
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub struct Staker {
```
KolbyML: all these usize's should be u64's as well

crates/common/consensus/lean/src/lib.rs
KolbyML: I mean it shouldn't take much time to just use anyhow, it would probably take the same amount of time as just changing the unwrap's to expect

crates/common/consensus/lean/src/lib.rs
```diff
@@ -1,4 +1,201 @@
...
+use ssz_types::{
+    VariableList,
+    typenum::{
+        U16777216, // 2**24
+    },
+};
+
+use crate::{block::Block, state::LeanState, vote::SignedVote, vote::Vote};
+
+pub type Hash = B256;
```
KolbyML:
```suggestion
```

I don't think a type alias like this is justified. Hash's don't need to be B256's we are just using 256 bit cryptography. But in general use should just use a B256 directly here

crates/common/consensus/lean/src/lib.rs
```diff
@@ -1,4 +1,201 @@
...
+pub mod vote;
+
+use std::collections::HashMap;
+
+use alloy_primitives::B256;
+use serde::{Deserialize, Serialize};
+use ssz_types::{
+    VariableList,
+    typenum::{
+        U16777216, // 2**24
```
KolbyML: A little weird to have a comment for this

crates/common/consensus/lean/src/block.rs
```diff
@@ -1,22 +1,36 @@
...
     pub message: Block,
     pub signature: PQSignature,
 }
 
-#[derive(
-    Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash, Default,
-)]
+// TODO: Add back #[derive(TreeHash)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, Default)]
 pub struct Block {
```
KolbyML, line 19: why was u64 changed to usize? also we should use B256 instead of the type alias

KolbyML:
If you are using usize for zkVM's as it would be a u32, we still shouldn't use usize we should use u32 and define it in the leanSpecs

crates/common/consensus/lean/src/vote.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl Vote {
+    pub fn compute_hash(&self) -> Hash {
```
KolbyML: same comments on the alias and usize

crates/common/consensus/lean/src/block.rs
syjn99: Yeah that's nice, thanks

crates/common/consensus/lean/src/staker.rs
```diff
@@ -0,0 +1,316 @@
...
+            post_states,
+            known_votes: VariableList::<Vote, U4096>::empty(),
+            new_votes: VariableList::<Vote, U4096>::empty(),
+            dependencies: HashMap::<B256, Vec<QueueItem>>::new(),
+            genesis_hash,
+            num_validators: genesis_state.config.num_validators,
+            safe_target: genesis_hash,
+            head: genesis_hash,
+        }
+    }
```
KolbyML:
```suggestion
impl Staker {
    pub fn new(validator_id: u64, genesis_block: Block, genesis_state: LeanState) -> Staker {
        let genesis_hash = genesis_block.compute_hash();

        Staker {
            validator_id,
            public_key: PublicKey {},
            time: 0,
            known_votes: VariableList::empty(),
            new_votes: VariableList::empty(),
            dependencies: HashMap::new(),
            genesis_hash,
            num_validators: genesis_state.config.num_validators,
            safe_target: genesis_hash,
            head: genesis_hash,
            chain: HashMap::from([(genesis_hash, genesis_block)]),
            post_states: HashMap::from([(genesis_hash, genesis_state)]),
        }
    }
```

We don't need to define these types, and we also can avoid the clones in this function

crates/common/consensus/lean/src/staker.rs
```diff
@@ -0,0 +1,316 @@
...
+
+// TODO: Split to Staker and Service
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub struct Staker {
+    pub validator_id: u64,
+    pub public_key: PublicKey, // Additional to 3SF-mini
+    pub chain: HashMap<B256, Block>,
+    pub time: u64, // TODO: update the time so on_tick() works properly
+    // TODO: Add back proper networking instead
+    // pub network: Weak<RefCell<P2PNetwork>>,
```
KolbyML: We are building a multi-threaded application so should we use a `Arc<Mutex<>>` over a `Weak<RefCell<>>`?

crates/common/consensus/lean/src/staker.rs
```diff
@@ -0,0 +1,316 @@
...
+    vote::{SignedVote, Vote},
+};
+
+// TODO: Split to Staker and Service
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub struct Staker {
+    pub validator_id: u64,
+    pub public_key: PublicKey, // Additional to 3SF-mini
+    pub chain: HashMap<B256, Block>,
+    pub time: u64, // TODO: update the time so on_tick() works properly
```
KolbyML:
```suggestion

    /// Additional to 3SF-mini
    pub public_key: PublicKey, 
    pub chain: HashMap<B256, Block>,
    
    /// TODO: update the time so on_tick() works properly
    pub time: u64,
```

Comments should be placed above the variables

crates/common/consensus/lean/src/staker.rs
```diff
@@ -0,0 +1,316 @@
...
+            signature: PQSignature {},
+        };
+
+        println!(
+            "voting (Staker {}), head = {}, t = {}, s = {}",
+            self.validator_id,
+            &self.chain.get(&self.head).unwrap().slot,
+            &target_block.slot,
+            &state.latest_justified_slot
+        );
```
KolbyML:
```suggestion
        info!(
            "voting (Staker {}), head = {}, t = {}, s = {}",
            self.validator_id,
            &self.chain.get(&self.head).unwrap().slot,
            &target_block.slot,
            &state.latest_justified_slot
        );
```
tracing should be used instead of print statements

crates/common/consensus/lean/src/staker.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    // Called when it's the staker's turn to propose a block
+    fn propose_block(&mut self) {
+        let new_slot = self.get_current_slot();
+
+        println!(
+            "proposing (Staker {}), head = {}",
+            self.validator_id,
+            self.chain.get(&self.head).unwrap().slot
+        );
```
KolbyML:
```suggestion
        info!(
            "proposing (Staker {}), head = {}",
            self.validator_id,
            self.chain.get(&self.head).unwrap().slot
        );
```

crates/common/consensus/lean/src/staker.rs
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    // Done upon processing new votes or a new block
+    fn recompute_head(&mut self) {
+        let justified_hash = get_latest_justified_hash(&self.post_states).unwrap();
+        self.head = get_fork_choice_head(&self.chain, &justified_hash, &self.known_votes, 0);
+    }
+
+    // Called every second
+    pub fn tick(&mut self) {
```
KolbyML:
```suggestion
    /// Done upon processing new votes or a new block
    fn recompute_head(&mut self) {
        let justified_hash = get_latest_justified_hash(&self.post_states).unwrap();
        self.head = get_fork_choice_head(&self.chain, &justified_hash, &self.known_votes, 0);
    }

    /// Called every second
    pub fn tick(&mut self) {
```

Doc comments

crates/common/consensus/lean/src/lib.rs
```diff
@@ -1,4 +1,193 @@
...
+pub fn get_latest_justified_hash(post_states: &HashMap<B256, LeanState>) -> Option<B256> {
+    post_states
+        .values()
+        .max_by_key(|state| state.latest_justified_slot)
+        .map(|state| state.latest_justified_hash)
+}
+
+// Use LMD GHOST to get the head, given a particular root (usually the
+// latest known justified block)
+pub fn get_fork_choice_head(
```
KolbyML:
```suggestion
/// Get the highest-slot justified block that we know about
pub fn get_latest_justified_hash(post_states: &HashMap<B256, LeanState>) -> Option<B256> {
    post_states
        .values()
        .max_by_key(|state| state.latest_justified_slot)
        .map(|state| state.latest_justified_hash)
}

/// Use LMD GHOST to get the head, given a particular root (usually the
/// latest known justified block)
pub fn get_fork_choice_head(
```

crates/common/consensus/lean/src/lib.rs
```diff
@@ -1,4 +1,193 @@
...
+
+    let delta = candidate_slot - finalized_slot;
+
+    delta <= 5
+    || (delta as f64).sqrt().fract() == 0.0 // any x^2
+    || (delta as f64 + 0.25).sqrt() % 1.0 == 0.5 // any x^2+x
+}
+
+// Given a state, output the new state after processing that block
+pub fn process_block(pre_state: &LeanState, block: &Block) -> LeanState {
```
KolbyML: same as above

crates/common/consensus/lean/src/vote.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub source: B256,
+    pub source_slot: u64,
+}
+
+impl Vote {
+    pub fn compute_hash(&self) -> B256 {
+        let serialized = serde_json::to_string(self).unwrap();
+        B256::from_slice(&hash(serialized.as_bytes()))
+    }
+}
```
KolbyML: The hash of `Vote` is computed by calling `.tree_root_hash()` on the object like how it is done in the beacon chain right?, I don't think we need this function

crates/common/consensus/lean/src/block.rs
```diff
@@ -1,22 +1,36 @@
...
+    pub state_root: Option<B256>,
+}
 
-    /// Empty root
-    pub body: B256,
+impl Block {
+    pub fn compute_hash(&self) -> B256 {
+        let serialized = serde_json::to_string(self).unwrap();
+        B256::from_slice(&hash(serialized.as_bytes()))
+    }
```
KolbyML: same comment as above

crates/common/consensus/lean/src/state.rs
```diff
@@ -1,14 +1,30 @@
...
+    pub justified_slots: VariableList<bool, U4096>,
+
+    pub justifications: HashMap<B256, Vec<bool>>,
+}
+
+impl LeanState {
+    pub fn compute_hash(&self) -> B256 {
+        let serialized = serde_json::to_string(self).unwrap();
+        B256::from_slice(&hash(serialized.as_bytes()))
+    }
```
KolbyML: same comment as above

KolbyML (commented):
Left some feedback

KolbyML:
Generally ``make lint`` should be done before reviews are done, but eh, I think it is fine to give some light direction rn

crates/common/consensus/lean/src/vote.rs
syjn99: I guess this is a ported function from the original code (by Vitalik). Of course it will be deleted after using SSZ here.

https://github.com/ethereum/research/blob/d225a6775a9b184b5c1fd6c830cc58a375d9535f/3sf-mini/consensus.py#L46-L50

crates/common/consensus/lean/src/lib.rs
unnawut: I kept this implementation as is because I think it's intuitive that `get_latest_justified_hash()` returns `None` when the `post_states` is empty.

On the other hand I updated its caller (`recompute_head()` to return error instead).

syjn99: Agreed based on offline discussion! Thanks

crates/common/consensus/lean/src/lib.rs
unnawut: Really good catch! I'm supposed to used this to find the latest_votes a few lines below. Fixed.

crates/common/consensus/lean/src/block.rs
unnawut: There's quite a bit of vector manipulation which I think requires usize.

But I updated the code now to u64 and do `as usize` only when working with vectors.

crates/common/consensus/lean/src/vote.rs
unnawut: I've resolved all issues with SSZ compatibility and so replaced `compute_hash()` with `TreeHash` now.

crates/common/lean_chain/src/staker.rs
```diff
@@ -0,0 +1,327 @@
...
+    /// Process new votes that the staker has received. Vote processing is done
+    /// at a particular time, because of safe target and view merge rule
+    fn accept_new_votes(&mut self) {
+        let mut known_votes = self.known_votes.clone().into_iter();
+
+        for new_vote in &self.new_votes {
+            if !known_votes.any(|known_vote| known_vote == *new_vote) {
+                self.known_votes.push(new_vote.clone());
+            }
+        }
```
syjn99:
```suggestion
        for new_vote in &self.new_votes {
            if !self.known_votes.contains(new_vote) {
                self.known_votes.push(new_vote.clone());
            }
        }
```

I think this is more readable in this case. Same comment for `receive` function.

crates/common/lean_chain/src/staker.rs
```diff
@@ -0,0 +1,327 @@
...
+            state = process_block(head_state, &new_block);
+
+            let mut new_votes_to_add = Vec::<Vote>::new();
+            for vote in self.known_votes.clone().into_iter() {
+                if vote.source == state.latest_justified_hash
+                    && !new_block.votes.clone().into_iter().any(|v| v == vote)
+                {
+                    new_votes_to_add.push(vote);
+                }
+            }
```
syjn99:
```suggestion
            let new_votes_to_add = self
                .known_votes
                .clone()
                .into_iter()
                .filter(|vote| vote.source == state.latest_justified_hash)
                .filter(|vote| !new_block.votes.contains(vote))
                .collect::<Vec<_>>();
```

I personally prefer chaining and iterator. Up to you to pick this suggestion!

crates/common/consensus/lean/src/state.rs
```diff
@@ -1,14 +1,84 @@
...
+    pub latest_finalized_hash: B256,
+    pub latest_finalized_slot: u64,
+
+    pub historical_block_hashes: VariableList<B256, U4096>,
+    pub justified_slots: VariableList<bool, U4096>,
+
+    // Diverged from Python implementation:
+    // Originally `justifications: Dict[str, List[bool]]`
+    pub justifications_roots: VariableList<B256, U4096>,
+    pub justifications_roots_validators: BitList<U16777216>,
```
syjn99: So I guess these numbers imply that the chain can have maximum 4,096 validators (= 16777216 / 4096), right?

crates/common/consensus/lean/src/state.rs
```diff
@@ -1,14 +1,84 @@
...
+}
+
+impl LeanState {
+    fn get_justifications_roots_index(&self, root: &B256) -> Option<usize> {
+        self.justifications_roots.iter().position(|r| r == root)
+    }
+
+    pub fn initialize_justifications_for_root(&mut self, root: &B256) {
```
syjn99: Don't we have to also append `justifications_roots_validators` with `false`s? `BitList` is a variable sized types so I'm worried this doesn't fully initialize it... Doesn't this make `.set()` function return error like `OutOfBounds`?

```python
state.justifications[vote.target] = [False] * state.config.num_validators
```

I'm looking into it.

crates/common/consensus/lean/src/state.rs
unnawut: Yep it's max slots 4096 * max validators 4096.

But actually now that you mentioned I think 4096 is too low for max slots (last only 4 hours). I'll be updating max slots to 262144 (12 days) but validators will remain at 4096.

crates/common/lean_chain/src/staker.rs
unnawut: Much better. I was following the Python impl too blindly

crates/common/consensus/lean/src/state.rs
unnawut: My understanding is that the underlying data structure is a vector initialized to max length, prefilled with `00000000` here: https://github.com/sigp/ethereum_ssz/blob/main/ssz/src/bitfield.rs#L281-L295

```rs
    /// Instantiate a new `Bitfield` with a fixed-length of `N` bits.
    ///
    /// All bits are initialized to `false`.
    pub fn new() -> Self {
        Self {
            bytes: smallvec![0; bytes_for_bit_len(Self::capacity())],
            len: Self::capacity(),
            _phantom: PhantomData,
        }
    }

    /// Returns `N`, the number of bits in `Self`.
    pub fn capacity() -> usize {
        N::to_usize()
    }
```

So what I did is zerofill on removal here: https://github.com/ReamLabs/ream/pull/672/files#diff-a7d630b406cdd46523f5b854dc367f8677aed953706b4a41e9813554b6bc3c89R77-R82 (I havn't seen a better function to do this except for directly modifying the underlying bytes but I think it's too far for this PR)

crates/common/consensus/lean/src/state.rs
```diff
@@ -1,14 +1,90 @@
...
+
+impl LeanState {
+    fn get_justifications_roots_index(&self, root: &B256) -> Option<usize> {
+        self.justifications_roots.iter().position(|r| r == root)
+    }
+
+    fn get_justifications_roots_range(&self, index: &usize) -> (usize, usize) {
+        // Start from index * MAX_HISTORICAL_BLOCK_HASHES, ends at start + VALIDATOR_REGISTRY_LIMIT
+        let start_range = index * U262144::to_usize();
+        let end_range = start_range + U4096::to_usize();
```
KolbyML:
```suggestion
        let start_range = index * 262144;
        let end_range = start_range + 4096;
```

Can't we just do this?

crates/common/lean_chain/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+ssz_types.workspace = true
+tokio.workspace = true
+tracing.workspace = true
+tree_hash.workspace = true
+
+# ream dependencies
+ream-consensus-lean.workspace = true
+ream-network-spec.workspace = true
+ream-p2p.workspace = true
+ream-pqc.workspace = true
```
KolbyML: should we structure this like our other crates like `common/consensus/*`, `maybe common/chain/*`, so this crate would become ream-chain-lean, and the beacon one would be ream-chain-beacon

crates/common/lean_chain/src/staker.rs
```diff
@@ -0,0 +1,324 @@
...
+    /// Process new votes that the staker has received. Vote processing is done
+    /// at a particular time, because of safe target and view merge rule
+    fn accept_new_votes(&mut self) {
+        for new_vote in &self.new_votes {
+            if !self.known_votes.contains(new_vote) {
+                self.known_votes.push(new_vote.clone());
+            }
+        }
+
+        self.new_votes = Vec::new();
```
KolbyML: shouldn't we consume `new_votes` instead of cloning its items, rather than reallocating it at the end? This seems a little wasteful

crates/common/consensus/lean/src/state.rs
unnawut: Btw. I was also thinking of using `VariableList` but opted out basically because of the need to loop zerofill both at init and removal, plus I _think_ `VariableList` takes 1 byte per `bool`.

crates/common/consensus/lean/src/state.rs
syjn99: I think `justifications_roots_validators` also needs to be extended to `U1073741824` (=  262144 * 4096 = 2^18 * 2^12 = 2^30)if you want to cap validator set size as 4096.

crates/common/lean_chain/src/staker.rs
```diff
@@ -0,0 +1,324 @@
...
+                        self.chain.insert(block_hash, block.clone());
+                        self.post_states.insert(block_hash, state);
+
+                        let mut known_votes = self.known_votes.clone().into_iter();
+
+                        for vote in &block.votes {
+                            if !known_votes.any(|known_vote| known_vote == *vote) {
+                                self.known_votes.push(vote.clone());
+                            }
+                        }
```
KolbyML:
```suggestion
    /// Called by the p2p network
    fn receive(&mut self, queue_item: QueueItem) {
        match queue_item {
            QueueItem::BlockItem(block) => {
                let block_hash = block.tree_hash_root();

                // If the block is already known, ignore it
                if self.chain.contains_key(&block_hash) {
                    return;
                }

                match self.post_states.get(&block.parent) {
                    Some(parent_state) => {
                        let state = process_block(parent_state, block);


                        let mut known_votes = self.known_votes.clone().into_iter();

                        for vote in &block.votes {
                            if !known_votes.any(|known_vote| known_vote == *vote) {
                                self.known_votes.push(vote.clone());
                            }
                        }
                        
                        self.post_states.insert(block_hash, state);
                        self.chain.insert(block_hash, block);
```

we can remove this clone no? I read the other uses of this function, and I don't think there is a reason for `queue_item: &QueueItem` should be a reference

crates/common/lean_chain/src/staker.rs
```diff
@@ -0,0 +1,324 @@
...
+                if is_known_vote || is_new_vote {
+                    // Do nothing
+                } else if self.chain.contains_key(&vote.data.head) {
+                    self.new_votes.push(vote.data.clone());
+                } else {
+                    self.dependencies
+                        .entry(vote.data.head)
+                        .or_default()
+                        .push(queue_item.clone());
+                }
```
KolbyML:
```suggestion
                    self.new_votes.push(vote.data);
                } else {
                    self.dependencies
                        .entry(vote.data.head)
                        .or_default()
                        .push(QueueItem::VoteItem(vote));
                }
```
same here

crates/common/lean_chain/src/staker.rs
syjn99: This part also can be simplified as similar as [this](https://github.com/ReamLabs/ream/pull/672#discussion_r2247918447).

KolbyML (commented):
Could we remove the rest of the `unwrap()`'s as well, although I think we should just use anyhow, as it isn't much work

crates/common/consensus/lean/src/state.rs
syjn99: > My understanding is that the underlying data structure is a vector initialized to max length, prefilled with 00000000 here: https://github.com/sigp/ethereum_ssz/blob/main/ssz/src/bitfield.rs#L281-L295

Actually you are referencing `Bitfield<Fixed<N>>` type. 

```rust
/// A heap-allocated, ordered, variable-length collection of `bool` values, limited to `N` bits.
pub type BitList<N> = Bitfield<Variable<N>>;
```

We need to look up `Bitfield<Variable<N>>` in this case, which starts from https://github.com/sigp/ethereum_ssz/blob/5b6dd174f5465978496a083c697d36498192c261/ssz/src/bitfield.rs#L122. It supports `with_capacity` method to make a new instance of it. 

Also it says "`Bitlist` cannot grow or shrink.". I'm still okay if we initialize `justifications_roots_validators` with full capacity like:

```rust
justifications_roots_validators: BitList<typenum::U16777216>::with_capacity(16777216).unwrap()
```

But it seems we don't fully enjoy what `Bitlist` provides: dynamic allocation. We might use `Bitvector` here as serialization is [more efficient](https://eth2book.info/capella/part2/building_blocks/ssz/#bitlists).

Or we might choose to replace `justifications_roots_validators` every time with new `Bitlist` instance with limited capacity.

crates/common/lean_chain/Cargo.toml
unnawut: Did this in a separate PR so the diffs are cleaner: https://github.com/ReamLabs/ream/pull/676

crates/common/consensus/lean/src/state.rs
unnawut: I just realized my current implementation by zerofilling `justifications_roots_validators` won't work because removing an item from `justifications_roots` will change the index of all following roots and so I would need to shift `justifications_roots_validators` by `MAX_VALIDATOR_REGISTRY` as well. While I can fix this plus change to `BitVector`, I think it's over-optimization at this point.

A variable `BitList` also represents this data more accurately so I'll go with your suggestion "replace `justifications_roots_validators` every time with new Bitlist instance with limited capacity."

crates/common/consensus/lean/src/state.rs
```diff
@@ -1,14 +1,330 @@
...
+    // Diverged from Python implementation:
+    // Originally `justifications: Dict[str, List[bool]]`
+    pub justifications_roots: VariableList<B256, U262144>,
+    // The size is MAX_HISTORICAL_BLOCK_HASHES * VALIDATOR_REGISTRY_LIMIT
+    // to accommodate equivalent to `justifications[root][validator_id]`
+    pub justifications_roots_validators: BitList<U1073741824>,
+}
+
+impl LeanState {
+    pub fn new(num_validators: u64) -> anyhow::Result<LeanState> {
```
syjn99: I feel this shouldn't return `Error` type, maybe using `expect()` to avoid returning `Result` type? (`BitList::with_capacity(0)` will *never* fail)

crates/common/consensus/lean/src/state.rs
```diff
@@ -1,14 +1,330 @@
...
+            })?,
+        })
+    }
+
+    fn get_justifications_roots_index(&self, root: &B256) -> Option<usize> {
+        self.justifications_roots.iter().position(|r| r == root)
+    }
+
+    pub fn initialize_justifications_for_root(&mut self, root: &B256) -> anyhow::Result<()> {
+        if !self.justifications_roots.contains(root) {
```
syjn99: How about return early using `if self.justifications_roots.contains(root)` to avoid unnecessary indentation?

crates/common/consensus/lean/src/state.rs
```diff
@@ -1,14 +1,330 @@
...
+
+    pub fn set_justification(
+        &mut self,
+        root: &B256,
+        validator_id: &u64,
+        value: bool,
+    ) -> anyhow::Result<()> {
+        let index = self
+            .get_justifications_roots_index(root)
+            .ok_or(anyhow!("Failed to find the justifications index to set"))?;
```
syjn99:
```suggestion
            .ok_or_else(|| anyhow!("Failed to find the justifications index to set for {root}"))?;
```
1. using `ok_or_else` seems more consistent here
2. Sorry for nitpicking here, but inlining the argument would help. Similar comments for other ad-hoc errors using `anyhow!`.

crates/common/consensus/lean/src/state.rs
```diff
@@ -1,14 +1,330 @@
...
+        self.justifications_roots.remove(index);
+
+        let new_length = self.justifications_roots.len() * VALIDATOR_REGISTRY_LIMIT as usize;
+        let mut new_justifications_roots_validators =
+            BitList::<U1073741824>::with_capacity(new_length).map_err(|e| {
+                anyhow!(
+                    "Failed to recreate state's justifications_roots_validators: {:?}",
+                    e
+                )
+            })?;
```
syjn99:
```suggestion
        let mut new_justifications_roots_validators =
            BitList::<U1073741824>::with_capacity(new_length).map_err(|err| {
                anyhow!("Failed to recreate state's justifications_roots_validators: {err:?}")
            })?;
```

Another nitpicking (sry)

unnawut: All valid point! I reviewed and made changes to other files as well

PR #673, opened 2025-08-01
ci: lint PR title instead of commits

syjn99:
### What was wrong?

 

It wasn't wrong to check the convention, but as we chose "Squash & Merge" as our primary merge options, I think it would be better to check PR title only.

### How was it fixed?

 

- Mostly inspired from an example in [`amannn/action-semantic-pull-request`](https://github.com/amannn/action-semantic-pull-request) 
- Use [`amannn/action-semantic-pull-request`](https://github.com/amannn/action-semantic-pull-request) and replace `conventional-commits.yml`.
- Add some automated action for commenting to notify contributors to fix.
- Change one TODO item in our PR template.

### To-Do

 
### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -22,6 +22,8 @@ jobs:
       pull-requests: write
     steps:
       - uses: amannn/action-semantic-pull-request@v5
+        # Hack from Github Issue: https://github.com/amannn/action-semantic-pull-request/issues/236#issuecomment-1695654373
+        if: ${{ github.event_name != 'merge_group' }}
         id: lint_pr_title 
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/pr-title.yml
```diff
@@ -0,0 +1,45 @@
+name: Pull Request Title
+
+on:
+  pull_request:
```
KolbyML, line 4:
```suggestion
on:
  pull_request:
    types:
      - opened
      - edited
      - reopened
```

Don't you need to declare the types edited, to make it so the CI runs on title change

syjn99: oh good point

syjn99: also added `synchronized` as well. thanks!

syjn99:
Actions for commenting is not working in **this PR**. I guess this PR comes from my repository (which is [forked repo](https://stackoverflow.com/questions/75744849/permission-to-repo-denied-to-github-actions-for-pr-coming-from-a-fork-only)).

syjn99:
Oh god this cannot be triggered using merge_group

syjn99:
@KolbyML 

https://github.com/amannn/action-semantic-pull-request/issues/236#issuecomment-1695654373

I found similar issue here. Hope that it fixes the issue (Still not so happy as it's kinda hacky :/)

PR #674, opened 2025-08-02
feat: disconnect peer if disconnected from swarm

Kayden-ML:
### What was wrong?

peers would still be connected if disconnected to the swarm

### How was it fixed?

disconnected peer if its disconnected from the swarm


KolbyML (approved):
:shipit: great find, looking forward too seeing more P2P improvements soon :saluting_face:

PR #675, opened 2025-08-04
feat: only insert peers into peers_to_ping after they are connected

Kayden-ML:
### What was wrong?

Peers were being pinged before they were connected and it would throw a bunch of logs saying peer not connected skipping peer

### How was it fixed?

made it to Insert peers into peers_to_ping after they are connected


KolbyML (approved):
:shipit: looks good

Issue #393, opened 2025-05-09
P2P Meta Tracking issue


KolbyML:
### Describe the feature

This issue is for tracking all P2P subissues

KolbyML:
@Nanle100 this is a meta issue, if you want you can make your own issue to work on

PR #678, opened 2025-08-05
chore: add constant for lean

syjn99:
### What was wrong?

 

We might need some constants different from the beacon chain. (e.g., `INTERVALS_PER_SLOT` is 4)

Part of #393.

### How was it fixed?

 

This PR simply add a module in `ream-consensus-misc` for constant. Also my clippy complains about `let-chain` so I fixed it also. (Therefore `make pr` is now happy)

### To-Do

 

unnawut:
I'm not sure if `ream_consensus_misc` needs to host both beacon and lean stuff? They seem entirely independent. Even misc could go into ream_consensus_beacon/lean. But maybe I don't know there's a reason for splitting misc from consensus crate?

But if so I think splitting the crates into one of these pairs make more sense, but just 2 cents:

1. `ream_consensus_beacon::misc::constants` & `ream_consensus_lean::misc::constants`
2. `ream_misc_beacon::constants` & `ream_misc_lean::constants` but personally 1 makes more sense.

PR #681, opened 2025-08-06
feat: Import anyhow instead of adding it everytime its used

Kayden-ML:
### What was wrong?

we where calling anyhow::anyhow( everytime it was used

### How was it fixed?

Imports anyhow


shariqnaiyer (approved):
looks good 👍

Issue #685, opened 2025-08-08
Initialize lean libp2p


Kayden-ML:
### Describe the feature

Initialize lean libp2p

PR #686, opened 2025-08-08
feat: initialize lean libp2p

Kayden-ML:
### What was wrong?

Fixes: https://github.com/ReamLabs/ream/issues/685

### How was it fixed?

I Initialized lean libp2p

### DIFF
--- a/crates/networking/p2p/src/network/lean.rs
+++ b/crates/networking/p2p/src/network/lean.rs
@@ -1,28 +1,134 @@
-use std::sync::Arc;
+use std::{
+    num::{NonZeroU8, NonZeroUsize},
+    sync::Arc,
+};
 
+use anyhow::anyhow;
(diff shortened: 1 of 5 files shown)


crates/networking/p2p/src/network/lean.rs
```diff
@@ -1,28 +1,134 @@
...
+                .with_max_established_per_peer(Some(1));
+
+            connection_limits::Behaviour::new(limits)
+        };
+
+        let local_key = Keypair::generate_secp256k1();
+        let identify = {
+            let local_public_key = local_key.public();
+            let identify_config =
+                identify::Config::new("eth2/1.0.0".into(), local_public_key.clone())
```
syjn99, line 67: Protocol version is same with libp2p of beacon chain, does this matter?

PR #690, opened 2025-08-10
feat(lean): static peer connections

### LINKED ISSUE #682: Support static peer connections for the lean node
### Describe the feature

We first need to have a feature in our network service that supports static peer connection. A node will be aware of the list of the other nodes with some kind of config file.

This task might include the basic peer management implementation to support the main feature.

### Additional context

_No response_

varun-doshi:
### What was wrong?

Fixes #682 
### How was it fixed?

 
- On network startup, lean node dials a list of static known peers from a file
- Reuses exists `Bootnodes` cli argument from beacon to allow user to specify enr from cli
- On successful dialing, adds the `peer_id` to a `peer_table` which is defined as a hashmap of `(PeerId, ConnectionState)` for now
- Run test using:
```
 RUST_LOG=info cargo test --package ream-p2p --lib -- network::lean::tests::test_two_nodes_connection --exact --show-output
```

### How to run
- run first node using:
```
cargo run --release lean_node --network ./bin/ream/assets/lean/sample_spec.yml --socket-port 9000
```
From the logs copy the `peer_id` and run the second node passing in the `multiaddr` of the first node in the cmd:
```
cargo run --release lean_node --network ./bin/ream/assets/lean/sample_spec.yml --socket-port 9001 --bootnodes /ip4/127.0.0.1/tcp/9000/p2p/16Uiu2HAmPfwBy2kJogK4E2iu5J4TTKQEK76fx28b4ioa9XryDLAv
```

### To-Do

I'm not sure whether we need `enr` in the static file or some other format(direct `multiaddr`). Will update this if needed depending on suggestions ✅


crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,7 +154,77 @@ impl LeanNetworkService {
...
+            for multiaddr in multiaddrs {
+                if let Err(err) = self.swarm.dial(multiaddr) {
+                    warn!("Failed to dial peer: {err:?}");
+                } else {
+                    successfully_dialed = true;
+                }
+            }
+
+            if !successfully_dialed {
+                trace!("Failed to dial any multiaddr for peer: {:?}", enr);
```
Kayden-ML: Can you inline this.

crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,7 +154,77 @@ impl LeanNetworkService {
...
+    match enr.public_key() {
+        CombinedPublicKey::Secp256k1(public_key) => {
+            let encoded_public_key = public_key.to_encoded_point(true);
+            let public_key = Secp256k1PublicKey::try_from_bytes(encoded_public_key.as_bytes())
+                .ok()?
+                .into();
+            Some(PeerId::from_public_key(&public_key))
+        }
+        _ => None,
+    }
```
KolbyML: we shouldn't be duplicating this, please reuse this function across the 2 files

crates/networking/p2p/src/network/lean.rs
```diff
@@ -98,16 +107,33 @@ impl LeanNetworkService {
...
         info!(
             "Current LeanChain head: {}",
             self.lean_chain.read().await.head
         );
+
+        let initial_peers = match peer_config {
+            Bootnodes::Default => Bootnodes::get_static_lean_peers(),
+            Bootnodes::None => vec![],
+            Bootnodes::Custom(enrs) => [enrs, Bootnodes::get_static_lean_peers()].concat(),
+        };
```
KolbyML: this code should be within the Bootnodes struct

crates/networking/p2p/src/network/lean.rs
```diff
@@ -98,16 +107,33 @@ impl LeanNetworkService {
...
+        let mut peers = HashMap::new();
+        for peer in initial_peers {
+            peers.insert(peer, None);
+        }
+
+        self.connect_to_peers(peers).await;
```
KolbyML: this can be inlined with an iter no?

crates/networking/p2p/src/network/lean.rs
```diff
@@ -98,16 +107,33 @@ impl LeanNetworkService {
                 .with_swarm_config(|_| config)
                 .build()
         };
+        let peer_table = RwLock::new(HashMap::new());
```
KolbyML: just inline this

KolbyML (commented):
Here is some feedback

syjn99:
> I'm not sure whether we need enr in the static file or some other format(direct multiaddr). Will update this if needed depending on suggestions

I think it would be nice to support multiaddr form, because this will make our lives easier when testing it in locally

varun-doshi:
> I think it would be nice to support multiaddr form, because this will make our lives easier when testing it in locally

Yes, thats what I was thinking as well. But on actually starting to implement this, here is a issue/open question:
- if we switch to connecting using `multiaddr` we cannot use the `Bootnodes` enum(atleast in the form it is currently). afaik direct conversion from multiaddr to enr is not possible as we need keypair. (considering here multiaddr is something like `/ip4/127.0.0.1/tcp/40837/p2p/12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X`). So the question is do we move away from enr format all together(and by extension drop usage of `Bootnodes`) or still allow passing in both formats(modify `Bootnodes` enum)

KolbyML:
> > I think it would be nice to support multiaddr form, because this will make our lives easier when testing it in locally
> 
> Yes, thats what I was thinking as well. But on actually starting to implement this, here is a issue/open question:
> 
> * if we switch to connecting using `multiaddr` we cannot use the `Bootnodes` enum(atleast in the form it is currently). afaik direct conversion from multiaddr to enr is not possible as we need keypair. (considering here multiaddr is something like `/ip4/127.0.0.1/tcp/40837/p2p/12D3KooWPjceQrSwdWXPyLLeABRXmuqt69Rg3sBYbU1Nft9HyQ6X`). So the question is do we move away from enr format all together(and by extension drop usage of `Bootnodes`) or still allow passing in both formats(modify `Bootnodes` enum)

Probably just adding a Bootnodes::Multiaddr(multiaddrs) would work right? We will be using Enr's in the future regardless I am assuming, so whatever "MultiAddr" thing we add will eventually be removed for Enr's, so I think just patching Bootnodes with the temporary change would be best, rather then building something distinct. Maybe make a function for Bootnodes called `to_multiaddrs`, which just returns a vector or hashmap of MultiAddr's

crates/networking/p2p/src/network/lean.rs
```diff
@@ -58,6 +62,7 @@ pub struct LeanNetworkConfig {
 pub struct LeanNetworkService {
     lean_chain: Arc<RwLock<LeanChain>>,
     swarm: Swarm<ReamBehaviour>,
+    peer_table: RwLock<HashMap<PeerId, ConnectionState>>,
```
syjn99: hmm I don't think we have to use `tokio::sync::RwLock` here. using `parking_lot::RwLock` would work as `peer_table` won't be shared in the concurrency context.
Plus we can remove some `.await` parts here

crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,15 +133,22 @@ impl LeanNetworkService {
...
 
-    pub async fn start(mut self) -> anyhow::Result<()> {
+    pub async fn start(mut self, peer_config: Bootnodes) -> anyhow::Result<()> {
         info!("LeanNetworkService started");
         info!(
             "Current LeanChain head: {}",
             self.lean_chain.read().await.head
         );
+
+        self.connect_to_peers(peer_config.get_static_lean_peers())
```
syjn99: Can this part run at `new` function? It makes more sense to provide config in initialization IMO.

crates/networking/p2p/resources/lean_peers.yaml
```diff
@@ -0,0 +1,2 @@
+- /ip4/4.196.214.4/tcp/9000/p2p/16Uiu2HAkytpf5xwHoRu13aCHPmqwFPF4YM5QT2jeGyiiKUc81c3F
+- /ip4/4.157.240.54/tcp/9000/p2p/16Uiu2HAkw949aUhLTe7QPCG9N8wfELtNVwzXXYXuuwknkA582bcX
```
syjn99: Just wonder where this peer list came from. Is this like a placeholder only for this PR?

crates/networking/p2p/src/bootnodes.rs
```diff
@@ -55,6 +70,56 @@ impl Bootnodes {
...
+                    && let Some(tcp6) = enr.tcp6()
+                {
+                    let mut multiaddr: Multiaddr = ip6.into();
+                    multiaddr.push(Protocol::Tcp(tcp6));
+                    multiaddr.push(Protocol::P2p(peer_id));
+                    multiaddrs.push(multiaddr);
+                }
+            }
         }
+        multiaddrs
```
syjn99: I found this similar code in other function(`Network::handle_discovered_peers`). I think this doesn't have to be one of the member in `Bootnodes` struct: rather it makes more sense to be a pure util function. This function just gets a list of ENR and converts them to Multiaddrs, which doesn't need any state.

syjn99 (commented):
Initial comments! Thanks!

crates/networking/p2p/resources/lean_peers.yaml
varun-doshi: Ahh yes I converted some mainnet bootnodes enr to the respective multiaddrs, I think these were the Teku nodes. But yes these are just placeholders

crates/networking/p2p/src/network/lean.rs
varun-doshi: Hmm actually `connect_to_peers` this mutates `self` so generally speaking I feel it should not be in the new function. 
`start` takes a mutable reference regardless and so it might be a better place for that call.
What do you think?

syjn99: oh yes, your point is valid. Gonna mark this convo as resolved.

crates/networking/p2p/src/utils.rs
```diff
@@ -28,3 +30,28 @@ pub fn read_meta_data_from_disk(path: PathBuf) -> anyhow::Result<GetMetaDataV2>
     GetMetaDataV2::from_ssz_bytes(&fs::read(meta_data_path)?)
         .map_err(|err| anyhow!("Failed to decode meta data: {err:?}"))
 }
+
+pub fn to_multiaddrs(enrs: Vec<Enr>) -> Vec<Multiaddr> {
+    let mut multiaddrs: Vec<Multiaddr> = Vec::new();
+    for enr in enrs {
+        if let Some(peer_id) = crate::network::beacon::Network::peer_id_from_enr(&enr) {
```
syjn99: I guess `peer_id_from_enr` also can be placed in this file

crates/networking/p2p/src/network/beacon.rs
```diff
@@ -990,8 +977,8 @@ mod tests {
         let mut network_1 = runtime
             .block_on(create_network(
                 "127.0.0.1".parse::<IpAddr>().unwrap(),
-                9000,
-                9001,
+                9090,
+                9091,
```
syjn99, line 981: Could you tell me why those ports are changed?

bin/ream/src/cli/lean_node.rs
```diff
@@ -15,4 +18,20 @@ pub struct LeanNodeConfig {
...
+        long,
+        help = "One or more comma-delimited base64-encoded ENR's of peers to initially connect to. Use 'default' to use the default bootnodes for the network. Use 'none' to disable bootnodes."
+    )]
+    pub bootnodes: Bootnodes,
+
+    #[arg(long, help = "Set P2P socket address", default_value_t = DEFAULT_SOCKET_ADDRESS)]
+    pub socket_address: IpAddr,
+
+    #[arg(long, help = "Set P2P socket port (TCP)", default_value_t = DEFAULT_SOCKET_PORT)]
+    pub socket_port: u16,
```
syjn99, line 36: Just a side note: we will probably use QUIC as default. I think it is okay to tackle this issue in next PRs.

syjn99: Opened an issue: https://github.com/ReamLabs/ream/issues/697

crates/networking/discv5/src/lean.rs
```diff
@@ -0,0 +1,45 @@
+use std::net::{IpAddr, Ipv4Addr};
+
+use discv5::{ConfigBuilder, ListenConfig};
+
+#[derive(Debug, Clone)]
+pub struct LeanDiscoveryConfig {
```
syjn99: Another side note: I think this can be refactored and reused similarly with Beacon node. I don't think the priority of it is high enough.

bin/ream/src/main.rs
```diff
@@ -162,7 +171,10 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
         chain_service.start().await;
     });
     let network_future = executor.spawn(async move {
-        if let Err(err) = network_service.start().await {
+        if let Err(err) = network_service
+            .start(config.bootnodes, discovery_config)
```
syjn99: I think we don't have to pass `discovery_config` here as it is already passed in `new` function.

crates/networking/p2p/src/network/lean.rs
```diff
@@ -153,11 +182,156 @@ impl LeanNetworkService {
...
-                Some(ReamBehaviourEvent::Identify(_)) => None,
-                _ => None,
-            },
+            SwarmEvent::Behaviour(_) => None,
+            SwarmEvent::ConnectionEstablished { peer_id, .. } => {
+                self.peer_table
+                    .write()
+                    .insert(peer_id, ConnectionState::Connected);
+                info!("Connected to peer: {peer_id:?}");
+                None
```
syjn99, line 193: Why do all arms return `None`? I think logging also can be handled in our `start` function by wrapping those result with `ReamNetworkEvent`. If you want to print the events prettier, we might derive it using `derive_more` crate or sth.

crates/networking/p2p/src/network/lean.rs
```diff
@@ -153,11 +182,156 @@ impl LeanNetworkService {
...
+                info!("Dialing peer: {peer_id:?}",);
+                self.peer_table
+                    .write()
+                    .insert(peer_id, ConnectionState::Connecting);
+            }
+        }
+    }
+}
+
+#[allow(unused_imports)]
```
syjn99: Can this be deleted?

bin/ream/src/cli/lean_node.rs
```diff
@@ -15,4 +18,20 @@ pub struct LeanNodeConfig {
...
+
+    #[arg(long, help = "Discovery 5 listening port (UDP)", default_value_t = DEFAULT_DISCOVERY_PORT)]
+    pub discovery_port: u16,
```
KolbyML:
```suggestion
```

There are no plans to use discv5 yet, so we shouldn't add a config argument for it

crates/networking/p2p/src/network/lean.rs
```diff
@@ -153,11 +182,156 @@ impl LeanNetworkService {
...
+                .start(bootnodes, config2.discovery_config.clone())
+                .await
+        });
+
+        tokio::time::sleep(Duration::from_secs(2)).await;
+
+        node1_handle.abort();
+        node2_handle.abort();
+
+        Ok(())
```
syjn99, line 320: Thanks for adding the tests! I think it is enough at this moment.

One minor concern on top of my head is that it actually doesn't verify anything. Devs should check tracing logs manually. There are two possible improvements:

1. Verify `peer_table` of each node: this requires refactoring our code to inject the `peer_table` from outside.
2. Verify tracing logs using `tracing_test`: This is much easier. See examples like `logs_contain` [here](https://docs.rs/tracing-test/latest/tracing_test/).

I don't want to block this work by adding tests, so let's handle it in subsequent PRs.

bin/ream/src/main.rs
```diff
@@ -144,9 +145,17 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
     // TODO 1: Load keystores from the config.
     // TODO 2: Add RPC service for lean node.
     let chain_service = LeanChainService::new(lean_chain.clone(), chain_receiver).await;
+    let gossipsub_config = LeanGossipsubConfig::default();
```
KolbyML: why did you un-inline this? It doesn't look like this variable is used anywhere and it isn't complex so it should just be inlined again

bin/ream/src/main.rs
```diff
@@ -144,9 +145,17 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
...
+    let discovery_config = LeanDiscoveryConfig::new(
```
KolbyML: why is this called a discovery config, these variables should just be apart of the already existing LeanNetworkConfig, as the socket ip/port are just network configurations not Discv5 configurations

crates/networking/discv5/src/lean.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub discv5_config: discv5::Config,
+    pub socket_address: IpAddr,
+    pub socket_port: u16,
+    pub discovery_port: u16,
+    pub disable_discovery: bool,
```
KolbyML: I don't know why there are 3 fields for discv5 here, but as I said above socket ip/port should go into the network config. So this file should be removed

crates/networking/p2p/src/bootnodes.rs
```diff
@@ -19,12 +23,25 @@ impl FromStr for Bootnodes {
...
+                if let Ok(addrs) = s
+                    .split(',')
+                    .map(Multiaddr::from_str)
+                    .collect::<Result<Vec<_>, _>>()
+                {
+                    return Ok(Bootnodes::Multiaddr(addrs));
+                }
+
+                Err(anyhow!("Failed to parse bootnodes as ENR or Multiaddr"))
+            }
```
KolbyML: we should probably add a prefix like enr:xyz and multi:abc, so that the error here isn't ambiguous

crates/networking/p2p/src/bootnodes.rs
```diff
@@ -19,12 +23,25 @@ impl FromStr for Bootnodes {
...
+            _ => {
+                if let Ok(enrs) = s
+                    .split(',')
+                    .map(Enr::from_str)
+                    .collect::<Result<Vec<_>, _>>()
+                {
+                    return Ok(Bootnodes::Custom(enrs));
+                }
+
+                if let Ok(addrs) = s
```
KolbyML:
```suggestion
                if let Ok(addresses) = s
```

crates/networking/p2p/src/bootnodes.rs
```diff
@@ -55,6 +72,31 @@ impl Bootnodes {
             Bootnodes::Default => bootnodes,
             Bootnodes::None => vec![],
             Bootnodes::Custom(bootnodes) => bootnodes,
+            Bootnodes::Multiaddr(_) => vec![],
+        }
+    }
+
+    pub fn get_static_lean_peers(&self) -> Vec<Multiaddr> {
+        match self {
```
KolbyML: this functions name isn't descriptive

```suggestion
    pub fn as_multiaddrs(&self) -> Vec<Multiaddr> {
        match self {
```

This function also has nothing to do with lean

crates/networking/p2p/src/network/beacon.rs
syjn99: I guess running this with newly added tests has some conflicts regarding ports

crates/networking/p2p/src/utils.rs
```diff
@@ -28,3 +33,28 @@ pub fn read_meta_data_from_disk(path: PathBuf) -> anyhow::Result<GetMetaDataV2>
...
+            {
+                let mut multiaddr: Multiaddr = ip6.into();
+                multiaddr.push(Protocol::Tcp(tcp6));
+                multiaddr.push(Protocol::P2p(peer_id));
+                multiaddrs.push(multiaddr);
+            }
+        }
+    }
+    multiaddrs
+}
```
KolbyML: Please implement the from trait, so do From<Enr> for Multiaddr then just do an iter where you call to_multiaddrs

KolbyML (changes requested):
Here is some feedback

crates/networking/p2p/resources/lean_peers.yaml
syjn99: Hmm do we need placeholders that eventually won't be used in a near future?

crates/networking/p2p/src/network/lean.rs
KolbyML: There shouldn't be any unused imports I would think :+1:

crates/networking/p2p/src/utils.rs
syjn99: I also considered to comment this, but can a single ENR be converted to multiple Multiaddrs? `From` trait might not be suitable for this case IMO.

crates/networking/p2p/src/network/beacon.rs
varun-doshi: Yes

crates/networking/p2p/src/utils.rs
KolbyML: can't you do for Vec<x>?, then and use a flatten iter

KolbyML: I guess it is fine then, but regardless Vec<Enr> should be a slice I would think, it doesn't seem like we need to consume the enrs

crates/networking/p2p/src/bootnodes.rs
varun-doshi: wdym? This function fetches the static lean peers from the yaml file

KolbyML: Could you rename this to_multiaddrs_lean then, also there is a lot of code duplication that can be reduced. Change `to_enrs` to `to_enrs_beacon` or something

crates/networking/p2p/src/bootnodes.rs
varun-doshi: You mean to change the format of the actual enr/multiaddr?

crates/networking/p2p/src/network/lean.rs
varun-doshi: We can verify the peer table easily without needing any injection from outside so its fine I have added the assertions in this

crates/networking/p2p/src/network/lean.rs
varun-doshi: I've returned the approproate network events. I think we can prettify the output later

crates/networking/p2p/resources/lean_peers.yaml
syjn99, line 1: Still I don't understand why this file is needed, as [I previously said](https://github.com/ReamLabs/ream/pull/690/files#r2272070834).. If it is a placeholder, it makes sense to be an empty file so that the node doesn't parse anything wrong.

```
2025-08-13T11:34:21.825832Z  WARN ream_p2p::network::lean: Failed to connect to Some(PeerId("16Uiu2HAkw949aUhLTe7QPCG9N8wfELtNVwzXXYXuuwknkA582bcX")): Transport([(/ip4/4.157.240.54/tcp/9000/p2p/16Uiu2HAkw949aUhLTe7QPCG9N8wfELtNVwzXXYXuuwknkA582bcX, Other(Custom { kind: Other, error: Other(Right(Custom { kind: Other, error: Transport(Custom { kind: Other, error: Timeout }) })) }))])
```

We will always face this kind of error log when we run the node and test it.

crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,36 +137,191 @@ impl LeanNetworkService {
...
+
+    use libp2p::{Multiaddr, multiaddr::Protocol};
+    use ream_chain_lean::lean_chain::LeanChain;
+    use ream_network_spec::networks::{LeanNetworkSpec, set_lean_network_spec};
+
+    use super::*;
+    use crate::bootnodes::Bootnodes;
+
+    #[tokio::test]
+    async fn test_two_lean_nodes_connection() -> anyhow::Result<()> {
```
syjn99, line 257: I'm afraid that this test is wrong actually. There are a couple of flaws here.

1. Two asynchronous tasks are aborted immediately they got spawned.
2. `node1.start(bootnodes).await.unwrap();` will never finished, which means the assertion would never happen. Also it would never wait for 2 seconds.

You may simply check by changing the assertion statement (e.g., `assert_eq!(binding.len(), 2)`). The test will success.

I guess you chose the first option from [my previous suggestion](https://github.com/ReamLabs/ream/pull/690/files#r2272061224):

> Verify `peer_table` of each node: this requires refactoring our code to inject the `peer_table` from outside.

^ This is actually not properly managed. As the ownership of `node1` and `node2` are moved away to newly generated tokio tasks, in this design we cannot look up `node1`'s `peer_table` after it is spawned.

syjn99 (changes requested):
I think the test should be fixed

crates/networking/p2p/resources/lean_peers.yaml
unnawut: Also is multiaddr still used? I only looked up quickly and it seems [ENR already replaced multiaddr](https://ethereum.org/en/developers/docs/networking-layer/network-addresses/#enr):

> Ethereum Node Records (ENRs) are a standardized format for network addresses on Ethereum. They supersede multiaddr's and enodes.

syjn99: Nope, for bootnodes config, we should use ENR as it is more descriptive.

varun-doshi: Context for this @unnawut : https://github.com/ReamLabs/ream/pull/690#issuecomment-3172964196
multiaddr is supported just so that we can test locally...it is not expected to be used when we actually run interop

crates/networking/p2p/resources/lean_peers.yaml
varun-doshi: got it...i've cleared the file now

crates/networking/p2p/src/bootnodes.rs
```diff
@@ -55,6 +72,31 @@ impl Bootnodes {
...
+                static_peers.extend(enrs.iter().cloned());
+                to_multiaddrs(&static_peers)
+            }
+            Bootnodes::Multiaddr(multiaddrs) => {
+                let mut static_peers: Vec<Multiaddr> =
+                    serde_yaml::from_str(include_str!("../resources/lean_peers.yaml"))
+                        .expect("should deserialize static lean peers");
+                static_peers.extend(multiaddrs.clone());
+                static_peers
+            }
```
KolbyML:
```suggestion
    pub fn to_multiaddrs_lean(&self) -> Vec<Multiaddr> {
        match self {
            Bootnodes::Default => {
                serde_yaml::from_str(include_str!("../resources/lean_peers.yaml"))
                    .expect("should deserialize static lean peers")
            }
            Bootnodes::None => vec![],
            Bootnodes::Custom(enrs) => {
                to_multiaddrs(&enrs)
            }
            Bootnodes::Multiaddr(multiaddrs) => {
                multiaddrs
            }
```

if we are using custom peers don't we want to use the custom peers? As this seems like an intentional choice, so I think we should only pull the default peers file if the default option is selected

crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,36 +137,189 @@ impl LeanNetworkService {
...
+        Ok(LeanNetworkService {
+            lean_chain,
+            network_config,
+            swarm,
+            peer_table: parking_lot::RwLock::new(HashMap::new()),
+        })
     }
 
-    pub async fn start(mut self) -> anyhow::Result<()> {
+    pub async fn start(&mut self, peer_config: Bootnodes) -> anyhow::Result<()> {
```
KolbyML: why is the arguement called peer_config, if we are passing in Bootnodes?

crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,36 +137,189 @@ impl LeanNetworkService {
                 .build()
         };
 
-        Ok(LeanNetworkService { lean_chain, swarm })
+        Ok(LeanNetworkService {
+            lean_chain,
+            network_config,
+            swarm,
+            peer_table: parking_lot::RwLock::new(HashMap::new()),
```
KolbyML:
```suggestion
            peer_table: RwLock::new(HashMap::new()),
```

can't we remove the qualifer

syjn99:
@varun-doshi CI is now fixed, could you resolve some conflicts?

crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,36 +137,189 @@ impl LeanNetworkService {
...
+            if let Err(err) = self.swarm.dial(peer.clone()) {
+                warn!("Failed to dial peer: {err:?}");
+            } else {
+                successfully_dialed = true;
+            }
+
+            if !successfully_dialed {
+                trace!("Failed to dial any multiaddr for peer: {peer:?}");
+                continue;
+            }
```
KolbyML:
```suggestion
            if let Err(err) = self.swarm.dial(peer.clone()) {
                warn!("Failed to dial peer: {err:?}");
                continue;
            }
```

^ it looks like this was just copy pasted without reading the code, I cleaned it up

crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,36 +137,189 @@ impl LeanNetworkService {
...
+
+            if let Some(Protocol::P2p(peer_id)) =
+                peer.iter().find(|p| matches!(p, Protocol::P2p(_)))
+            {
+                info!("Dialing peer: {peer_id:?}",);
+                self.peer_table
+                    .write()
+                    .insert(peer_id, ConnectionState::Connecting);
+            }
+        }
```
KolbyML: Could you explain the purpose of this if statement to me, I am confused on what it is for

crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,36 +137,189 @@ impl LeanNetworkService {
...
+        multi_addr.push(Protocol::Tcp(self.network_config.socket_port));
+
+        match self.swarm.listen_on(multi_addr.clone()) {
+            Ok(listener_id) => {
+                info!("Listening on {multi_addr:?}: {listener_id:?} ");
+            }
+            Err(err) => {
+                error!("Failed to start libp2p peer listen on {multi_addr:?}, error: {err:?}",);
+            }
+        }
```
KolbyML: if this errors it should fail and end the program no? so maybe move this code to new and add some doc comments

KolbyML (commented):
Here is some feedback

crates/networking/p2p/src/network/lean.rs
varun-doshi: there are two RwLock's in this file used in different places...so importing both causes a conflict

crates/networking/p2p/src/network/lean.rs
varun-doshi: it fetches the peer id from the multiaddr since we need that specifically to insert into the table
My thinking was its possible a multiaddr is missing the p2p section i.e. this is a valid multiaddr as well `/ip4/127.0.0.1/tcp/9000`

crates/networking/p2p/src/network/lean.rs
varun-doshi: I'll move this to `new` but wouldn't this already stop the program as it panics on receieving error from the `start` function?
```
let network_future = executor.spawn(async move {
        if let Err(err) = network_service.start(config.bootnodes).await {
            panic!("Network service exited with error: {err}");
        }
    });
```

Cargo.toml
```diff
@@ -109,6 +109,7 @@ tokio = { version = "1.42", features = ["rt", "rt-multi-thread", "sync", "signal
 tokio-util = { version = "0.7", features = ["compat"] }
 tracing = "0.1"
 tracing-subscriber = "0.3"
+tracing-test = "0.2.5"
```
syjn99, line 112: I think this should be in `[dev-dependencies]` section at `crates/networking/p2p/Cargo.toml` as it is not for production dependency.

crates/networking/p2p/src/network/lean.rs
KolbyML: just do `use parking_lot::RwLock as ParkingRwLock;`

crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,36 +137,187 @@ impl LeanNetworkService {
...
+                    if let Some(event) = self.parse_lean_swarm_event(event).await {
                         info!("Swarm event: {event:?}");
                     }
                 }
             }
         }
     }
 
-    async fn parse_swarm_event(
+    async fn parse_lean_swarm_event(
```
KolbyML:
```suggestion
    async fn parse_swarm_event(
```

Why why lean added to this function name? it seems redundent here

crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,36 +137,187 @@ impl LeanNetworkService {
...
+        ));
+
+        match lean_network_service.swarm.listen_on(multi_addr.clone()) {
+            Ok(listener_id) => {
+                info!("Listening on {multi_addr:?}: {listener_id:?} ");
+            }
+            Err(err) => {
+                error!("Failed to start libp2p peer listen on {multi_addr:?}, error: {err:?}",);
+            }
+        }
```
KolbyML:
```suggestion
        lean_network_service.swarm.listen_on(multi_addr.clone())?;
```

We should be forwarding the error up, not just logging it. If we can't listen on the port the node won't work. If `listen_on()` error's Ream should close with an error. If you want you can use map_err

crates/networking/p2p/src/network/lean.rs
KolbyML: > I'll move this to `new` but wouldn't this already stop the program as it panics on receieving error from the `start` function?
> 
> ```
> let network_future = executor.spawn(async move {
>         if let Err(err) = network_service.start(config.bootnodes).await {
>             panic!("Network service exited with error: {err}");
>         }
>     });
> ```


``error!()`` logs an error to the terminal, it doesn't return an error and end the function

crates/networking/p2p/src/network/lean.rs
KolbyML: oh ok

crates/networking/p2p/src/network/lean.rs
```diff
@@ -128,15 +138,36 @@ impl LeanNetworkService {
                 .build()
         };
 
-        Ok(LeanNetworkService { lean_chain, swarm })
+        let mut lean_network_service = LeanNetworkService {
+            lean_chain,
+            network_config,
+            swarm,
+            peer_table: parking_lot::RwLock::new(HashMap::new()),
```
KolbyML: you forgot to update this

PR #691, opened 2025-08-11
feat: implement initial gossipsub p2p code with block and vote

Kayden-ML:
### What was wrong?

Implements first half of this issue: https://github.com/ReamLabs/ream/issues/671

### How was it fixed?

I Implemented the gossipsub p2p code with the block and vote

PR #693, opened 2025-08-11
fix: calls correct network spec function and imports correct structs

Kayden-ML:
### What was wrong?

Fixes issue mentioned by @varun-doshi  in my previous PR: https://github.com/ReamLabs/ream/pull/691

### How was it fixed?

Imported correct lean network function and struct

### DIFF
--- a/crates/networking/p2p/src/gossipsub/lean/configurations.rs
+++ b/crates/networking/p2p/src/gossipsub/lean/configurations.rs
@@ -2,7 +2,7 @@ use std::time::Duration;
 
 use libp2p::gossipsub::{Config, ConfigBuilder, MessageId, ValidationMode};
 use ream_consensus_misc::constants::beacon::SLOTS_PER_EPOCH;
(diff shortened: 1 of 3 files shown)


crates/networking/p2p/src/gossipsub/lean/configurations.rs
```diff
@@ -31,7 +31,7 @@ impl Default for LeanGossipsubConfig {
             .history_gossip(3)
             .max_messages_per_rpc(Some(500))
             .duplicate_cache_time(Duration::from_secs(
-                SLOTS_PER_EPOCH * beacon_network_spec().seconds_per_slot * 2,
+                SLOTS_PER_EPOCH * lean_network_spec().seconds_per_slot * 2,
```
syjn99, line 34: I don't think we need `SLOTS_PER_EPOCH` here. I just found this, so could you fix this one in the other PR?

syjn99 (approved):

PR #698, opened 2025-08-13
feat: initialize lean rpc

### LINKED ISSUE #684: Add RPC service for API to the lean node
### Describe the feature

https://github.com/ReamLabs/ream/tree/master/crates/rpc/lean

We need a way to fetch the current head, state, etc. There is already a crate for it, so we only have to add a service for it. 

https://github.com/ReamLabs/ream/blob/b69906450d44dfd2c655194547d18172bbcc61b4/crates/common/chain/lean/src/lean_chain.rs#L27

I think fetching current head would be the best point to start.

### Additional context

_No response_

varun-doshi:
### What was wrong?

Fixes #684 
### How was it fixed?

- Initializes and runs a http future for the RPC service on lean node.
- Currently has one endpoint to fetch the current head.
- Renames some folders to distinguish between beacon types and lean types.
- Currently endpoints are `v0` to keep in line with how beacon apis also have a version number on them

### How to run

Run lean node using:
```
cargo run --release lean_node --network ./bin/ream/assets/lean/sample_spec.yml
```

Make GET request to: `http://127.0.0.1:5052/lean/v0/head`


Cargo.toml
```diff
@@ -115,7 +116,7 @@ url = "2.5"
 
 # ream dependencies
 ream-account-manager = { path = "crates/account_manager" }
-ream-beacon-api-types = { path = "crates/common/beacon_api_types" }
+ream-beacon-api-types = { path = "crates/common/api_types/beacon" }
```
syjn99:
```suggestion
ream-api-types-beacon = { path = "crates/common/api_types/beacon" }
```

I think this fits with our convention

crates/common/api_types/lean/Cargo.toml
syjn99, line 1: I don't think we don't need all dependencies (guessing copy-pasted from `crate/common/api_types/beacon/Cargo.toml`). Please keep it as small as possible.

Same comment for `crates/rpc/lean/Cargo.toml`.

bin/ream/src/main.rs
```diff
@@ -175,6 +182,9 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
         }
     });
 
+    let http_future =
+        executor.spawn(async move { start_lean_server(server_config, lean_chain).await });
+
```
syjn99: Haha sorry for nitpicking but could you please remove the empty line above this line so that task spawning seems consolidated? Thanks in advance

crates/rpc/lean/src/config.rs
```diff
@@ -0,0 +1,16 @@
+use std::net::{IpAddr, SocketAddr};
+
+#[derive(Debug, Clone)]
+pub struct LeanRpcServerConfig {
```
syjn99, line 4: I think we don't have to maintain two identical `RpcServerConfig`. Let's refactor this later on. Also, `start_server` and `start_lean_server` have identical responsibility, this can be refactored using builder pattern.

crates/rpc/lean/src/handlers/head.rs
```diff
@@ -0,0 +1,18 @@
...
+use ream_chain_lean::lean_chain::LeanChain;
+use ream_lean_api_types::head::Head;
+use tokio::sync::RwLock;
+
+// GET /lean/v0/head
+#[get("/head")]
+pub async fn get_head(
+    lean_chain: Data<Arc<RwLock<LeanChain>>>,
+) -> Result<impl Responder, ApiError> {
+    println!("got request");
```
syjn99: Gotcha

syjn99 (changes requested):
Some comments here. BTW it works well, nice!

crates/common/api_types/lean/Cargo.toml
varun-doshi: interesting how this was not caught on my local `make pr`

crates/rpc/lean/src/config.rs
varun-doshi: Created an issue for it https://github.com/ReamLabs/ream/issues/699

crates/rpc/lean/src/lib.rs
```diff
@@ -1,14 +1,55 @@
...
 }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
+#[allow(dead_code)]
+impl StopHandle {
+    pub(crate) fn register(&self, handle: ServerHandle) {
+        *self.inner.lock() = Some(handle);
+    }
```
KolbyML: if we aren't using this code can we remove it, that goes for the beacon copy as well, also code shouldn't be duplicated, a 3rd crate should be made to avoid overlaping code. but in this case the StopHandle can just be removed as it isn't used, or if you want the code can be used

crates/rpc/lean/src/lib.rs
```diff
@@ -1,14 +1,55 @@
...
 
-    #[test]
-    fn it_works() {
-        let result = add(2, 2);
-        assert_eq!(result, 4);
+    pub(crate) fn stop(&self, graceful: bool) {
+        #[allow(clippy::let_underscore_future)]
+        let _ = self.inner.lock().as_ref().unwrap().stop(graceful);
```
KolbyML: something really seems wrong here if we need ``#[allow(clippy::let_underscore_future)]``, I think that clippy allowed should be removed, but if you decide to delete StopHandle altogether that is good too

varun-doshi: Sure

KolbyML (approved):
:shipit: PR looks great

Issue #658, opened 2025-07-25
Perf update `ethereum_hashing` to only use sha2 for zkVM's as it is super slow on MacOS, didn't test x86? would be good to know if sha2 should be used on x86 or not


varun-doshi:
### Describe the feature

Use ring in `ethereum_hashing` instead of sha2 if not using a zkVM

syjn99:
And it feels like integrating a new version of `ethereum_hashing` is quite hard.. We might need a way to decouple `ring` dependency by some Rust feature flags. Currently `ethereum_hashing` always includes `ring` as its dependency.

unnawut:
> [@unnawut](https://github.com/unnawut) [@varun-doshi](https://github.com/varun-doshi) Actually this shouldn't be closed as we have to update our dependency in `Cargo.lock` file.

Right sorry ser 🙇‍♂️

KolbyML:
I fixed the issue in https://github.com/ReamLabs/ethereum_hashing/pull/6 The way you avoid including `ring` as a dependency is by excluding using 

```
![target.'cfg(not(target_os = "zkvm"))'.dependencies]
ring = "0.17"
```

This makes it so ring is only a dependency for non-zkvm Operating systems

PR #700, opened 2025-08-14
perf: bump ethereum_hashing to c14a9ad96436e6b8b631faf49a3666e4558b71ae

KolbyML:
### What was wrong?

We removed ring from ethereum_hashing to make our crates compile to zkvm's which caused ethereum_hashing to only support sha2 as a backend which is extremely slow on Macos https://github.com/ReamLabs/ream/issues/658

### How was it fixed?

bump ethereum_hashing to the new version with the fix

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2139,8 +2139,18 @@ dependencies = [
 [[package]]
 name = "ethereum_hashing"
 version = "0.7.0"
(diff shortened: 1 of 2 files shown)


shariqnaiyer (approved):

Kayden-ML (approved):
looks good 👍

PR #701, opened 2025-08-14
feat: add Checkpoint container and match with spec

syjn99:
### What was wrong?

 

`Checkpoint` container is missing, which is likely to be included in the spec.

### How was it fixed?

 

This PR adds `Checkpoint` container and matches with the current (draft) spec. (https://github.com/leanEthereum/leanSpec/pull/8)

### To-Do

 
### DIFF
--- a/crates/common/consensus/lean/src/lib.rs
+++ b/crates/common/consensus/lean/src/lib.rs
@@ -1,4 +1,5 @@
 pub mod block;
+pub mod checkpoint;
 pub mod config;
 pub mod state;
 pub mod vote;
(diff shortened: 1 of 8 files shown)


crates/common/consensus/lean/src/lib.rs
```diff
@@ -195,21 +196,16 @@ pub fn get_fork_choice_head(
...
+            .iter()
+            .max_by_key(|child_hash| {
+                let vote_weight = vote_weights.get(*child_hash).unwrap_or(&0);
+                let slot = blocks.get(*child_hash).map(|block| block.slot).unwrap_or(0);
+                (*vote_weight, slot, *(*child_hash))
+            })
+            .ok_or_else(|| anyhow!("No children found for current root: {current_root}"))?;
     }
+
+    Ok(current_root)
```
syjn99, line 210: Note: I found my linter keeps complaining for this part, so I rewrite this more idiomatically

PR #708, opened 2025-08-16
chore: update README to Lean chain from Beam chain

jveer634:
### What was wrong?

Update beam chain to lean chain (added consensus when needed)

### To-Do

Check the title on the Video attached. Do we change that too ?

 
### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,11 @@ One of our goals is to build an ecosystem where developers can seamlessly build 
 
 The Lean Chain is the next generation of Ethereum Consensus, incorporating all of the greatest and latest ideas from the Ethereum research roadmap. Its goal is to transition quickly and safely from the Beacon Chain to a consensus layer design much closer to the final design of Ethereum.
 
-[Video of the announcement of the Beam Chain with more information](https://youtu.be/lRqnFrqpq4k?si=YODLo0MBgkVMblmE)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,19 +5,19 @@
 
 ## What is Ream
 
-Ream aims to be modular, contributor-friendly, and blazingly fast implementation of the Beam Chain specification.
+Ream aims to be modular, contributor-friendly, and blazingly fast implementation of the Lean Chain Consensus specification.
```
KolbyML:
```suggestion
Ream aims to be modular, contributor-friendly, and blazingly fast implementation of the Lean Consensus specification.
```

etc etc

jveer634: The Lean Chain is the next generation of Ethereum, incorporating all of the greatest and latest ideas from the Ethereum research roadmap. Its goal is to transition quickly and safely from the Beacon Chain to a consensus layer design much closer to the final design of Ethereum.

Is this ok ?

KolbyML: Oh I am more so poking at is instead of the `Beam Chain` it was renamed to `Lean Consensus` there is no longer the `chain` part to my knowledge

jveer634: I pulled up somethings from the Justin's blog as well. I think that fits the description.

PR #709, opened 2025-08-16
feat: base code for lean metrics

### LINKED ISSUE #687: Setup base for lean metrics
## Criteria:

1. The client reliably sends telemetries defined below to a prometheus instance
2. Since the standards are not yet agreed, let's keep it to the minimal of 1 trace and 1 metrics to have the minimal working infra while avoiding excessive rework later
    - Trace: [`LeanChain::propose_block()`](https://github.com/ReamLabs/ream/blob/master/crates/common/chain/lean/src/lean_chain.rs#L98-L146)
    - Metrics: [`state.latest_justified_slot`](https://github.com/ReamLabs/ream/blob/master/crates/common/consensus/lean/src/state.rs#L19)
3. Each telemetry should also be labeled with the `validator_id`
4. The telemetries should follow [Trace semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/trace/) and [Metrics semantic convention](https://opentelemetry.io/docs/specs/semconv/general/metrics/) respectively

### Additional context

_No response_

unnawut:
### What was wrong?

Set up the minimum codebase for preparing metrics for prometheus. Closes #687. 

### How was it fixed?

- Reused [prometheus and grafana code from trin](https://github.com/ethereum/trin/tree/master/bin/trin-execution/metrics) per https://github.com/ReamLabs/ream/issues/687#issuecomment-3167006049
- Reused [instrumenting code from trin](https://github.com/ethereum/trin/blob/master/bin/trin-execution/src/metrics.rs) per https://github.com/ReamLabs/ream/issues/687#issuecomment-3167006049
- Added these metrics:
  - `lean_propose_block_time` (histogram)
      - Also broken down to sub-metrics: `initialize_block`, `add_valid_votes_to_block`, `compute_state_root`, `compute_block_root`
  - `lean_head_slot` (gauge)
  - `lean_justified_slot` (gauge)
  - `lean_finalized_slot` (gauge)
- Created a basic grafana dashboard to visualize the metrics
- Added docker compose and override files for ease of testing locally

### Usage

You need to start prometheus & grafana services, then start the lean node service with the metrics flag. For example:

1. Start prometheus & grafana locally by enabling the docker compose overrides:
    ```bash
    cp compose.override-example.yml compose.override.yml
    ```

2. Run prometheus & grafana on docker compose:
    ```bash
    docker compose up
    ````

    Prometheus should then be listening by default at `localhost:9090`, and grafana at `localhost:3000`

3. Start a lean node with `--metrics`:

    ```bash
    cargo run --release lean_node --network ephemery --metrics
    ```
    
    Metrics should be exposed at `localhost:8080`.

4. Enjoy grafana at http://localhost:3000 with the default username/password of admin/admin. The included dashboard should be available by default.

<figure>
Screenshot 2025-08-16 at 23 39 30
<figcaption>Yes my example halted midway but that's why observability is useful :p</figcaption>
</figure>

### To-Do


crates/common/metrics/lean/compose.yml
```diff
@@ -0,0 +1,27 @@
...
+      - ./tools/grafana/datasource.yaml:/etc/grafana/provisioning/datasources/datasource.yaml
+      - ./tools/grafana/dashboard.yaml:/etc/grafana/provisioning/dashboards/main.yaml
+      - ./tools/grafana/dashboards:/var/lib/grafana/dashboards
+    ports:
+      - 3000:3000
+
+volumes:
+  prometheus_data:
+
+networks: {}
```
KolbyML: please move the code being copy/pasted to `ream/metrics` it shouldn't be in this crate. 

image

^ also I would match this folder structure instead of making a tools folder, as that folder would be redundent, once this isn't no longer in the crate.

KolbyML (changes requested):
Mostly looks good, just move the Metrics docker compose to it's own folder `ream/metrics` and remove the `tools` folder. As the grafana/promethus setup logic I wrote doesn't make sense to be in that crate.

crates/common/metrics/lean/compose.yml
unnawut: Fixed! Thanks!

.gitignore
```diff
@@ -27,3 +27,6 @@ testing/ef-tests/mainnet/*
 
 # DS_Store: Desktop Services Store
 .DS_Store
+
+# Docker compose override files
+compose.override.yml
```
KolbyML: remove this too

crates/common/metrics/lean/src/helpers.rs
```diff
@@ -0,0 +1,28 @@
...
+        .expect("failed to create histogram")
+}
+
+pub fn start_timer_vec(histogram_vec: &HistogramVec, label_values: &[&str]) -> HistogramTimer {
+    histogram_vec.with_label_values(label_values).start_timer()
+}
+
+pub fn stop_timer(timer: HistogramTimer) {
+    timer.observe_duration()
+}
```
KolbyML: these might as well stay in lib.rs, the name `helpers.rs` isn't the most descriptive I doubt lib.rs will ever get better then 100-200 lines or so, so I think it is fine in this case to leave it in lib.rs

.gitignore
unnawut: `compose.override.yml` is for customizing the docker compose to each person's preference, so I think it makes sense to be ignored. If something in the override file needs to be shared across the team then it should go into `compose.yml`

crates/common/metrics/lean/src/helpers.rs
unnawut: Hmm I want to counter-argue a bit that those are the setup functions, but this part is provisioning code so I'd prefer to separate them:

```rs
    pub static ref PROPOSE_BLOCK_TIME: HistogramVec = create_histogram_vec(
        "lean_propose_block_time",
        "Duration of the sections it takes to propose a new block",
        &["section"]
    );
    // ...
```

not big deal but I think they should be separate

.gitignore
KolbyML: sure keep the ignore, and remove the added files then

crates/common/metrics/lean/src/helpers.rs
KolbyML: Place the provisioning code above the functions 

- mod's
- imports
- static/consts
- functions/structs/impls

.gitignore
unnawut: Ah thanks for spotting! I committed the file before adding ignore

crates/common/metrics/lean/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+        &[]
+    );
+}
+
+use prometheus_exporter::prometheus::{
+    HistogramTimer, HistogramVec, IntGaugeVec, default_registry,
+    register_histogram_vec_with_registry, register_int_gauge_vec_with_registry,
+};
+
+pub fn create_int_gauge_vec(name: &str, help: &str, label_names: &[&str]) -> IntGaugeVec {
```
KolbyML:
```suggestion
use prometheus_exporter::prometheus::{
    HistogramTimer, HistogramVec, IntGaugeVec, default_registry,
    register_histogram_vec_with_registry, register_int_gauge_vec_with_registry,
};

// Provisioning each metrics
lazy_static::lazy_static! {
    pub static ref PROPOSE_BLOCK_TIME: HistogramVec = create_histogram_vec(
        "lean_propose_block_time",
        "Duration of the sections it takes to propose a new block",
        &["section"]
    );

    pub static ref HEAD_SLOT: IntGaugeVec = create_int_gauge_vec(
        "lean_head_slot",
        "The current head slot",
        &[]
    );

    pub static ref JUSTIFIED_SLOT: IntGaugeVec = create_int_gauge_vec(
        "lean_justified_slot",
        "The current justified slot",
        &[]
    );

    pub static ref FINALIZED_SLOT: IntGaugeVec = create_int_gauge_vec(
        "lean_finalized_slot",
        "The current finalized slot",
        &[]
    );
}

pub fn create_int_gauge_vec(name: &str, help: &str, label_names: &[&str]) -> IntGaugeVec {
```

Shouldn't it be like this?

crates/common/metrics/lean/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "ream-metrics-lean"
+authors.workspace = true
```
KolbyML: shouldn't we just have `ream-metrics`? I don't think we should have a dedicated crate to distinguish against `beacon`/`lean` metrics, as there isn't much to distinguish them in this specific case as metrics are just a universal utility for us, so 1 crate makes sense here

PR #710, opened 2025-08-17
fix: genesis time bug

varun-doshi:
### What was wrong?

This [commit](https://github.com/ReamLabs/ream/commit/b790d433987b50cd4239cb012fe5270d591208a8) introduced a bug in the codebase. Essentially the piece of code below was removed which caused the `lean_node` to crash on startup
https://github.com/ReamLabs/ream/blob/8b4854a08f1df498e72a7a89ed38e87c9e8b4b97/bin/ream/src/main.rs#L121C1-L134C1
It caused `genesis time` to be 0(which should never happen in the context of `Unix` time)

### How was it fixed?

This PR re-adds the above logic to the `set_lean_network_spec` function to ensure correct interval calculation
 
To test, ensure the lean node runs successfully:
```
cargo run --release lean_node --network ./bin/ream/assets/lean/sample_spec.yml
```


varun-doshi:
Update: After discussion with @syjn99 we will be keeping the "0 is invalid gen time" flow. Prefer to use `--network ephemery` flag and only use config file path workflow when needed.
This improves the error logging and displays a more descriptive error:
```
Chain service exited with error: Failed to create clock interval: Genesis time is 0 but should be greater than 1755439178:
```

syjn99 (approved):
LGTM

PR #713, opened 2025-08-17
feat: add quic support for lean

### LINKED ISSUE #697: Support QUIC as default
### Describe the feature

We [aims QUIC](https://github.com/leanEthereum/pm/blob/ab5b1653fbd954671df97824a1b1ac59db53f347/pq-interop/pq-devnet-0.md#active-discussions) as our primary network protocol in upcoming devnets. Current codebase doesn't support QUIC at all.

### Additional context

Currently blocked by #690.

varun-doshi:
### What was wrong?

Fixes #697 

### How was it fixed?

- Adds `Quic` support for the `lean_node`
- Nodes can now start in Quic mode by passing `--enable-quic`. If not, by default it runs in TCP mode.
- Nodes can dial and connect with both TCP/QUIC nodes.
- Tests are added for the following cases:
  - `tcp lean node <-> tcp lean node` : 
  - `quic lean node <-> quic lean node`
  - `tcp lean node <-> quic lean node`

### How to run

To run in quic node:
```
cargo run --release lean_node --network ephemery --enable-quic
```

Example of tcp address:
```
/ip4/127.0.0.1/tcp/9000/p2p/16Uiu2HAmLsfh4CHNEKYx42NGLcVJn5zxpSjnqJ7PsCrUrU6ugAvu
```

Example of quic address:
```
/ip4/127.0.0.1/udp/9003/quic-v1/p2p/16Uiu2HAmLsfh4CHNEKYx42NGLcVJn5zxpSjnqJ7PsCrUrU6ugAvu
```

### DIFF
--- a/bin/ream/src/cli/lean_node.rs
+++ b/bin/ream/src/cli/lean_node.rs
@@ -43,4 +43,7 @@ pub struct LeanNodeConfig {
 
     #[arg(long, default_value_t = DEFAULT_HTTP_ALLOW_ORIGIN)]
     pub http_allow_origin: bool,
+
+    #[arg(long, help = "Enable QUIC")]
+    pub enable_quic: bool,
 }
--- a/bin/ream/src/main.rs
+++ b/bin/ream/src/main.rs
@@ -135,6 +135,7 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
         }),
         lean_chain.clone(),
         executor.clone(),
+        config.enable_quic,
     )
     .await
     .expect("Failed to create network service");
--- a/book/cli/ream/lean_node.md
+++ b/book/cli/ream/lean_node.md
@@ -25,6 +25,8 @@ Options:
           Set HTTP Port [default: 5052]
       --http-allow-origin
 
+      --enable-quic
+          Enable QUIC
   -h, --help
           Print help
 ```
--- a/crates/networking/p2p/src/network/lean.rs
+++ b/crates/networking/p2p/src/network/lean.rs
@@ -75,6 +75,7 @@ impl LeanNetworkService {
         network_config: Arc<LeanNetworkConfig>,
         lean_chain: Arc<RwLock<LeanChain>>,
         executor: ReamExecutor,
+        enable_quic: bool,
     ) -> anyhow::Result<Self> {
         let connection_limits = {
             let limits = ConnectionLimits::default()
@@ -117,7 +118,7 @@ impl LeanNetworkService {
             }
         };
 
-        let transport = build_transport(local_key.clone())
+        let transport = build_transport(local_key.clone(), enable_quic)
             .map_err(|err| anyhow!("Failed to build transport: {err:?}"))?;
 
         let swarm = {
@@ -146,9 +147,17 @@ impl LeanNetworkService {
         };
 
         let mut multi_addr: Multiaddr = lean_network_service.network_config.socket_address.into();
-        multi_addr.push(Protocol::Tcp(
-            lean_network_service.network_config.socket_port,
-        ));
+        if enable_quic {
+            multi_addr.push(Protocol::Udp(
+                lean_network_service.network_config.socket_port,
+            ));
+            multi_addr.push(Protocol::QuicV1);
+        } else {
+            multi_addr.push(Protocol::Tcp(
+                lean_network_service.network_config.socket_port,
+            ));
+        }
+        info!("Listening on {multi_addr:?}");
 
         lean_network_service
             .swarm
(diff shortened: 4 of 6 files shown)


book/cli/ream/lean_node.md
```diff
@@ -25,6 +25,8 @@ Options:
           Set HTTP Port [default: 5052]
       --http-allow-origin
 
+      --enable-quic
+          Enable QUIC
```
unnawut: Can we enable quic by default for lean and remove this flag?

crates/networking/p2p/src/network/misc.rs
```diff
@@ -40,7 +44,20 @@ pub fn build_transport(local_private_key: Keypair) -> io::Result<Boxed<(PeerId,
...
+        let transport = tcp
+            .or_transport(quic)
+            .map(|either_output, _| match either_output {
+                Either::Left((peer_id, muxer)) => (peer_id, StreamMuxerBox::new(muxer)),
+                Either::Right((peer_id, muxer)) => (peer_id, StreamMuxerBox::new(muxer)),
+            });
+        transport.boxed()
+    } else {
+        tcp.boxed()
+    };
```
syjn99: Does this code intended to support **both** TCP and QUIC? I guess we don't have to (according to the current consensus on PQ Devnet spec), and also at `LeanNetworkService::new` function we **specify** multiaddr for QUIC, which means `Either::Left` case won't be used that often.

So I think the issue (#697) would be more descriptive: Replace current TCP transport into QUIC. What do you think?

crates/networking/p2p/src/network/lean.rs
```diff
@@ -252,40 +261,144 @@ mod tests {
...
 
-        let config2 = Arc::new(LeanNetworkConfig {
-            gossipsub_config: LeanGossipsubConfig::default(),
-            socket_address: Ipv4Addr::new(127, 0, 0, 1).into(),
-            socket_port: 9001,
+    // Test to check connection between 2 TCP lean nodes
+    #[tokio::test]
+    #[traced_test]
+    async fn test_two_tcp_lean_nodes_connection() -> anyhow::Result<()> {
+        let socket_port1 = 9000;
```
syjn99: I think it is fine at this phase, but this looks annoying to find an empty port and hard-code it in each test code. I don't have much knowledge on it, is there any alternative that you looked for?

bin/ream/src/cli/lean_node.rs
```diff
@@ -43,4 +43,7 @@ pub struct LeanNodeConfig {
 
     #[arg(long, default_value_t = DEFAULT_HTTP_ALLOW_ORIGIN)]
     pub http_allow_origin: bool,
+
+    #[arg(long, help = "Enable QUIC")]
+    pub enable_quic: bool,
```
syjn99: If we would replace TCP to QUIC, I think this flag isn't needed.

book/cli/ream/lean_node.md
syjn99: Oh I found this comment quite late, actually I raised [similar concern](https://github.com/ReamLabs/ream/pull/713#discussion_r2281359251).

varun-doshi: Just confirming that we dont want to support tcp at all for devnet0(as per spec too it seems thats what is intended)
all nodes run on Quic only

crates/networking/p2p/src/network/misc.rs
varun-doshi: Yup understood; we dont want any nodes running on tcp

crates/networking/p2p/src/network/lean.rs
varun-doshi: Yeah it should be possible but I was facing some issues trying to get that to work so its manual now
I think if this takes time to figure out we can push this to a separate issue

crates/networking/p2p/src/network/lean.rs
```diff
@@ -75,6 +75,7 @@ impl LeanNetworkService {
         network_config: Arc<LeanNetworkConfig>,
         lean_chain: Arc<RwLock<LeanChain>>,
         executor: ReamExecutor,
+        enable_quic: bool,
```
syjn99: I think we don't need this param

crates/networking/p2p/src/network/misc.rs
```diff
@@ -27,7 +27,10 @@ impl libp2p::swarm::Executor for Executor {
     }
 }
 
-pub fn build_transport(local_private_key: Keypair) -> io::Result<Boxed<(PeerId, StreamMuxerBox)>> {
+pub fn build_transport(
```
syjn99: hmm I think we don't have to stick into this function. Adding QUIC support in this function feels too overkill.

Building `Swarm` would be enough for this code block:

```rust
let swarm = {
    let config = Config::with_executor(Executor(executor))
        .with_notify_handler_buffer_size(NonZeroUsize::new(7).expect("Not zero"))
        .with_per_connection_event_buffer_size(4)
        .with_dial_concurrency_factor(NonZeroU8::new(1).unwrap());

    SwarmBuilder::with_existing_identity(local_key.clone())
        .with_tokio()
        .with_quic()
        .with_behaviour(|_| behaviour)?
        .with_swarm_config(|_| config)
        .build()
};
```

which means we don't have to rely on `build_transport` function.

bin/ream/src/main.rs
syjn99, line 1: Could you please change the `socket_port` description in `LeanNodeConfig` in `bin/ream/src/cli/lean_node.rs`? After merging this PR, we won't use TCP as the transport layer.

PR #715, opened 2025-08-18
feat: add peer rpc for lean

### LINKED ISSUE #714: Peer RPC for Lean
### Describe the feature

Implement RPC endpoints for peers in lean node. 

### Additional context

Few endpoints for peers could be:
- fetch state of peer connections
- fetch all connected peers

These are not new endpoints per se. Beacon api has a few peer related endpoints as well
Since pq devnet is going to be heavily focused on interop we should be able to check the state of peer connections on our node. Moreover there is no way to check connected peers other than scanning the logs to see when we received a SwarmEvent which is inconvenient especially when we are connecting to a large peer set

varun-doshi:
### What was wrong?

Fixes #714 
### How was it fixed?

- Adds RPC support for fetching peer data for `lean_node`
- Currently 2 endpoints have been added:
  - `/lean/v0/node/peers`: to get list of all peers
  - `/lean/v0/node/peer_count`: to get connection state
- Tried and tested with local nodes

### DIFF
--- a/bin/ream/src/main.rs
+++ b/bin/ream/src/main.rs
@@ -180,6 +180,8 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
     .await
     .expect("Failed to create network service");
 
+    let peer_table = network_service.peer_table();
+
     let keystores = load_validator_registry(&config.validator_registry_path)
         .expect("Failed to load validator registry");
     let validator_service =
@@ -207,8 +209,8 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
             panic!("Validator service exited with error: {err}");
         }
     });
-    let http_future =
-        executor.spawn(async move { start_lean_server(server_config, lean_chain_reader).await });
+    let http_future = executor
+        .spawn(async move { start_lean_server(server_config, lean_chain_reader, peer_table).await });
 
     tokio::select! {
         _ = chain_future => {
--- a/bin/ream/src/main.rs
+++ b/bin/ream/src/main.rs
@@ -209,8 +209,9 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
             panic!("Validator service exited with error: {err}");
         }
     });
-    let http_future = executor
-        .spawn(async move { start_lean_server(server_config, lean_chain_reader, peer_table).await });
+    let http_future = executor.spawn(async move {
+        start_lean_server(server_config, lean_chain_reader, peer_table).await
+    });
 
     tokio::select! {
         _ = chain_future => {
--- a/crates/rpc/lean/src/handlers/peer.rs
+++ b/crates/rpc/lean/src/handlers/peer.rs
@@ -2,42 +2,34 @@ use std::{collections::HashMap, sync::Arc};
 
 use actix_web::{HttpResponse, Responder, get, web::Data};
 use libp2p::PeerId;
-use parking_lot::RwLock;
+use parking_lot::Mutex;
 use ream_api_types_beacon::error::ApiError;
-use ream_p2p::peer::ConnectionState;
+use ream_p2p::network::peer::ConnectionState;
 use ream_rpc_beacon::handlers::peers::PeerCount;
 
 // /lean/v0/node/peers
 #[get("/node/peers")]
 pub async fn list_peers(
-    peer_table: Data<Arc<RwLock<HashMap<PeerId, ConnectionState>>>>,
+    peer_table: Data<Arc<Mutex<HashMap<PeerId, ConnectionState>>>>,
 ) -> Result<impl Responder, ApiError> {
-    Ok(HttpResponse::Ok().json(peer_table.read().clone()))
+    Ok(HttpResponse::Ok().json(peer_table.lock().clone()))
 }
 
-// /lean/v0/node/peers
+// /lean/v0/node/peer_count
 #[get("/node/peer_count")]
 pub async fn get_peer_count(
-    peer_table: Data<Arc<RwLock<HashMap<PeerId, ConnectionState>>>>,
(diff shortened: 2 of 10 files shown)


crates/rpc/lean/src/handlers/peer.rs
```diff
@@ -0,0 +1,43 @@
...
+
+// /lean/v0/node/peers
+#[get("/node/peers")]
+pub async fn list_peers(
+    peer_table: Data<Arc<RwLock<HashMap<PeerId, ConnectionState>>>>,
+) -> Result<impl Responder, ApiError> {
+    Ok(HttpResponse::Ok().json(peer_table.read().clone()))
+}
+
+// /lean/v0/node/peers
```
syjn99: This comments seems to be inconsistent with the corresponding handler

bin/ream/src/main.rs
```diff
@@ -139,6 +139,8 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
     .await
     .expect("Failed to create network service");
 
+    let peer_table = network_service.peer_table();
```
syjn99, line 183: At first I didn't expect `peer_table` will be shared between tokio tasks, but if we want to use in the async context I think we might use `tokio::sync::RwLock` instead of the one at `parking_lot`.

KolbyML: We shouldn't use `tokio::sync::RwLock` unless it is needed, and in this case I don't think it is. Just because we can use async code doesn't mean we should etc etc.

syjn99: @KolbyML Yeah I studied a bit more about it, and also concluded to use `parking_lot`. It seems we don't have to add async complexity in this phase. Thanks!

@varun-doshi apologizes for your additional work 🙏

KolbyML: There are also timing considerations. parking_lot::RWLock should only be used if there is 0 delay in accessing the resource. I.e. the resource being accessed is in memory. Quite a few cases a Mutex can be straight up faster then a RwLock

crates/networking/p2p/src/network/lean.rs
```diff
@@ -253,16 +253,16 @@ impl LeanNetworkService {
                 None
             }
             SwarmEvent::ConnectionEstablished { peer_id, .. } => {
-                self.peer_table
-                    .write()
-                    .insert(peer_id, ConnectionState::Connected);
+                let mut table = self.peer_table.lock();
+
+                table.insert(peer_id, ConnectionState::Connected);
```
KolbyML: can you inline this

crates/networking/p2p/src/network/lean.rs
```diff
@@ -253,16 +253,16 @@ impl LeanNetworkService {
...
                 info!("Connected to peer: {peer_id:?}");
                 None
             }
             SwarmEvent::ConnectionClosed { peer_id, .. } => {
-                self.peer_table
-                    .write()
-                    .insert(peer_id, ConnectionState::Disconnected);
+                let mut table = self.peer_table.lock();
+
+                table.insert(peer_id, ConnectionState::Disconnected);
```
KolbyML: inline

crates/networking/p2p/src/network/lean.rs
```diff
@@ -295,13 +295,17 @@ impl LeanNetworkService {
                 .find(|protocol| matches!(protocol, Protocol::P2p(_)))
             {
                 info!("Dialing peer: {peer_id:?}",);
-                self.peer_table
-                    .write()
-                    .insert(peer_id, ConnectionState::Connecting);
+                let mut table = self.peer_table.lock();
+
+                table.insert(peer_id, ConnectionState::Connecting);
```
KolbyML: inline

crates/rpc/lean/src/handlers/peer.rs
```diff
@@ -0,0 +1,44 @@
...
+use parking_lot::Mutex;
+use ream_api_types_beacon::error::ApiError;
+use ream_p2p::peer::ConnectionState;
+use ream_rpc_beacon::handlers::peers::PeerCount;
+use tokio::sync::RwLock;
+
+// /lean/v0/node/peers
+#[get("/node/peers")]
+pub async fn list_peers(
+    peer_table: Data<Arc<RwLock<HashMap<PeerId, ConnectionState>>>>,
```
KolbyML: this won't work, as the underlying data structure being passed in is a Mutex now, not a Rwlock

crates/rpc/lean/src/handlers/peer.rs
```diff
@@ -0,0 +1,44 @@
...
+) -> Result<impl Responder, ApiError> {
+    Ok(HttpResponse::Ok().json(peer_table.read().await.clone()))
+}
+
+// /lean/v0/node/peer_count
+#[get("/node/peer_count")]
+pub async fn get_peer_count(
+    peer_table: Data<Arc<Mutex<HashMap<PeerId, ConnectionState>>>>,
```
KolbyML, line 21: same as above

crates/rpc/lean/src/handlers/peer.rs
```diff
@@ -0,0 +1,44 @@
...
+            ConnectionState::Disconnecting => disconnecting += 1,
+        }
+    }
+
+    Ok(HttpResponse::Ok().json(&PeerCount {
+        connected,
+        connecting,
+        disconnected,
+        disconnecting,
+    }))
```
KolbyML: implement default for PeerCount using an auto derive, then you can just modify PeerCount directly

crates/rpc/lean/src/handlers/peer.rs
varun-doshi: Not sure I understand this one...the underlying data is Mutex and this takes in a Mutex as well

Issue #635, opened 2025-07-07
Test gossiped beacon blocks


varun-doshi:
### Describe the feature

Add tests for all cases in validate gossiped Beacon Block
~Blocked by #618~

SWASTIC-7:
@varun-doshi is it still blocked? Else I would love to take this up.

SWASTIC-7:
Ok @varun-doshi I will add those cases too

PR #721, opened 2025-08-20
added test to cover other cases in validate_beacon_block

SWASTIC-7:
### What was wrong?

Test to cover different cases for validate_beacon_block function was needed #635
 
### How was it fixed?

Added more tests to check other cases
- [x] Test future slot block
- [x] Test block is not from future slot
- [x] Test validator not found
- [x] Testing duplicate proposer signature
- [x] Testing bls to execution duplicate change


syjn99:
@SWASTIC-7 Thanks for your contribution!

CI is failing, could you run `make pr` for checking formatting, linting, etc.? Also we follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for PR title, so could you also change the PR title?

PR #722, opened 2025-08-20
test: added more test for validate_beacon_block 

SWASTIC-7:
### What was wrong?

Test to cover different cases for validate_beacon_block function was needed #635
 
### How was it fixed?

Added more tests to check other cases
- [x] Test future slot block
- [x] Test block is not from future slot
- [x] Test validator not found
- [x] Testing duplicate proposer signature
- [x] Testing bls to execution duplicate change

added some tests, would add some more test (covering positive side of validate_beacon_block fn) in new pr

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,8 @@ update-book-cli: build-debug ## Update book cli documentation.
 
 .PHONY: test
 test: # Run all tests.
-	cargo test --workspace -- --nocapture
+	cargo test --workspace --exclude ream-network-manager -- --nocapture
+	cargo test -p ream-network-manager -- --nocapture --test-threads=1
 
 clean-deps:
 	cargo +nightly udeps --workspace --tests --all-targets --release --exclude ef-tests
--- a/Makefile
+++ b/Makefile
@@ -56,8 +56,7 @@ update-book-cli: build-debug ## Update book cli documentation.
 
 .PHONY: test
 test: # Run all tests.
-	cargo test --workspace --exclude ream-network-manager -- --nocapture
-	cargo test -p ream-network-manager -- --nocapture --test-threads=1
+	cargo test --workspace -- --nocapture
 
 clean-deps:
 	cargo +nightly udeps --workspace --tests --all-targets --release --exclude ef-tests
--- a/testing/gossip-validation/Cargo.toml
+++ b/testing/gossip-validation/Cargo.toml
@@ -28,5 +28,5 @@ ream-network-spec.workspace = true
 ream-operation-pool.workspace = true
 ream-storage.workspace = true
 
-[dev-dependencies]  
-serial_test = "3.2"
+[dev-dependencies]
+tempdir = "0.3.7" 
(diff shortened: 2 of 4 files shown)


Makefile
syjn99, line 1: Wondering why does this file change.

Makefile
```diff
@@ -56,7 +56,8 @@ update-book-cli: build-debug ## Update book cli documentation.
 
 .PHONY: test
 test: # Run all tests.
-	cargo test --workspace -- --nocapture
+	cargo test --workspace --exclude ream-network-manager -- --nocapture
+	cargo test -p ream-network-manager -- --nocapture --test-threads=1
```
KolbyML: if you want certain tests to be done serially using https://crates.io/crates/serial_test

KolbyML: Wait you are using serial :skull:, Jun wonders why this changed, but there is 0 reasons why this should have changed. I am assuming this is a left over experiment

KolbyML (changes requested):

Makefile
SWASTIC-7: Actually I want to run my testfile as single thread,as parallel execution making initialize_test_network_spec() into race condition

Makefile
SWASTIC-7: Though i added serial, still running from workspace would run that in parallel

SWASTIC-7: is there a better way? I can fix that

KolbyML: There isn't a race condition

```
    pub async fn db_setup() -> (BeaconChain, CachedDB, B256) {
        let temp = std::path::PathBuf::from("ream_gossip_test");
        fs::create_dir_all(&temp).unwrap();
        let mut db = ReamDB::new(temp).unwrap();
```

^ this code is just wrong. It uses a static directory for tests, instead of a temporary directory for each test case

Update this code to use https://docs.rs/tempdir/latest/tempdir/ there are examples through our codebase as well I believe

KolbyML: Please remove the serial crate and the Makefile change, after this bug is fixed :saluting_face:

SWASTIC-7:
@KolbyML made the changes please check

testing/gossip-validation/Cargo.toml
```diff
@@ -27,3 +27,6 @@ ream-network-manager = { workspace = true, features = ["disable_ancestor_validat
 ream-network-spec.workspace = true
 ream-operation-pool.workspace = true
 ream-storage.workspace = true
+
+[dev-dependencies]
+tempdir = "0.3.7"
```
KolbyML:
```suggestion
tempdir.workspace = true
```

tempdir is already defined in our workspace, so we shouldn't be redefining the depends definition

SWASTIC-7: sorry i didn't notice that, making the fix

varun-doshi:
> added some tests, would add some more test (covering positive side of validate_beacon_block fn) in new pr

Note: For the positive case, we also want to test the actual networking functionality i.e. gossip. These tests only pick a static file and check if they are invalid which is fine. But we also want a positive case where the block is introduced in the pipeline through actual gossip

Issue #728, opened 2025-08-23
Validator shouldn't be able to directly modify `LeanChain`


KolbyML:
### Describe the feature

I think it should be `validator` -> `LeanChainService` <-> `P2P`. I don't think the Validator should be able to modify, or directly access P2P. The validator can request stuff from the LeanChainService with callbacks

PR #729, opened 2025-08-23
feat: remove mutating LeanChain from Validator, add exclusive writer multi reader primitive

KolbyML:
### What was wrong?

fixes https://github.com/ReamLabs/ream/issues/728

The validator was modifying the LeanChain directly, which is quite a bit of a overreach for it. Also there was comments in the Validator saying it would directly access P2P.

I don't think 
- the Validator should be able to modify the LeanChain
- the validator shouldn't be able to directly interact with P2P

### How was it fixed?

Add `Writer`/`Reader` primitives so that different parts of the codebase can access the `LeanChain`, but not modify it if they aren't supposed to. Like the `Validator` it is the `LeanChainServices`'s job to modify the LeanChain, it is the `Validator`'s job to propose and vote on new blocks.

I modified the code so only the `LeanChainService` can modify the LeanChain, but we can have global readers of the LeanChain

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5229,6 +5229,7 @@ dependencies = [
  "ream-rpc-beacon",
  "ream-rpc-lean",
  "ream-storage",
+ "ream-sync",
  "ream-validator-beacon",
  "ream-validator-lean",
  "tokio",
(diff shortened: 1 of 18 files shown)


Pato234:
Great

Issue #655, opened 2025-07-23
Implement an interface crate for PQ cryptography


syjn99:
### Describe the feature

Integrate [`b-wagn/hash-sig`](https://github.com/b-wagn/hash-sig) with the current codebase.

PR #732, opened 2025-08-24
feat: implement interface over hash sig crate

shariqnaiyer:
### What was wrong?

Fixes: #655 

### How was it fixed?

I studied the implementation of the hash sig crate and the signature scheme traits and reviewed our BLS crate implementation. 

I structured the pqc crate similar to the hash sig crate. 

I found we were using an old version of the hash sig crate that exposed methods with different parameters for key gen. I imported the latest version which led to some conflicts in account manager. I believe account manager will require some changes after this PR. Me or Partha could take up these changes. I commented out some code to avoid dealing with dependency conflicts in Partha's code.

I don't think we necessarily need to define traits as there is only one implementation but I have added it for consistency with ream-bls. I can remove the traits if deemed appropriate.

### To-Do

 
### DIFF
--- a/crates/crypto/pqc/src/hashsig/private_key.rs
+++ b/crates/crypto/pqc/src/hashsig/private_key.rs
@@ -1,15 +1,10 @@
-use hashsig::{
-    MESSAGE_LENGTH,
-    signature::{
-        SignatureScheme,
-        generalized_xmss::instantiations_poseidon::lifetime_2_to_the_18::winternitz::SIGWinternitzLifetime18W4,
-    },
-};
+use hashsig::{MESSAGE_LENGTH, signature::SignatureScheme};
 use rand::Rng;
 
-use crate::hashsig::{errors::SigningError, public_key::PublicKey, signature::Signature};
+use crate::hashsig::{
+    HashSigScheme, errors::SigningError, public_key::PublicKey, signature::Signature,
+};
 
-pub type HashSigScheme = SIGWinternitzLifetime18W4;
 pub type HashSigPrivateKey = <HashSigScheme as SignatureScheme>::SecretKey;
 
 pub struct PrivateKey {
--- a/crates/crypto/pqc/src/hashsig/mod.rs
+++ b/crates/crypto/pqc/src/hashsig/mod.rs
@@ -1,4 +1,8 @@
+use hashsig::signature::generalized_xmss::instantiations_poseidon::lifetime_2_to_the_18::winternitz::SIGWinternitzLifetime18W4;
+
 pub mod errors;
 pub mod private_key;
 pub mod public_key;
 pub mod signature;
+
+pub type HashSigScheme = SIGWinternitzLifetime18W4;
--- a/crates/crypto/pqc/src/hashsig/public_key.rs
+++ b/crates/crypto/pqc/src/hashsig/public_key.rs
@@ -1,6 +1,6 @@
 use hashsig::signature::SignatureScheme;
 
-use crate::hashsig::private_key::HashSigScheme;
+use crate::hashsig::HashSigScheme;
 
 type HashSigPublicKey = <HashSigScheme as SignatureScheme>::PublicKey;
 
--- a/crates/crypto/pqc/src/hashsig/signature.rs
+++ b/crates/crypto/pqc/src/hashsig/signature.rs
@@ -1,6 +1,6 @@
 use hashsig::{MESSAGE_LENGTH, signature::SignatureScheme};
 
-use crate::hashsig::{private_key::HashSigScheme, public_key::PublicKey};
+use crate::hashsig::{HashSigScheme, public_key::PublicKey};
 
 type HashSigSignature = <HashSigScheme as SignatureScheme>::Signature;
 


bin/ream/src/cli/account_manager.rs
```diff
@@ -24,6 +21,14 @@ pub struct AccountManagerConfig {
...
     #[arg(short, long)]
     pub seed_phrase: Option<String>,
+
+    /// Activation epoch for the validator
+    #[arg(long, default_value_t = 0)]
+    pub activation_epoch: usize,
+
+    /// Number of active epochs
+    #[arg(long, default_value_t = 1 << 28)]
+    pub num_active_epochs: usize,
```
KolbyML: image

These default values should be a constants

crates/crypto/pqc/src/traits.rs
syjn99, line 1: I wonder: what's the benefit of having these traits? I mean it isn't bad indeed, but what does this abstraction layer do?

KolbyML: If we don't need the trait, then we probably wouldn't need these wrappers at all tbh

shariqnaiyer: As I said:
> I don't think we necessarily need to define traits as there is only one implementation but I have added it for consistency with ream-bls. I can remove the traits if deemed appropriate.

I did it one for consistency and just incase we have another implementation for these signatures come out. That may be too optimistic tho.

crates/crypto/pqc/src/hashsig/private_key.rs
```diff
@@ -0,0 +1,94 @@
...
+            <HashSigScheme as SignatureScheme>::key_gen(rng, activation_epoch, num_active_epochs);
+
+        (PublicKey::new(public_key), Self::new(private_key))
+    }
+}
+
+impl PQSignable for PrivateKey {
+    type Error = SigningError;
+
+    fn sign(&self, message: &[u8], epoch: u32) -> Result<Signature, Self::Error> {
```
syjn99: If we only want to force the size of message, how about letting compiler know it by defining `&[u8; 32]`? Then the error can be removed

crates/crypto/pqc/src/hashsig/private_key.rs
```diff
@@ -0,0 +1,94 @@
...
+        if message.len() != MESSAGE_LENGTH {
+            return Err(SigningError::InvalidMessageLength(message.len()));
+        }
+
+        Ok(Signature::new(
+            <HashSigScheme as SignatureScheme>::sign(
+                &mut rand::rng(),
```
syjn99: This makes us use only one instance of RNG, which isn't consistent with `generate`. I think this also should be generic.

crates/crypto/pqc/src/traits.rs
syjn99, line 1: Can we use `anyhow::Result` here?

syjn99: I guess the core value of wrapper is able to switch the signature scheme as we want, but it seems the current implementation sticks to one single signature scheme.

shariqnaiyer: This initial implementation uses just the SIGWinternitzLifetime18W4 scheme but I agree we should make it more modular.

crates/crypto/pqc/src/hashsig/private_key.rs
shariqnaiyer: I used &[u8; MESSAGE_LENGTH] for context. Do let me know if that works!

crates/crypto/pqc/src/traits.rs
syjn99:
```rust
// Export the default one
pub type DefaultSignatureScheme = SIGWinternitzLifetime18W4;

// and use it in other places...
let (pk, sk) = DefaultSignatureScheme::key_gen(...);
```

So I think newly added codes in this PR are equivalent to the above type alias. I'm quite sure we won't see other implementation in the near future, so it feels like too overkill to me TBH.

Meanhwhile, I think a wrapper can be helpful for code consistency: `key_gen` -> `generate`, etc. But this also can be done not using traits. 
Or the wrapper can be stateful, that means it tracks which epochs are already used for signing and return `Error` when the user of this crate tries to sign with already used epoch (or path). (Reference: https://github.com/b-wagn/hash-sig/blob/287517a763edba7e518b0c1ee5beb868f26f1f66/src/signature.rs#L19-L22)

Cargo.toml
```diff
@@ -93,8 +93,8 @@ libp2p-mplex = "0.43"
 lru = "0.15.0"
 parking_lot = "0.12.3"
 prometheus_exporter = "0.8.5"
-rand = "0.8"
-rand_chacha = "0.3"
+rand = "0.9"
+rand_chacha = "0.9"
```
syjn99, line 97: Is there any reason for bumping those dependencies? (for my own understanding :))

crates/crypto/pqc/src/hashsig/private_key.rs
```diff
@@ -0,0 +1,80 @@
...
+    signature::{
+        SignatureScheme,
+        generalized_xmss::instantiations_poseidon::lifetime_2_to_the_18::winternitz::SIGWinternitzLifetime18W4,
+    },
+};
+use rand::Rng;
+
+use crate::hashsig::{errors::SigningError, public_key::PublicKey, signature::Signature};
+
+pub type HashSigScheme = SIGWinternitzLifetime18W4;
```
syjn99: hmm just a nit 🙏 but I think this type alias can be located in other places, it is hard to predict `private_key.rs` will have the default signature scheme.

Cargo.toml
shariqnaiyer: They were giving dependency conflicts with the latest hash-sig.  (The old ones)

crates/crypto/pqc/src/hashsig/private_key.rs
shariqnaiyer: I have moved it to mod.rs I think that is appropriate. Only other spot will be signature.rs but let me know if mod.rs works!

PR #737, opened 2025-08-24
refactor: separate beacon/lean ReqResp message logic + implement lean message types + protocol id's

KolbyML:
### What was wrong?

This is the first step in support Lean's Request Response domain. Filling the Foundation so to speak. We didn't implement the lean message types, also I req/resp code was structured with only the beacon chain in mind

### How was it fixed?

move stuff and added the Lean message types 

### Future PR's

make it so we can share a majority of our Request/Response logic so important are seen across both domains, and less code to maintain


Kayden-ML (approved):
Looks good nice work 👍

Issue #730, opened 2025-08-23
Only publish received blocks/votes to p2p if blocks/votes are processed successfully


Kayden-ML:
### Describe the feature

Only publish received blocks/votes to p2p if blocks/votes are processed successfully

PR #738, opened 2025-08-25
feat: Don't propagate lean block/vote if processing failed

Kayden-ML:
### What was wrong?

Fixes: https://github.com/ReamLabs/ream/issues/730

### How was it fixed?

if it processes successfully, it publishes if it fails it logs an error


KolbyML (approved):
:shipit: looks good

PR #739, opened 2025-08-25
feat: add code for Lean ReqResp domain

KolbyML:
### What was wrong?

Our Req/Resp code didn't support Lean

### How was it fixed?

Updating our ReqResp code to support Lean


Kayden-ML (approved):
Looks solid nice work 👍

PR #742, opened 2025-08-25
feat: add `--validator-registry-path` flag for lean node

### LINKED ISSUE #731: Add keystore flag for Lean node
### Describe the feature

So far, we mostly rely on `--ephemery` flag, which is only for **one single node with four validators**. As P2P goals are mostly achieved, I think we can achieve multi-node devnet which is also mentioned at #688.

For multi-node devnet, we need a couple of things to do:

### Pass Keystore from the command.

https://github.com/ReamLabs/ream/blob/5eaa73c625e532fc8bc359c3420b750f8c0cafe7/bin/ream/src/main.rs#L144
https://github.com/ReamLabs/ream/blob/5eaa73c625e532fc8bc359c3420b750f8c0cafe7/bin/ream/src/main.rs#L170-L172

Currently `ValidatorService` isn't aware of local keystore, and the code is just passing empty list of keystore every time. For the first devnet, it is enough for passing its validator ID.

### A script for multi-node devnet

#688 describes this, so let the issue handle this task.

### Additional context

_No response_

syjn99:
### What was wrong?

 

Fixes #731 

### How was it fixed?

 

```yaml
validators:
- validator_id: 0
  keystore_path: /path/to/keystore
- validator_id: 1
  keystore_path: /path/to/keystore
# ...
```

Add `--validator-registry-path` flag for registering a YAML file which validators are managed by the current node. As a keystore would not responsible for the validator index itself, I added `validator_id` entry. `keystore_path` will be ignored, as we don't have a keystore now.

> [!Note]
> For keystore, we might use our existing keystore package to decrypt. Or I think it is still enough to save raw serialized bytes of the private key, and deserialize and derive public key from it. This topic is still TBD.

---

https://github.com/ReamLabs/ream/blob/320a3cea7d50eef0fa1d889ba6792440e1edf13c/crates/common/validator/lean/src/service.rs#L38-L45

Removed hacky code above. If you still want to run an single-node, ephemeral devnet, you can use this command:

```
cargo run --release lean_node \
    --network ephemery \
    --validator-registry-path ./bin/ream/assets/lean/validator_registry.yml
```

Otherwise, I was also testing a multi-node devnet ([Diff](https://github.com/syjn99/ream/pull/2), AI-generated bash scripts, be careful) based on this branch. Then the command will look like:

```
cargo run --release lean_node \
    --network devnet/devnet-spec.yaml \
    --validator-registry-path devnet/node0/registry.yaml
```

### To-Do

 

crates/common/validator/lean/src/registry.rs
```diff
@@ -0,0 +1,43 @@
...
+pub fn load_validator_registry<P: AsRef<Path>>(path: P) -> anyhow::Result<Vec<LeanKeystore>> {
+    let content = fs::read_to_string(&path).map_err(|err| {
+        anyhow::anyhow!(
+            "Failed to read validator registry file {:?}: {err}",
+            path.as_ref(),
+        )
+    })?;
+
+    let registry: ValidatorRegistryYaml = serde_yaml::from_str(&content)
+        .map_err(|err| anyhow::anyhow!("Failed to parse validator registry YAML: {}", err))?;
```
KolbyML:
```suggestion
    let registry = serde_yaml::from_str::<ValidatorRegistryYaml>(&content)
        .map_err(|err| anyhow::anyhow!("Failed to parse validator registry YAML: {err}"))?;
```

PR #749, opened 2025-08-26
fix: send only signed data through p2p network & add more message types for `LeanChainServiceMessage`

### LINKED ISSUE #743: Gossip decoding fails in Lean devnet
### Describe the bug

I [wrote a script](https://github.com/syjn99/ream/pull/2) for bootstrapping and run multi-node devnets based on #742. It seems a node fails to decode a gossiped message for the block.

```log
2025-08-25T06:44:14.002978Z  INFO ream_validator_lean::service: Not proposer for slot 2 (proposer is validator 2), skipping
2025-08-25T06:44:14.004893Z  WARN ream_p2p::network::lean: gossip decode failed: InvalidData("Failed to decode ssz: OffsetIntoFixedPortion(2)")
2025-08-25T06:44:15.002872Z  INFO ream_validator_lean::service: Starting vote phase at slot 2 (tick 1): 1 validator(s) voting
```

I guess this is because of discrepancy of 1) what we expect from gossipsub and 2) what we actually publish to the network.

### Gossipsub message = `SignedBlock`

https://github.com/ReamLabs/ream/blob/320a3cea7d50eef0fa1d889ba6792440e1edf13c/crates/networking/p2p/src/gossipsub/lean/message.rs#L8-L12

### What we actually publish = `Block`

https://github.com/ReamLabs/ream/blob/320a3cea7d50eef0fa1d889ba6792440e1edf13c/crates/networking/p2p/src/network/lean/mod.rs#L200-L213

https://github.com/ReamLabs/ream/blob/320a3cea7d50eef0fa1d889ba6792440e1edf13c/crates/networking/p2p/src/network/lean/mod.rs#L302-L306

### Steps to reproduce

- Use devnet scripts setup: https://github.com/syjn99/ream/pull/2
- Run `cargo build --release`
- Run `./devnet/start_devnet.sh`

### Platform(s)

_No response_

### What version/commit are you on?

`ream fe092b3-fe092b3 darwin-aarch64 rustc1.91.0-nightly`

### If you've built Ream from source, provide the full command you used

_No response_

### Code of Conduct

- [x] I agree to follow the Code of Conduct

syjn99:
### What was wrong?

 

Fixes #743.

Also this PR includes new arms `LeanChainMessage` for enum. Currently, we mostly use `QueueItem` for passing the block and vote, which is originated designed for the member of [`dependencies` hash map](https://github.com/ethereum/research/blob/d225a6775a9b184b5c1fd6c830cc58a375d9535f/3sf-mini/p2p.py#L31). As we now have `SignedBlock` and `SignedVote`, this `QueueItem` doesn't fit so well for our P2P environment as all messages on the gossipsub should be signed.

#### Major changes

- Remove `QueueItem` message in `LeanChainMessage`, and add two separate messages: `ProcessBlock` and `ProcessVote`. 
  -  `is_trusted` was added for distinguishing the source of the message.
  - `need_gossip` was added because the node will "enqueue" the blocks and votes if it cannot be processed right now, and then consume the queue when it's available.
- Fix the type of the message of channel for `outbound_p2p_*`: As Kolby [mentioned](https://github.com/ReamLabs/ream/pull/749#discussion_r2299935727), this can be extended later on for Req/Resp. This PR only adds `LeanGossipRequest` enum, which only contain the **signed** block and vote.
- Log the consensus information for every slot.

---

#### Common setup

I relied on our metrics setup, which can be easily bootstrapped with following command:

```bash
$ cd metrics && docker compose up
```

By default, Prometheus will target to `localhost:8080`.

#### Ephemery

image

- One single node, with 4 validators on.
- No P2P.
- Run a single process with single command line:

```bash
cargo run --release lean_node \
    --network ephemery \
    --validator-registry-path ./bin/ream/assets/lean/validator_registry.yml \
    --metrics
```

#### Local Devnet

image

- Four nodes with 1 validator on.
- Each node is aware of other parties through `--bootnodes` flag.
- Consensus is achieved through P2P.

This devnet is relied on [my script for spinning up a local devnet](https://github.com/syjn99/ream/pull/2), including:

- Generate `genesis_spec.yml` in ad-hoc way.
- Fetch each peer ID for `--bootnodes` setup.
- Run four processes in the background, and connect `stdout` stream into log files.

Running devnet:

```bash
$ ./devnet/start_devnet.sh
```

Stopping it:

```bash
$ ./devnet/stop_devnet.sh
```

### To-Do

 
### DIFF
--- a/bin/ream/src/main.rs
+++ b/bin/ream/src/main.rs
@@ -18,8 +18,8 @@ use ream::cli::{
 };
 use ream_api_types_beacon::id::{ID, ValidatorID};
 use ream_chain_lean::{
-    genesis as lean_genesis, gossip_request::LeanGossipRequest, lean_chain::LeanChain,
-    messages::LeanChainServiceMessage, service::LeanChainService,
+    genesis as lean_genesis, lean_chain::LeanChain, messages::LeanChainServiceMessage,
+    p2p_request::LeanP2PRequest, service::LeanChainService,
 };
 use ream_checkpoint_sync::initialize_db_from_checkpoint;
 use ream_consensus_misc::{
@@ -139,8 +139,7 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
 
     // Initialize the services that will run in the lean node.
     let (chain_sender, chain_receiver) = mpsc::unbounded_channel::<LeanChainServiceMessage>();
-    let (outbound_p2p_sender, outbound_p2p_receiver) =
-        mpsc::unbounded_channel::<LeanGossipRequest>();
+    let (outbound_p2p_sender, outbound_p2p_receiver) = mpsc::unbounded_channel::<LeanP2PRequest>();
 
     let chain_service = LeanChainService::new(
         lean_chain_writer,
--- a/crates/common/chain/lean/src/service.rs
+++ b/crates/common/chain/lean/src/service.rs
@@ -14,8 +14,8 @@ use tracing::{error, info, warn};
 use tree_hash::TreeHash;
 
 use crate::{
-    clock::create_lean_clock_interval, gossip_request::LeanGossipRequest,
-    lean_chain::LeanChainWriter, messages::LeanChainServiceMessage, queue_item::QueueItem,
+    clock::create_lean_clock_interval, lean_chain::LeanChainWriter,
+    messages::LeanChainServiceMessage, p2p_request::LeanP2PRequest, queue_item::QueueItem,
     slot::get_current_slot,
 };
 
@@ -28,7 +28,7 @@ pub struct LeanChainService {
     lean_chain: LeanChainWriter,
     receiver: mpsc::UnboundedReceiver<LeanChainServiceMessage>,
     sender: mpsc::UnboundedSender<LeanChainServiceMessage>,
-    outbound_gossip: mpsc::UnboundedSender<LeanGossipRequest>,
+    outbound_gossip: mpsc::UnboundedSender<LeanP2PRequest>,
     // Objects that we will process once we have processed their parents
     dependencies: HashMap<B256, Vec<QueueItem>>,
 }
@@ -38,7 +38,7 @@ impl LeanChainService {
         lean_chain: LeanChainWriter,
         receiver: mpsc::UnboundedReceiver<LeanChainServiceMessage>,
         sender: mpsc::UnboundedSender<LeanChainServiceMessage>,
-        outbound_gossip: mpsc::UnboundedSender<LeanGossipRequest>,
+        outbound_gossip: mpsc::UnboundedSender<LeanP2PRequest>,
     ) -> Self {
         LeanChainService {
             lean_chain,
@@ -108,7 +108,7 @@ impl LeanChainService {
                                 warn!("Failed to handle process block message: {err}");
                             }
 
-                            if need_gossip && let Err(err) = self.outbound_gossip.send(LeanGossipRequest::Block(signed_block)) {
+                            if need_gossip && let Err(err) = self.outbound_gossip.send(LeanP2PRequest::GossipBlock(signed_block)) {
                                 warn!("Failed to send item to outbound gossip channel: {err}");
                             }
                         }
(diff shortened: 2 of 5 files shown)


bin/ream/src/main.rs
```diff
@@ -140,8 +138,9 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
         Writer::new(LeanChain::new(genesis_block, genesis_state));
 
     // Initialize the services that will run in the lean node.
-    let (chain_sender, chain_receiver) = mpsc::unbounded_channel::<LeanChainServiceMessage>();
-    let (outbound_p2p_sender, outbound_p2p_receiver) = mpsc::unbounded_channel::<QueueItem>();
+    let (chain_sender, chain_receiver) = mpsc::unbounded_channel::<LeanChainMessage>();
+    let (outbound_p2p_sender, outbound_p2p_receiver) =
+        mpsc::unbounded_channel::<LeanGossipRequest>();
```
KolbyML: This channel sends request's to the P2P layer, it isn't specific to gossipsub, why wouldn't we use it for the requests/response domain as well etc.

syjn99: Yes for sure, but current codebase only sends `QueueItem` which is specific to the `LeanChain` itself. The type of the message can be extended later.

syjn99:
@KolbyML 
Almost ready right now, tested a couple of times for 1) single-node with 4 validators / 2) four independent nodes with one validator each. I'm gonna improve documentation from now, so feel free to review this!

crates/common/chain/lean/src/gossip_request.rs
```diff
@@ -0,0 +1,7 @@
+use ream_consensus_lean::{block::SignedBlock, vote::SignedVote};
+
+#[derive(Debug, Clone)]
+pub enum LeanGossipRequest {
+    Block(SignedBlock),
+    Vote(SignedVote),
+}
```
KolbyML:
```suggestion
pub enum LeanP2PRequest {
    GossipBlock(SignedBlock),
    GossipVote(SignedVote),
}
```

Shouldn't we do this, then it always for sending other types of requests to the p2p service

crates/common/chain/lean/src/service.rs
```diff
@@ -208,13 +231,16 @@ impl LeanChainService {
...
+    async fn handle_process_vote(
+        &mut self,
+        signed_vote: SignedVote,
+        is_trusted: bool,
+    ) -> anyhow::Result<()> {
+        let vote = if is_trusted {
+            signed_vote.data
+        } else {
+            // TODO: Validate the signature.
+            signed_vote.data
```
```suggestion
        if !is_trusted {
            // TODO: Validate the signature.
        }
        let vote = signed_vote.data;
```

The if else seems redundant now

crates/common/chain/lean/src/messages.rs
```diff
@@ -1,33 +1,23 @@
...
-}
-
-impl ProduceBlockMessage {
-    pub fn new(slot: u64, response: oneshot::Sender<Block>) -> Self {
-        ProduceBlockMessage { slot, response }
-    }
+pub enum LeanChainMessage {
+    ProduceBlock {
+        slot: u64,
+        response: oneshot::Sender<Block>,
```
KolbyML:
```suggestion
        sender: oneshot::Sender<Block>,
```

I know I original put `response`, but what do you think about `sender` or `callback` instead?

crates/common/chain/lean/src/service.rs
```diff
@@ -120,47 +143,24 @@ impl LeanChainService {
...
+        &mut self,
+        signed_block: SignedBlock,
+        is_trusted: bool,
+    ) -> anyhow::Result<()> {
+        let block = if is_trusted {
+            signed_block.message
+        } else {
+            // TODO: Validate the signature.
+            signed_block.message
+        };
```
```suggestion
        if !is_trusted {
            / TODO: Validate the signature.
        };
        let block = signed_block.message;
```

KolbyML (commented):
Mostly looks good just some minor comments

crates/common/chain/lean/src/messages.rs
syjn99: `sender` looks good

crates/common/chain/lean/src/gossip_request.rs
syjn99: Then I think this enum should be defined under our p2p crate. I will move it while resolving this issue!

KolbyML (approved):
:shipit: looks good and nice testing!

PR #757, opened 2025-08-31
feat: Initial version of a key generation utility for ream

ch4r10t33r:
### What was wrong?

- Generation of a key pair for each message_type to be signed.
- Automatic storing of the generated keystore in the default paths.
- Partial implementation of the keystore file as per the new EIP (https://github.com/ReamLabs/EIPs/blob/010eb5698ccb94d12cc6b551aa61d7f02cb9b49a/EIPS/eip-new.md)

### To-Do

 

bin/ream/src/main.rs
```diff
@@ -380,13 +398,55 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
...
-        &seed_phrase,
-        config.activation_epoch,
-        config.num_active_epochs,
-    );
+
+    // Parallelize key generation for each message type
+    let message_types: Vec<MessageType> = MessageType::iter().collect();
+
+    message_types
+        .par_iter()
```
syjn99: FYI) `key_gen` function is [already built on](https://github.com/b-wagn/hash-sig/blob/287517a763edba7e518b0c1ee5beb868f26f1f66/src/signature/generalized_xmss.rs#L123) parallelism.

ch4r10t33r: Yes, I am aware. This is me generating multiple keys in parallel (one for each message type). The PR is still WIP though.

.gitignore
```diff
@@ -27,3 +27,6 @@ testing/ef-tests/mainnet/*
 
 # DS_Store: Desktop Services Store
 .DS_Store
+
+# keystore
+.keystore
```
syjn99: Please add newline here.

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,327 @@
+use chrono::{DateTime, Utc};
+use hex;
+use rand;
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct QsKeystore {
```
syjn99: How about `QuantumSecureKeystore`? Is it too long? I'm not sure `Qs` is intuitive enough for inferring its meaning.

crates/common/account_manager/src/lib.rs
```diff
@@ -1,27 +1,51 @@
...
 pub fn generate_keys(
     seed_phrase: &str,
+    wallet_index: u32,
     activation_epoch: usize,
     num_active_epochs: usize,
+    passphrase: &str,
 ) -> (PublicKey, PrivateKey) {
     info!(
-        "Generating lean consensus validator keys  with activation_epoch={activation_epoch}, num_active_epochs={num_active_epochs}....."
+        "Generating lean consensus validator keys for index {} with activation_epoch={activation_epoch}, num_active_epochs={num_active_epochs}.....",
```
syjn99:
```suggestion
        "Generating lean consensus validator keys for index {wallet_index} with activation_epoch={activation_epoch}, num_active_epochs={num_active_epochs}.....",
```

Same as all logging lines like `info!` (will not comment those for not messing up the comment section)

bin/ream/src/main.rs
```diff
@@ -366,7 +369,22 @@ pub async fn run_validator_node(config: ValidatorNodeConfig, executor: ReamExecu
...
+        ║    ┌─ Generating Keys ─┐                              ║
+        ║    │ ████████████████ │ ▓▓▓ Cryptographic Keys        ║
+        ║    │ ████████████████ │ ▓▓▓ Message Attestation       ║
+        ║    └──────────────────┘                               ║
+        ║                                                       ║
+        ║    🔐 Secure • 🔑 Deterministic • 📝 Multi-Purpose    ║
+        ╚═══════════════════════════════════════════════════════╝
+        
+    "#
+    );
```
syjn99: What does this printing mean? Why are there white boxes in it? TBH I don't understand the full box.

bin/ream/src/cli/account_manager.rs
```diff
@@ -60,14 +77,32 @@ impl AccountManagerConfig {
...
     }
 
     pub fn get_seed_phrase(&self) -> String {
         if let Some(phrase) = &self.seed_phrase {
             phrase.clone()
         } else {
-            "default_seed_phrase".to_string()
+            // Generate a new BIP39 mnemonic with 24 words (256 bits of entropy)
+            let entropy: [u8; 32] = rand::random();
+            let mnemonic = Mnemonic::from_entropy_in(bip39::Language::English, &entropy).unwrap();
```
syjn99: Can you use `expect()`?

crates/common/account_manager/src/message_types.rs
```diff
@@ -0,0 +1,22 @@
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum MessageType {
+    Attestation,
+    Blocks,
```
syjn99:
```suggestion
    Block,
```

Can this be singular? And also could you change the file name into `message_type.rs`?

bin/ream/src/cli/account_manager.rs
syjn99: Also just another nit here: can we just use `Mnemonic::from_entropy` so that we don't have to specify the language? That will also remove the explicit dependency of `bip39` in `./bin/ream/Cargo.toml`.

bin/ream/src/cli/account_manager.rs
```diff
@@ -24,13 +26,25 @@ pub struct AccountManagerConfig {
...
     /// Activation epoch for the validator
     #[arg(long, default_value_t = DEFAULT_ACTIVATION_EPOCH)]
     pub activation_epoch: usize,
 
     /// Number of active epochs
     #[arg(long, default_value_t = DEFAULT_NUM_ACTIVE_EPOCHS)]
     pub num_active_epochs: usize,
+
+    /// Path for keystore file
+    #[arg(long, default_value = "./.keystore/")]
```
syjn99: I think `./.keystore/` should be a constant so that it can be referenced in `main.rs` also.

bin/ream/src/main.rs
```diff
@@ -380,17 +398,58 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
...
+            let keystore = QsKeystore::from_seed_phrase(
+                &seed_phrase,
+                config.lifetime as u32,
+                config.activation_epoch as u32,
+                Some(format!("Ream validator keystore for {:?}", message_type)),
+                Some(format!("m/44'/60'/0'/0/{}", index)),
+            );
+
+            // Write keystore to file with enum name
+            let filename = format!("{:?}.json", message_type);
```
syjn99: You can just implement `std::fmt::Display` trait to `MessageType` so that not using `Debug` trait here.

bin/ream/src/main.rs
```diff
@@ -380,17 +398,58 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
...
+                index as u32,
+                config.activation_epoch,
+                config.num_active_epochs,
+                config.passphrase.as_deref().unwrap_or(""),
+            );
+
+            // Create keystore file using QsKeystore
+            let keystore = QsKeystore::from_seed_phrase(
+                &seed_phrase,
+                config.lifetime as u32,
```
syjn99: If `lifetime` should be under `2**32`, can we just define it as `u32`? Same comment for other parameters.

bin/ream/src/main.rs
```diff
@@ -380,17 +398,58 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
...
+                config.activation_epoch as u32,
+                Some(format!("Ream validator keystore for {:?}", message_type)),
+                Some(format!("m/44'/60'/0'/0/{}", index)),
```
syjn99: This path is quite random, right? It's just for generating EOA (as `60` indicates). If you want to put it as a placeholder, I guess there's at least a comment for this.

bin/ream/src/main.rs
```diff
@@ -380,17 +398,58 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
...
+
+            // Write keystore to file with enum name
+            let filename = format!("{:?}.json", message_type);
+            let keystore_file_path = keystore_dir.join(filename);
+            let keystore_json = keystore.to_json().expect("Failed to serialize keystore");
+
+            fs::write(&keystore_file_path, keystore_json).expect("Failed to write keystore file");
+
+            info!("Keystore written to: {:?}", keystore_file_path);
+            // info!("Public key for {:?}: {:?}", message_type, public_key.inner.to_string());
```
syjn99: Please delete this line if it is not for usage.

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,327 @@
...
+
+        let mut keystore = Self::new(crypto, keytype, uuid);
+        keystore.description = description;
+        keystore.path = path;
+
+        keystore
+    }
+
+    /// Validate the keystore structure
+    pub fn validate(&self) -> Result<(), String> {
```
syjn99: Can you make all values into a constant so that we can find it better after this PR? (e.g., version, argon2id, aes-256-gcm, ...)

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,327 @@
...
+        // Generate random nonce for AES-GCM (12 bytes)
+        let nonce = hex::encode(rand::random::<[u8; 12]>());
+
+        // Generate random tag for AES-GCM (16 bytes)
+        let tag = hex::encode(rand::random::<[u8; 16]>());
+
+        // Store the seed phrase as encrypted data (hex encoded)
+        let ciphertext = hex::encode(seed_phrase.as_bytes());
+
+        let kdf = KdfParams::new_full(65536, 3, 4, salt);
```
syjn99: Are `iterations` and `parallelism` different with your EIP document? (What I see is `iterations = 4` and `parallelism = 2`.)

bin/ream/Cargo.toml
```diff
@@ -17,17 +17,23 @@ path = "src/main.rs"
 [dependencies]
 alloy-primitives.workspace = true
 anyhow.workspace = true
+bip39.workspace = true
+chrono.workspace = true
```
syjn99: We don't need `chrono` here.

bin/ream/Cargo.toml
```diff
@@ -17,17 +17,23 @@ path = "src/main.rs"
...
 rand_chacha.workspace = true
+rayon.workspace = true
+serde.workspace = true
+serde_json.workspace = true
 tokio.workspace = true
 tracing = { workspace = true, features = ["log"] }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 unicode-normalization.workspace = true
 url.workspace = true
+uuid.workspace = true
```
syjn99: We also don't need `uuid` here.

crates/common/account_manager/src/lib.rs
```diff
@@ -1,27 +1,51 @@
...
+        &mut <ChaCha20Rng as SeedableRng>::from_seed(derived_seed),
         activation_epoch,
         num_active_epochs,
-    )
+    );
+
+    // Display public key contents
+    match serde_json::to_string_pretty(&public_key.inner) {
+        Ok(json) => info!("Public key contents: {}", json),
+        Err(_) => info!("Public key generated successfully (could not serialize)"),
```
syjn99: What does this mean? Why aren't newly generated `public_key` serializable?

crates/common/account_manager/src/lib.rs
```diff
@@ -1,27 +1,51 @@
...
+    let seed = mnemonic.to_seed(passphrase);
+
+    // Create a deterministic seed based on the original seed and wallet index
     let mut hasher = Sha256::new();
-    hasher.update(seed_phrase.as_bytes());
-    let seed = hasher.finalize().into();
-    info!("Seed: {seed:?}");
+    hasher.update(seed);
+    hasher.update(wallet_index.to_be_bytes());
+    let derived_seed: [u8; 32] = hasher.finalize().into();
```
syjn99, line 35: TBH this part is hard to understand. It seems there is some arbitrary decisions you made like hashing `seed` with `wallet_index`. Also `passphrase` and `seed_phrase` are too confusing.

These rules shouldn't reside in the code, but at least there needs a document that I can refer to. (Apologizes if the doc exists, but I can't find it in your EIP.)

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,327 @@
...
+            KdfParamsInner::Full { salt, .. } | KdfParamsInner::Short { salt, .. } => {
+                self.validate_hex_string(salt, "salt")?;
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Helper to validate hex strings
+    fn validate_hex_string(&self, hex_str: &str, field_name: &str) -> Result<(), String> {
```
syjn99: This doesn't have to member of `QsKeystore`. It is just a pure util function, right?

syjn99 (commented):
Most of the comments are about cosmetics for codebase consistency (like inlining variable in the logs...), and left some questions that comes into my mind while reading your code. Great job!

crates/common/account_manager/src/keystore.rs
KolbyML: Do we even need the words `QuantumSecure`, as the place we are going all Keystore's will be quantum secure, that is just the normal

crates/common/account_manager/src/lib.rs
```diff
@@ -1,27 +1,51 @@
...
+    let (public_key, private_key) = PrivateKey::generate(
+        &mut <ChaCha20Rng as SeedableRng>::from_seed(derived_seed),
         activation_epoch,
         num_active_epochs,
-    )
+    );
+
+    // Display public key contents
+    match serde_json::to_string_pretty(&public_key.inner) {
+        Ok(json) => info!("Public key contents: {}", json),
```
KolbyML: json should just be inlined

```suggestion
        Ok(json) => info!("Public key contents: {json}"),
```

crates/common/account_manager/src/lib.rs
KolbyML: why are we silencing the error?

crates/common/account_manager/src/keystore.rs
KolbyML: we should be using anyhow, our project should only use this_error or anyhow depending on the context

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,327 @@
...
+        if hex_str.chars().all(|c| c.is_ascii_hexdigit()) {
+            Ok(())
+        } else {
+            Err(format!("{} must be a valid hex string", field_name))
+        }
+    }
```
KolbyML:
```suggestion
    /// Helper to validate hex strings
    fn validate_hex_string(&self, hex_str: &str, field_name: &str) -> anyhow::Result<()> {
        ensure!(hex_str.chars().all(|c| c.is_ascii_hexdigit(), format!("{field_name} must be a valid hex string"));
        Ok(())
    }
```

please always inline variables in strings, also do Jun's suggestion, this doesn't have any logic unique to the keystore

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,327 @@
...
+            return Err("KDF function must be argon2id".to_string());
+        }
+
+        // Check cipher function
+        if self.crypto.cipher.function != "aes-256-gcm" {
+            return Err("Cipher function must be aes-256-gcm".to_string());
+        }
+
+        // Check keytype function
+        if self.keytype.function != "xmss-poisedon2-ots-seed" {
```
KolbyML: "xmss-poisedon2-ots-seed" should be a constant,

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,327 @@
...
+            return Err("quantum_secure must be true".to_string());
+        }
+
+        // Check KDF function
+        if self.crypto.kdf.function != "argon2id" {
+            return Err("KDF function must be argon2id".to_string());
+        }
+
+        // Check cipher function
+        if self.crypto.cipher.function != "aes-256-gcm" {
```
KolbyML: constant

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,327 @@
...
+            return Err("Version must be 5".to_string());
+        }
+
+        // Check quantum_secure flag
+        if !self.quantum_secure {
+            return Err("quantum_secure must be true".to_string());
+        }
+
+        // Check KDF function
+        if self.crypto.kdf.function != "argon2id" {
```
KolbyML: constant

I gave a few examples, there exure me others

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,327 @@
...
+
+        // Check cipher function
+        if self.crypto.cipher.function != "aes-256-gcm" {
+            return Err("Cipher function must be aes-256-gcm".to_string());
+        }
+
+        // Check keytype function
+        if self.keytype.function != "xmss-poisedon2-ots-seed" {
+            return Err("Keytype function must be xmss-poisedon2-ots-seed".to_string());
+        }
```
KolbyML: use ensure!() for all of these

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,327 @@
...
+        t: u32,
+        p: u32,
+        salt: String,
+    },
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CipherParams {
+    /// Cipher function name, must be "aes-256-gcm"
+    pub function: String,
```
KolbyML: this should be type enum not a string

bin/ream/src/main.rs
```diff
@@ -366,7 +369,22 @@ pub async fn run_validator_node(config: ValidatorNodeConfig, executor: ReamExecu
 /// This function initializes the account manager by validating the configuration,
 /// generating keys, and starting the account manager service.
 pub async fn run_account_manager(mut config: AccountManagerConfig) {
-    info!("starting up account manager...");
+    println!(
```
KolbyML: println! should never be used in production code

bin/ream/src/main.rs
```diff
@@ -380,17 +398,58 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
 
     let seed_phrase = config.get_seed_phrase();
 
+    // Create keystore directory if it doesn't exist
+    let default_path = "./.keystore/".to_string();
+    let keystore_path = config.path.as_ref().unwrap_or(&default_path);
+    let keystore_dir = Path::new(keystore_path);
+    if !keystore_dir.exists() {
+        fs::create_dir_all(keystore_dir).expect("Failed to create keystore directory");
+        info!("Created keystore directory: {:?}", keystore_dir);
```
KolbyML: inline

bin/ream/src/main.rs
```diff
@@ -380,17 +398,58 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
...
+                config.num_active_epochs,
+                config.passphrase.as_deref().unwrap_or(""),
+            );
+
+            // Create keystore file using QsKeystore
+            let keystore = QsKeystore::from_seed_phrase(
+                &seed_phrase,
+                config.lifetime as u32,
+                config.activation_epoch as u32,
+                Some(format!("Ream validator keystore for {:?}", message_type)),
```
KolbyML: inline

bin/ream/src/main.rs
KolbyML: inline

bin/ream/src/main.rs
KolbyML: inline

bin/ream/src/cli/account_manager.rs
```diff
@@ -60,14 +77,32 @@ impl AccountManagerConfig {
             self.lifetime >= MIN_LIFETIME,
             "Lifetime must be at least {MIN_LIFETIME}"
         );
+
+        // Ensure that if import-keystore is provided, seed-phrase is not provided
+        ensure!(
+            !(self.import_keystore.is_some() && self.seed_phrase.is_some()),
+            "Cannot provide both --seed-phrase and --import-keystore. Use one or the other."
+        );
```
KolbyML: we shouldn't be specifying this constraint ourselves, this constraint should be defined in the clap struct definition

KolbyML (changes requested):
Here is some feedback

crates/common/account_manager/src/keystore.rs
ch4r10t33r: Are you happy to call it just `Keystore` then?

bin/ream/src/main.rs
ch4r10t33r: It is just some ascii art for account_manager. I can remove it if you do not like it.

syjn99: Yeah I know it's an ascii art indeed, but I couldn't catch what you're trying to show with it.

crates/common/account_manager/src/lib.rs
ch4r10t33r: From my discussion with the Zeam team, the approach is to assign a unique key for each message type. In this setup, we reuse the seed together with the wallet_index, allowing the same seed to derive multiple keys (one per message type). You can find the proposed EIP here: https://github.com/ReamLabs/EIPs/pull/1

bin/ream/src/cli/account_manager.rs
ch4r10t33r: I am using `Mnemonic::from_entropy` but we will still need `bip39` dependency (for Mnemonic).

bin/ream/Cargo.toml
ch4r10t33r: We do need it as both chrono and bip39 are used in `common/account_manager` as well as `bin/ream` crates

ch4r10t33r:
> @ch4r10t33r Based on the last interop call, I guess `MessageType` is too overkill in this phase. Key management simplification can be discussed more, and it seems adding one of the method preemptively might make more code changes in the future.

It may seem like overkill at the moment, but this is the direction we’re heading. We should keep the code we’ve already built instead of removing it now and having to add it back later.

syjn99:
@ch4r10t33r Not all comments are resolved, please take a look.

crates/common/account_manager/src/keystore.rs
ch4r10t33r: I was experimenting with values to see which is quicker. Reverted them to the values defined in the EIP.

bin/ream/src/main.rs
ch4r10t33r: Noted. Yes, this is just a placeholder.

ch4r10t33r:
> @ch4r10t33r Not all comments are resolved, please take a look.

My apologies, the comments didnt show up  at my end. I have now addressed them all.

bin/ream/Cargo.toml
```diff
@@ -17,18 +17,24 @@ path = "src/main.rs"
 [dependencies]
 alloy-primitives.workspace = true
 anyhow.workspace = true
+bip39.workspace = true
+chrono.workspace = true
```
syjn99: I think we don't need to list `chrono` for here. It is enough to be declared in `account_manager` crate.

bin/ream/Cargo.toml
syjn99: @ch4r10t33r This is not resolved, please take a look.

bin/ream/Cargo.toml
```diff
@@ -17,18 +17,24 @@ path = "src/main.rs"
...
 clap = { workspace = true, features = ["derive", "env"] }
 discv5.workspace = true
 hashbrown.workspace = true
 libp2p-identity.workspace = true
 prometheus_exporter.workspace = true
 rand.workspace = true
 rand_chacha.workspace = true
+rayon.workspace = true
+serde.workspace = true
+serde_json.workspace = true
```
syjn99:
```suggestion
```

We don't need `serde` and `serde_json` here.

Cargo.toml
```diff
@@ -106,6 +109,7 @@ serde = { version = '1.0', features = ['derive', "rc"] }
 serde_json = "1.0.139"
 serde_yaml = "0.9"
 sha2 = "0.10"
+sha3 = "0.10.8"
```
syjn99: Do we need this?

crates/common/account_manager/src/lib.rs
syjn99: This isn't resolved.

crates/common/account_manager/src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+
+/// Validates that a string contains only valid hexadecimal characters
+///
+/// # Arguments
+/// * `hex_str` - The string to validate
+/// * `field_name` - The name of the field being validated (for error messages)
+///
+/// # Returns
+/// * `Result<()>` - Ok if valid, Err with descriptive message if invalid
+pub fn validate_hex_string(hex_str: &str, field_name: &str) -> Result<()> {
```
syjn99:
```suggestion
pub fn validate_hex_string(hex_str: &str, field_name: &str) -> anyhow::Result<()> {
```

For consistency

crates/common/account_manager/src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+    if !hex_str.chars().all(|c| c.is_ascii_hexdigit()) {
+        return Err(anyhow!("{} must be a valid hex string", field_name));
```
syjn99:
```suggestion
        return Err(anyhow!("{field_name} must be a valid hex string"));
```

For consistency

crates/common/account_manager/src/utils.rs
syjn99: Also this function doesn't have to have `field_name` as an argument, just handle it outside of this function.

crates/common/account_manager/src/keystore.rs
syjn99: Not yet resolved for `version`.

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,331 @@
...
+/// Symmetric encryption cipher used for encrypting the private key
+pub const CIPHER_FUNCTION: &str = "aes-256-gcm";
+
+/// Post-quantum signature scheme used for key generation and signing
+pub const KEYTYPE_FUNCTION: &str = "xmss-poisedon2-ots-seed";
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct Keystore {
+    /// Version number, must be 5
+    pub version: u32,
```
syjn99, line 43: We might use `typenum::u5` here if we really want to enforce it.

crates/common/account_manager/src/keystore.rs
syjn99: Not yet resolved @ch4r10t33r

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,331 @@
+use anyhow::{Result, anyhow};
```
syjn99: Please use `anyhow::Result` directly for consistency.

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,331 @@
...
+use hex;
+use rand;
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+use crate::utils::validate_hex_string;
+
+// Cryptographic algorithm constants
+/// Key derivation function used for password-based key derivation
+pub const KDF_FUNCTION: &str = "argon2id";
```
syjn99:
```rust
#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
pub enum KdfFunction {
    #[serde(rename = "argon2id")]
    Argon2id,
}
```

Instead of defining the constant, please use typed enum instead that will reduce most of the code in `validate` (by validating in the compile time)

bin/ream/src/main.rs
```diff
@@ -399,17 +402,58 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
...
-    let (_public_key, _private_key) = ream_account_manager::generate_keys(
-        &seed_phrase,
-        config.activation_epoch,
-        config.num_active_epochs,
-    );
+
+    // Parallelize key generation for each message type
+    let message_types: Vec<MessageType> = MessageType::iter().collect();
+
+    message_types
```
syjn99:
```suggestion
    MessageType::iter().collect()
```

We might inline this

bin/ream/src/cli/account_manager.rs
```diff
@@ -14,23 +17,35 @@ pub struct AccountManagerConfig {
...
     #[arg(long, default_value_t = DEFAULT_ACTIVATION_EPOCH)]
     pub activation_epoch: usize,
 
     /// Number of active epochs
     #[arg(long, default_value_t = DEFAULT_NUM_ACTIVE_EPOCHS)]
     pub num_active_epochs: usize,
+
+    /// Path for keystore file
+    #[arg(long, default_value = DEFAULT_KEYSTORE_PATH)]
+    pub path: Option<String>,
```
syjn99:
```suggestion
    pub path: String,
```

This doesn't have to be `Option<T>` because we have the default value. This will change some lines of code in `main.rs`, but Rust compiler will check this up so I won't jot them down here.

bin/ream/src/cli/account_manager.rs
```diff
@@ -1,10 +1,13 @@
...
 use clap::Parser;
+use tracing::info;
 
-const MIN_CHUNK_SIZE: u64 = 4;
-const MIN_LIFETIME: u64 = 18;
+const MIN_CHUNK_SIZE: u32 = 4;
+const MIN_LIFETIME: u32 = 18;
 const DEFAULT_ACTIVATION_EPOCH: usize = 0;
 const DEFAULT_NUM_ACTIVE_EPOCHS: usize = 1 << 18;
+const DEFAULT_KEYSTORE_PATH: &str = "./keystore/";
```
syjn99:
```suggestion
const DEFAULT_KEYSTORE_PATH: &str = "./.keystore/";
```

You might want this, as `.gitignore` implies.

crates/common/account_manager/src/lib.rs
syjn99: I read all the contents in this link: https://github.com/ReamLabs/EIPs/blob/pq-keystore/EIPS/eip-new.md.

> In this setup, we reuse the seed together with the wallet_index, allowing the same seed to derive multiple keys (one per message type)

I think this description isn't explicitly mentioned in the EIP, right? If then, is it clear for me to understand that this is our team's discussion to do so?

The EIP doesn't mention about the creation of a keystore for every message_type. The decision to use `wallet_index` along with the seed is our implementation. I am happy for us to have a discussion on it. However, this is just an initial implementation.

bin/ream/src/cli/account_manager.rs
```diff
@@ -14,23 +17,35 @@ pub struct AccountManagerConfig {
...
+    #[arg(short, long, conflicts_with = "import_keystore")]
     pub seed_phrase: Option<String>,
 
+    /// Optional BIP39 passphrase used with the seed phrase
+    #[arg(long)]
+    pub passphrase: Option<String>,
+
     /// Activation epoch for the validator
     #[arg(long, default_value_t = DEFAULT_ACTIVATION_EPOCH)]
     pub activation_epoch: usize,
```
syjn99: can it be `u32`? I see some weird pattern for using this like `.try_into().unwrap()`

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,331 @@
...
+        let salt = hex::encode(rand::random::<[u8; 32]>());
+
+        // Generate random nonce for AES-GCM (12 bytes)
+        let nonce = hex::encode(rand::random::<[u8; 12]>());
+
+        // Generate random tag for AES-GCM (16 bytes)
+        let tag = hex::encode(rand::random::<[u8; 16]>());
+
+        // Store the seed phrase as encrypted data (hex encoded)
+        let ciphertext = hex::encode(seed_phrase.as_bytes());
```
syjn99, line 188: This is really weird, so if I just try to decode `ciphertext` with hex (not "decryption" with (a)symmetric key), can I get the original `seed_phrase`?

The comment above looks counter-intuitive, hex encoding isn't an encryption.

syjn99 (changes requested):
Spotted some suspicious things with several issues (dependency, code formatting...)

Also there are some comments that are not resolved but marked as resolved. Please check those again.

crates/common/account_manager/src/keystore.rs
ch4r10t33r: As described earlier the feature of the keystore is still not complete. I have only a stub. I'll be changing this in the subsequent PR.

bin/ream/src/cli/account_manager.rs
ch4r10t33r: I’ve updated it to u32. However, please note that in common/crypto/post_quantum, the hash-sig wrapper still defines these variables as usize.

bin/ream/src/cli/account_manager.rs
ch4r10t33r: done.

crates/common/account_manager/src/keystore.rs
ch4r10t33r: I need `anyhow` function as well, no?

crates/common/account_manager/src/keystore.rs
ch4r10t33r: Fixed now.

crates/common/account_manager/src/utils.rs
ch4r10t33r: Fixed.

crates/common/account_manager/src/keystore.rs
ch4r10t33r: Done. Changed all of these to typed enums.

bin/ream/src/main.rs
ch4r10t33r: Done.

crates/common/account_manager/src/keystore.rs
ch4r10t33r: When I run `make pr` command, all these imports automatically get grouped. So I guess we have to live with this.

ch4r10t33r: Fixed.

crates/common/account_manager/src/keystore.rs
ch4r10t33r: I'll do this in the next PR if you don't mind.

bin/ream/Cargo.toml
```diff
@@ -17,13 +17,16 @@ path = "src/main.rs"
 [dependencies]
 alloy-primitives.workspace = true
 anyhow.workspace = true
+bip39.workspace = true
+chrono = { version = "0.4", features = ["serde"] }
```
syjn99: We don't need this at `bin/ream/Cargo.toml`

bin/ream/src/cli/account_manager.rs
```diff
@@ -14,23 +17,35 @@ pub struct AccountManagerConfig {
...
     #[arg(long, default_value_t = DEFAULT_ACTIVATION_EPOCH)]
-    pub activation_epoch: usize,
+    pub activation_epoch: u32,
 
     /// Number of active epochs
     #[arg(long, default_value_t = DEFAULT_NUM_ACTIVE_EPOCHS)]
-    pub num_active_epochs: usize,
+    pub num_active_epochs: u32,
+
+    /// Path for keystore file
```
syjn99:
```suggestion
    /// Path for keystore directory
```

Just a nit to clarify that `path` implies the directory.

bin/ream/src/main.rs
```diff
@@ -399,17 +402,56 @@ pub async fn run_account_manager(mut config: AccountManagerConfig) {
 
     let seed_phrase = config.get_seed_phrase();
 
+    // Create keystore directory if it doesn't exist
+    let keystore_path = &config.path;
+    let keystore_dir = Path::new(keystore_path);
+    if !keystore_dir.exists() {
+        fs::create_dir_all(keystore_dir).expect("Failed to create keystore directory");
+        info!("Created keystore directory: {keystore_path}");
+    }
```
syjn99:
```suggestion
    // Create keystore directory if it doesn't exist
    let keystore_dir = Path::new(&config.path);
    if !keystore_dir.exists() {
        fs::create_dir_all(keystore_dir).expect("Failed to create keystore directory");
        info!("Created keystore directory: {}", keystore_path.display());
    }
```

Removing unnecessary variable

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,420 @@
...
+
+impl KdfFunction {
+    pub const fn default() -> Self {
+        KdfFunction::Argon2Id
+    }
+
+    pub fn all() -> &'static [KdfFunction] {
+        &[KdfFunction::Argon2Id]
+    }
+}
```
syjn99: Hmm I guess all those implementations aren't necessary in this phase. If you want some arm in an enum to be default, make it like:

```rust
/// Key derivation function used for password-based key derivation
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, Default)]
pub enum KdfFunction {
    #[default]
    #[serde(rename = "argon2id")]
    Argon2Id,
}
```

Same for all other enums.

crates/common/account_manager/src/keystore.rs
syjn99: Could you add Github issues then? Meta tracking issues must be helpful in this context.

By looking into the current code, it seems wrong at a glance.

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,420 @@
...
+
+        // Store the seed phrase as encrypted data (hex encoded)
+        let ciphertext = hex::encode(seed_phrase.as_bytes());
+
+        let kdf = KdfParams::new_full(65536, 4, 2, salt);
+        let cipher = CipherParams::new(nonce, tag, ciphertext);
+        let crypto = CryptoParams { kdf, cipher };
+        let keytype = KeyType::new(lifetime, activation_epoch);
+
+        let mut keystore = Self::new(crypto, keytype, uuid);
```
syjn99:
```suggestion
        let mut keystore = Self::new(
            CryptoParams {
                kdf: KdfParams::new_full(65536, 4, 2, salt),
                cipher: CipherParams::new(nonce, tag, ciphertext),
            },
            KeyType::new(lifetime, activation_epoch),
            Uuid::new_v4(),
        );
```

I think this is much cleaner

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,420 @@
...
+        keystore.path = path;
+
+        keystore
+    }
+
+    /// Validate the keystore structure
+    pub fn validate(&self) -> anyhow::Result<()> {
+        // Validate required constants for external data
+        if self.version != KEYSTORE_VERSION {
+            return Err(anyhow!("Version must be {}", KEYSTORE_VERSION));
```
syjn99:
```suggestion
            return Err(anyhow!("Version must be {KEYSTORE_VERSION}"));
```

crates/common/account_manager/src/lib.rs
```diff
@@ -1,27 +1,51 @@
...
+    let (public_key, private_key) = PrivateKey::generate(
+        &mut <ChaCha20Rng as SeedableRng>::from_seed(derived_seed),
+        activation_epoch as usize,
+        num_active_epochs as usize,
+    );
+
+    // Display public key contents
+    match serde_json::to_string_pretty(&public_key.inner) {
+        Ok(json) => info!("Public key contents: {json}"),
+        Err(e) => info!("Public key generated successfully (could not serialize) due to {e}"),
```
syjn99: This doesn't resolved yet...

crates/common/account_manager/src/utils.rs
```diff
@@ -0,0 +1,34 @@
+use anyhow::anyhow;
+
+/// Validates that a string contains only valid hexadecimal characters
+///
+/// # Arguments
+/// * `hex_str` - The string to validate
+///
+/// # Returns
+/// * `Result<()>` - Ok if valid, Err with descriptive message if invalid
+pub fn validate_hex_string(hex_str: &str) -> anyhow::Result<()> {
```
syjn99: This shouldn't be returning error, just return boolean instead.

syjn99 (commented):
I think this is getting much closer!

crates/common/account_manager/src/keystore.rs
ch4r10t33r: Do you mind if we retain it as it is? As there is no logical flaw in the code.

crates/common/account_manager/src/lib.rs
ch4r10t33r: As discussed, I have applied the suggestion from @KolbyML

crates/common/account_manager/src/keystore.rs
ch4r10t33r: Removed the implementations.

bin/ream/src/cli/account_manager.rs
```diff
@@ -14,23 +17,35 @@ pub struct AccountManagerConfig {
...
+    pub activation_epoch: u32,
 
     /// Number of active epochs
     #[arg(long, default_value_t = DEFAULT_NUM_ACTIVE_EPOCHS)]
-    pub num_active_epochs: usize,
+    pub num_active_epochs: u32,
+
+    /// Path for keystore directory
+    #[arg(long, default_value = DEFAULT_KEYSTORE_PATH)]
+    pub path: String,
```
KolbyML: shouldn't we be using Ream's data-dir directory not a hidden local folder to the binary

bin/ream/src/cli/account_manager.rs
```diff
@@ -14,23 +17,35 @@ pub struct AccountManagerConfig {
...
+
+    /// Import existing keystore
+    #[arg(long, conflicts_with = "seed_phrase")]
+    pub import_keystore: Option<String>,
```
KolbyML:
```suggestion
    pub import_keystore: Option<PathBuf>,
```

if this is a supposed to respond to a Path, it shouldn't be a string

KolbyML:
```suggestion
```

It looks like this flag isn't used so it should be removed, it can be readded when functionality that uses it is added

bin/ream/src/cli/account_manager.rs
```diff
@@ -60,14 +78,26 @@ impl AccountManagerConfig {
...
             phrase.clone()
         } else {
-            "default_seed_phrase".to_string()
+            // Generate a new BIP39 mnemonic with 24 words (256 bits of entropy)
+            let entropy: [u8; 32] = rand::random();
+            let mnemonic = Mnemonic::from_entropy(&entropy).expect("Failed to generate mnemonic");
+            let phrase = mnemonic.words().collect::<Vec<_>>().join(" ");
+            info!(
+                "========================================================================================="
+            );
```
KolbyML:
```suggestion
info!("{}", "=".repeat(89));
```

bin/ream/src/cli/account_manager.rs
```diff
@@ -60,14 +78,26 @@ impl AccountManagerConfig {
...
+            info!("Generated new seed phrase (KEEP SAFE): {phrase}");
+            info!(
+                "========================================================================================="
+            );
```
KolbyML:
```suggestion
info!("{}", "=".repeat(89));
```

bin/ream/src/main.rs
```diff
@@ -9,6 +10,7 @@ use std::{
 use alloy_primitives::hex;
 use clap::Parser;
 use libp2p_identity::secp256k1;
+use rayon::prelude::*;
```
KolbyML: I don't think we should be using wildcard imports in production code

crates/common/account_manager/src/keystore.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    /// Convert to JSON string
+    pub fn to_json(&self) -> Result<String, serde_json::Error> {
+        serde_json::to_string_pretty(self)
+    }
+
+    /// Create from JSON string
+    pub fn from_json(json: &str) -> Result<Self, serde_json::Error> {
+        serde_json::from_str(json)
+    }
```
KolbyML:
```suggestion
```

This code is redundent we can use serde_json() directly

crates/common/account_manager/src/lib.rs
KolbyML:
```suggestion
        Err(err) => info!("Public key generated successfully (could not serialize) due to {err}"),
```

for consistency

crates/common/account_manager/src/message_types.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    // Iterator method to loop through all variants
+    pub fn iter() -> impl Iterator<Item = MessageType> + 'static {
+        Self::all().into_iter()
+    }
+
+    // Method to return the number of enum variants
+    pub const fn count() -> usize {
+        Self::all().len()
+    }
```
KolbyML:
```suggestion
    /// Method to get all enum variants as an array
    pub const fn all() -> [MessageType; 2] {
        [MessageType::Attestation, MessageType::Block]
    }

    /// Iterator method to loop through all variants
    pub fn iter() -> impl Iterator<Item = MessageType> + 'static {
        Self::all().into_iter()
    }

    /// Method to return the number of enum variants
    pub const fn count() -> usize {
        Self::all().len()
    }
```

these should be doc comments

bin/ream/src/cli/account_manager.rs
ch4r10t33r: The keystore should be stored in its own directory for separation of concerns, rather than being bundled into Ream’s datadir.

bin/ream/src/cli/account_manager.rs
ch4r10t33r: A cosmetic comment, which I’ve addressed, but I don’t see how it impacts the code. Are we establishing this as a standard? If so, can we document it somewhere to avoid similar comments in the future? This is exactly where I think AI could be useful.

crates/common/account_manager/src/keystore.rs
ch4r10t33r: Isn't `keystore.to_json()` is cleaner than `serde_json::to_string_pretty(&keystore)`?  Also they provide a clean, type-specific API for JSON serialisation of the `Keystore` type. 

If we are defining a standard of using serde_json directly going forward then we ought to document this somewhere. Wdyt?

crates/common/account_manager/src/lib.rs
ch4r10t33r: Another standard that should added/recommend to be followed within the codebase. `No single letter variables`. @KolbyML

crates/common/account_manager/src/keystore.rs
ch4r10t33r: Done. https://github.com/ReamLabs/ream/issues/792

Cargo.toml
```diff
@@ -98,6 +99,7 @@ parking_lot = "0.12.3"
 prometheus_exporter = "0.8.5"
 rand = "0.9"
 rand_chacha = "0.9"
+rayon = "1.11.0"
```
syjn99: Please remove this as well.

syjn99 (commented):
Also could you please change the PR title? `account_manager changes` are quite vague

Cargo.toml
ch4r10t33r: done.

bin/ream/src/cli/account_manager.rs
KolbyML: I still don't see why the default directory should be a hidden path in the current active directory

crates/common/account_manager/src/lib.rs
```diff
@@ -1,27 +1,51 @@
...
+pub mod utils;
+
 pub fn generate_keys(
     seed_phrase: &str,
-    activation_epoch: usize,
-    num_active_epochs: usize,
+    wallet_index: u32,
+    activation_epoch: u32,
+    num_active_epochs: u32,
+    passphrase: &str,
```
KolbyML, line 19: I think this function should be implemented on the PrivateKey struct, the PublicKey is a derivative of the PrivateKey

crates/common/account_manager/src/lib.rs
```diff
@@ -1,27 +1,51 @@
...
+use bip39::Mnemonic;
 use rand::SeedableRng;
 use rand_chacha::ChaCha20Rng;
 use ream_post_quantum_crypto::hashsig::{private_key::PrivateKey, public_key::PublicKey};
 use sha2::{Digest, Sha256};
 use tracing::info;
 
+pub mod keystore;
+pub mod message_types;
+pub mod utils;
```
KolbyML: mods should be at the top of the file

crates/common/account_manager/src/keystore.rs
KolbyML: I don't think we need to document this anywhere, it is pretty standard to use serde_X when you need it

KolbyML (changes requested):
Here is some feedback

crates/common/account_manager/src/lib.rs
ch4r10t33r: I am not sure what you mean. Are you referring to the `passphrase`? That is used as a seed to generate the keypair.

KolbyML: pub fn generate_keys(

crates/common/account_manager/src/keystore.rs
ch4r10t33r: I’ve amended this, but the approach I used has no logical flaw. It’s simply a matter of preference. Other developers might make similar choices, so to avoid such comments in the future, I strongly believe we should document this as a standard.

crates/common/account_manager/src/lib.rs
ch4r10t33r: There is already a generate function in the private key module: https://github.com/ReamLabs/ream/blob/c914186ea5df33cfeab6cd4d5798de88d47c905f/crates/crypto/post_quantum/src/hashsig/private_key.rs#L19

The generate_keys function in the account manager is simply a wrapper around it, adding functionality such as determining the salt.

bin/ream/src/cli/account_manager.rs
ch4r10t33r: The standard practice is to store data, secrets, and similar items in separate directories. Do you object only to the path being a hidden directory, or do you prefer storing these in a directory outside the current active directory?

crates/common/account_manager/src/keystore.rs
KolbyML: The code you added is redundant, this is a Rust codebase not a Typescript/Python/etc. A user might want to print it in json or yaml, or any format under the sun, that is possible with serde. If they want json, they will use json serde etc.

Code having no logical flaws, doesn't mean the code is good

crates/common/account_manager/src/lib.rs
KolbyML: it is fine to add another, `generate_keys` is just floating around right now, well it is a constructor for an object

KolbyML: why is it called generate_key's as well, when we are only generating 1 key? A public key and a private key are a pair, a public key isn't a seperate key, it is a derivation. So I don't think the plural makes sense

KolbyML: maybe use generate_key_pair_with_salt?

KolbyML: and rename generate to generate_key_pair

ch4r10t33r: > why is it called generate_key's as well, when we are only generating 1 key? A public key and a private key are a pair, a public key isn't a seperate key, it is a derivation. So I don't think the plural makes sense

It is called `generate_keys` because we are generating multiple key pairs (one for each `message_type`).

ch4r10t33r: Honestly, I am happy to rename the function names to something you like. I'll push the changes once done. 

Btw, pls note that the `generate` function of private key module was already part of the codebase (It was developed by @shariqnaiyer ).

KolbyML: > > why is it called generate_key's as well, when we are only generating 1 key? A public key and a private key are a pair, a public key isn't a seperate key, it is a derivation. So I don't think the plural makes sense
> 
> It is called `generate_keys` because we are generating multiple key pairs (one for each `message_type`).

```) -> (PublicKey, PrivateKey) {```

But the function here is returning one pair?

Comment (ch4r10t33r) [reply crates/common/account_manager/src/lib.rs:19]: Renamed the functions as per your suggestion.

Comment (KolbyML) [reply crates/common/account_manager/src/lib.rs:19]: I am not sure why this was marked as resolved, `pub fn generate_key_pair_with_salt(` should be a member of the PrivateKey struct
`

Comment (ch4r10t33r) [reply crates/common/account_manager/src/lib.rs:19]: I'm not clear on your point. I thought you wanted me to rename both functions. 

The `generate_key_pair_with_salt` function belongs to the `account_manager` crate, as it’s a core feature of `account_manager`, not the private key. We’ll expand this functionality with alternative approaches. The private key remains unaffected by these changes and shouldn’t need updates whenever we adjust the strategy in `account_manager`.

File: crates/common/account_manager/src/keystore.rs (hunk 57 above)
Comment (ch4r10t33r) [reply crates/common/account_manager/src/keystore.rs]: Again, this is subjective. But I have changed the code as per your suggestion so marking this as resolved.

File: crates/common/account_manager/src/lib.rs (hunk 60 above)
Comment (syjn99) [reply crates/common/account_manager/src/lib.rs:19]: Dropping my thoughts here:

How about adding a function

```rust
  pub fn derive_wallet_rng(
      seed_phrase: &str,
      wallet_index: u32,
      passphrase: &str,
  ) -> ChaCha20Rng
```

And use it like:

```rust
PrivateKey::generate(&mut derive_wallet_rng(...), ...)
```
?

So that we can remove `generate_key*` function name which derives some convo here.

ch4r10t33r: Sure, I can rename the function. But my main point of contention is that the functionality should remain in account_manager rather than PrivateKey.

ch4r10t33r: @KolbyML @syjn99 I will be raising a new PR in the next 2 days which requires additional changes and inclusion of new features related to keystore encryption etc. I can always handle these suggestions as part of that PR. Please can you approve this PR now?

bin/ream/src/cli/account_manager.rs
ch4r10t33r: Changed the keystore to be a sub directory of ream's data directory.

KolbyML (approved):
:shipit: PR looks great, good work 🥳

Issue #754, opened 2025-08-29
Support CLI flag for reusing the network identity


syjn99:
### Describe the feature

For PQ Devnets, we won't support discovery mechanism which means each node is only aware of their peers through some config files. The config file would be consisted of an array of ENR addresses.

ENR address can be precalculated if we can use pre-generated secp256k1 keypair. 

https://github.com/ReamLabs/ream/blob/aeb1d9463271b09e91f091f86fbb5e02fcb2c108/crates/networking/p2p/src/network/lean/mod.rs#L104

So far, a lean node generates a new keypair everytime it starts. If we can provide any `secret-key-path` encoded with specific scheme (e.g., protobuf), and decode it to derive a keypair, we can fix each node's identity every time.

PR #758, opened 2025-08-31
feat: add support CLI flag for reusing the network identity

Kayden-ML:
### What was wrong?

Fixes: https://github.com/ReamLabs/ream/issues/754

### How was it fixed?

adds a cli flag for reusing the network identity


KolbyML (approved):
:shipit: looks good

PR #760, opened 2025-09-01
chore: use `std::fmt::Display` for consistency

syjn99:
### What was wrong?

 

The node would not run in `no_std` environment, so use `std::fmt::Display` instead of `core::fmt::Display.`
Also did some cosmetics for all over the codebase.

### How was it fixed?

 

Use `std::fmt::Display`

### To-Do

 
### DIFF
--- a/crates/common/api_types/beacon/src/id.rs
+++ b/crates/common/api_types/beacon/src/id.rs
@@ -1,5 +1,4 @@
-use core::fmt;
-use std::{fmt::Display, str::FromStr};
+use std::str::FromStr;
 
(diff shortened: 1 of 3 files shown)


varun-doshi (commented):
> The node would not run in no_std environment, so use std::fmt::Display instead of core::fmt::Display

Would like to clarify my understanding; 
`core::Display` and `std::fmt::Display` both dont work in no-std environment. Afaik `std::Display` is just a re-export of `core::Display`.
Correct me if im mistaken

Update: Nevermind, I misread the description.

PR #768, opened 2025-09-03
chore: remove `actix-web` dependency in `api_types/beacon`

syjn99:
### What was wrong?

 

`cargo udeps` fails: https://github.com/ReamLabs/ream/actions/runs/17431884721/job/49491965211

### How was it fixed?

 

Delete `actix-web` dependency in `api_types/beacon` crate.

### To-Do

 
### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5233,7 +5233,6 @@ dependencies = [
 name = "ream-api-types-beacon"
 version = "0.1.0"
 dependencies = [
- "actix-web",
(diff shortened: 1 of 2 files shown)


jveer634:
@syjn99  Can we have this udeps check in the make pr command as well

shariqnaiyer:
> @syjn99 Can we have this udeps check in the make pr command as well

The udeps check takes about 7-8 mins so is likely not best to have it in make pr.

PR #772, opened 2025-09-03
feat: setup lean DB

### LINKED ISSUE #764: Set up DB for Lean Node
### Describe the feature

https://github.com/ReamLabs/ream/blob/406f50826e23aac334cffa1e0e307210d54a676b/crates/common/chain/lean/src/lean_chain.rs#L22-L36

Currently the lean node holds all items (`chain` and `post_states` in `LeanChain` struct) in the memory which won't be sustainable. Some outdated items can be pruned, but I guess at least blocks and states can be saved in the DB.

Before deciding which items to be saved, setting up DB for lean node should be needed.

### Additional context

_No response_

varun-doshi:
### What was wrong?

 
Fixes #764 

### How was it fixed?

 

Sets up the LeanDB in the `LeanChain` struct.


Kayden-ML (changes requested):
https://github.com/ReamLabs/ream/issues/773
Shouldn't this be done in this PR since its pretty much just copy and paste from the beacon code?

The work shouldn't take that much time I don't see why it couldn't be copy and pasted here?

crates/storage/src/lean/db.rs
```diff
@@ -0,0 +1,61 @@
...
+            } else {
+                fs::remove_file(&path)?;
+            }
+        }
+        info!("Database contents cleared successfully.");
+    } else {
+        info!("Operation canceled by user.");
+    }
+    Ok(())
+}
```
KolbyML: why are we duplicating perfectly reusable code

crates/storage/src/lean/db.rs
```diff
@@ -0,0 +1,61 @@
...
+use tracing::info;
+
+use crate::errors::StoreError;
+
+pub const REDB_FILE: &str = "lean.redb";
+
+/// The size of the cache for the database
+///
+/// 1 GiB
+pub const REDB_CACHE_SIZE: usize = 1_024 * 1_024 * 1_024;
```
KolbyML: we only need one database, also duplicate const

crates/storage/src/lean/db.rs
```diff
@@ -0,0 +1,61 @@
...
+pub struct ReamLeanDB {
+    pub db: Arc<Database>,
+    pub data_dir: PathBuf,
+}
+
+impl ReamLeanDB {
+    pub fn new(data_dir: PathBuf) -> Result<Self, StoreError> {
+        let db = Builder::new()
+            .set_cache_size(REDB_CACHE_SIZE)
+            .create(data_dir.join(REDB_FILE))?;
```
KolbyML: We should only have this once

We should have 

```ReamDB``` which as a constructor to build ReamDB, then it will have 2 functions

BeaconDB, LeanDB these will hold network specific functions. These will only be constructable from `ReamDB`

File: crates/storage/src/db/lean.rs [hunk 4]
```diff
@@ -0,0 +1,39 @@
...
+
+use crate::tables::lean::{
+    lean_block::LeanBlockTable, lean_state::LeanStateTable, slot_index::SlotIndexTable,
+    state_root_index::StateRootIndexTable,
+};
+
+#[derive(Clone, Debug)]
+pub struct LeanDB {
+    pub db: Arc<Database>,
+    pub data_dir: PathBuf,
```
KolbyML:
```suggestion
#[derive(Clone, Debug)]
pub struct LeanDB {
    pub db: Arc<Database>,
```

path isn't used in this db

crates/storage/src/db/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        write_txn.open_table(UNREALIZED_JUSTIFED_CHECKPOINT_FIELD)?;
+        write_txn.commit()?;
+
+        fs::create_dir_all(data_dir.join(BLOB_FOLDER_NAME))?;
+
+        Ok(BeaconDB {
+            db: ream_db.db.clone(),
+            data_dir,
+        })
+    }
```
KolbyML:
```suggestion
    pub fn init_beacon_db(&self, data_dir: PathBuf) -> Result<BeaconDB, StoreError> {
        let write_txn = self.db.begin_write()?;
        write_txn.open_table(BEACON_BLOCK_TABLE)?;
        write_txn.open_table(BEACON_STATE_TABLE)?;
        write_txn.open_table(BLOCK_TIMELINESS_TABLE)?;
        write_txn.open_table(CHECKPOINT_STATES_TABLE)?;
        write_txn.open_table(EQUIVOCATING_INDICES_FIELD)?;
        write_txn.open_table(FINALIZED_CHECKPOINT_FIELD)?;
        write_txn.open_table(GENESIS_TIME_FIELD)?;
        write_txn.open_table(JUSTIFIED_CHECKPOINT_FIELD)?;
        write_txn.open_table(LATEST_MESSAGES_TABLE)?;
        write_txn.open_multimap_table(PARENT_ROOT_INDEX_MULTIMAP_TABLE)?;
        write_txn.open_table(PROPOSER_BOOST_ROOT_FIELD)?;
        write_txn.open_table(SLOT_INDEX_TABLE)?;
        write_txn.open_table(STATE_ROOT_INDEX_TABLE)?;
        write_txn.open_table(TIME_FIELD)?;
        write_txn.open_table(UNREALIZED_FINALIZED_CHECKPOINT_FIELD)?;
        write_txn.open_table(UNREALIZED_JUSTIFICATIONS_TABLE)?;
        write_txn.open_table(UNREALIZED_JUSTIFED_CHECKPOINT_FIELD)?;
        write_txn.commit()?;

        fs::create_dir_all(data_dir.join(BLOB_FOLDER_NAME))?;

        Ok(BeaconDB {
            db: self.db.clone(),
            data_dir,
        })
    }
```

crates/storage/src/db/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        write_txn.open_table(LEAN_STATE_TABLE)?;
+        write_txn.open_table(LEAN_SLOT_INDEX_TABLE)?;
+        write_txn.open_table(LEAN_STATE_ROOT_INDEX_TABLE)?;
+        write_txn.commit()?;
+
+        Ok(LeanDB {
+            db: ream_db.db.clone(),
+            data_dir,
+        })
+    }
```
KolbyML:
```suggestion
    pub fn init_lean_db(&self) -> Result<LeanDB, StoreError> {
        let write_txn = self.db.begin_write()?;
        write_txn.open_table(LEAN_BLOCK_TABLE)?;
        write_txn.open_table(LEAN_STATE_TABLE)?;
        write_txn.open_table(LEAN_SLOT_INDEX_TABLE)?;
        write_txn.open_table(LEAN_STATE_ROOT_INDEX_TABLE)?;
        write_txn.commit()?;

        Ok(LeanDB {
            db: self.db.clone(),
        })
    }
```

crates/storage/src/db/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        },
+        lean::{
+            lean_block::LEAN_BLOCK_TABLE, lean_state::LEAN_STATE_TABLE,
+            slot_index::LEAN_SLOT_INDEX_TABLE, state_root_index::LEAN_STATE_ROOT_INDEX_TABLE,
+        },
+    },
+};
+
+pub const REDB_BEACON_FILE: &str = "ream.redb";
+pub const REDB_LEAN_FILE: &str = "lean.redb";
```
KolbyML:
```suggestion
pub const REDB_FILE: &str = "ream.redb";
```

crates/storage/src/db/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+#[derive(Clone, Debug)]
+pub struct ReamDB {
+    db: Arc<Database>,
+}
+
+impl ReamDB {
+    pub fn new(data_dir: PathBuf) -> Result<Self, StoreError> {
+        let db = Builder::new()
+            .set_cache_size(REDB_CACHE_SIZE)
+            .create(data_dir.join(REDB_BEACON_FILE))?;
```
KolbyML:
```suggestion
impl ReamDB {
    pub fn new(data_dir: PathBuf) -> Result<Self, StoreError> {
        let db = Builder::new()
            .set_cache_size(REDB_CACHE_SIZE)
            .create(data_dir.join(REDB_FILE))?;
```

We should only use one database file, we could be running lean,beacon, or lean+beacon for example during the transition

varun-doshi:
Not ready for review yet

bin/ream/src/main.rs
```diff
@@ -133,10 +133,22 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
 
     set_lean_network_spec(config.network);
 
+    let ream_dir = setup_data_dir(APP_NAME, config.data_dir.clone(), config.ephemeral)
+        .expect("Unable to initialize database directory");
+
+    if config.purge_db {
+        reset_db(ream_dir.clone()).expect("Unable to delete database");
+    }
```
KolbyML: I think the ReamDB should be initialized before we call `run_lean_node` or `run_beacon_node`, then we will pass ReamDB into these functions and call `init_x_db`. As we will possibly running

- beacon
- lean
- or beacon + lean during the transition

varun-doshi: the issue with creating `ReamDB` in `main` is we dont have access to `data_dir` there. `data_dir` is only in scope inside the `run_x_node` functions

varun-doshi: and that is also why i made the `init_x_db` the entrypoint, not `new`. just to clarify:
```
let db = Builder::new()
            .set_cache_size(REDB_CACHE_SIZE)
            .create(data_dir.join(REDB_FILE))?;
```

this will only create the db once even if called more than once so that should cause any issues imo

varun-doshi: I've pushed your other suggestions, let me know what you think about this issue

crates/storage/src/db/mod.rs
varun-doshi: explanation: https://github.com/ReamLabs/ream/pull/772#discussion_r2321379796

crates/storage/src/db/mod.rs
varun-doshi: explanation: https://github.com/ReamLabs/ream/pull/772#discussion_r2321379796

bin/ream/src/cli/lean_node.rs
```diff
@@ -22,6 +22,22 @@ pub struct LeanNodeConfig {
...
+
+    #[arg(
+        long,
+        short,
+        help = "Use new data directory, located in OS temporary directory. If used together with --data-dir, new directory will be created there instead."
+    )]
+    pub ephemeral: bool,
+
+    #[arg(long, help = "Purges the database.")]
+    pub purge_db: bool,
```
KolbyML: move these flags to the main ReamConfig, as the datadir is for Ream, not a specific command. remove them from BeaconNodeConfig as well

bin/ream/src/main.rs
KolbyML: With this suggestion ``move these flags to the main ReamConfig, as the datadir is for Ream remove them from BeaconNodeConfig as well``, my suggestion from before should be more straightforward now

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -231,3 +239,23 @@ impl LeanChain {
...
+            post_states: HashMap::new(),
+            known_votes: Vec::new(),
+            head: B256::default(),
+            safe_target: B256::default(),
+            new_votes: Vec::new(),
+            genesis_hash: B256::default(),
+            num_validators: 0,
+        }
+    }
+}
```
KolbyML: hmm I think we should just remove Default for LeanChain now, as it doesn't make much sense to be initializing the database like this, it is kind of a trap

crates/storage/src/db/mod.rs
KolbyML: With the comment I left, my suggestion should be straight forward now

bin/ream/src/cli/lean_node.rs
KolbyML: This is also how geth operates for example

KolbyML (commented):
Left some feedback

crates/common/chain/lean/src/lean_chain.rs
varun-doshi: Yeah this was a bit hacky; mostly becasue `LeanDB` does not implement `Default`. 
Hoenstly this is used in only 1 place for a test. So I'd be okay with enabling this only for tests. Wdyt?

KolbyML: What do you mean by enabling this? We should remove default and just create LeanChain manually in the test

varun-doshi: I meant enabling Default only for tests

crates/storage/src/db/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        write_txn.open_table(LEAN_STATE_ROOT_INDEX_TABLE)?;
+        write_txn.commit()?;
+
+        Ok(LeanDB {
+            db: self.db.clone(),
+        })
+    }
+}
+
+pub fn reset_db(db_path: PathBuf) -> anyhow::Result<()> {
```
KolbyML:
```suggestion
pub fn reset_db(db_path: &PathBuf) -> anyhow::Result<()> {
```
can't this take a reference?

bin/ream/src/main.rs
```diff
@@ -77,13 +77,21 @@ fn main() {
...
     let executor = ReamExecutor::new().expect("unable to create executor");
     let executor_clone = executor.clone();
+    let ream_dir = setup_data_dir(APP_NAME, cli.data_dir.clone(), cli.ephemeral)
+        .expect("Unable to initialize database directory");
+
+    if cli.purge_db {
+        reset_db(ream_dir.clone()).expect("Unable to delete database");
+    }
+
+    let ream_db = ReamDB::new(ream_dir.clone()).expect("unable to init Ream Database");
```
```suggestion
    let ream_dir = setup_data_dir(APP_NAME, cli.data_dir.clone(), cli.ephemeral)
        .expect("Unable to initialize database directory");

    if cli.purge_db {
        reset_db(&ream_dir).expect("Unable to delete database");
    }

    let ream_db = ReamDB::new(ream_dir).expect("unable to init Ream Database");
```

Then we should be able to remove the clones

bin/ream/src/main.rs
```diff
@@ -133,10 +141,17 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
 
     set_lean_network_spec(config.network);
 
+    // Initialize the lean database
+    let lean_db = ream_db
+        .init_lean_db()
+        .expect("unable to init Ream Lean Database");
+
+    info!("ream lean database initialized ");
```
```suggestion
    info!("Ream's lean database has been initialized");
```

or you can do `is` idk, I mostly made this suggestion due to the extra space at the end

bin/ream/src/main.rs
```diff
@@ -239,24 +254,20 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor) {
...
-    }
-
-    let ream_db = ReamDB::new(ream_dir.clone()).expect("unable to init Ream Database");
+    // Initialize the beacon database
+    let beacon_db = ream_db
+        .init_beacon_db()
+        .expect("unable to init Ream Beacon Database");
 
-    info!("ream database initialized ");
+    info!("ream beacon database initialized ");
```
```suggestion
    info!("ream beacon database initialized");
```

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -36,10 +39,12 @@ pub struct LeanChain {
 }
 
 impl LeanChain {
-    pub fn new(genesis_block: Block, genesis_state: LeanState) -> LeanChain {
+    pub fn new(genesis_block: Block, genesis_state: LeanState, db: LeanDB) -> LeanChain {
         let genesis_hash = genesis_block.tree_hash_root();
 
         LeanChain {
+            // Database
+            store: Arc::new(Mutex::new(db)),
```
KolbyML: can you move this comment and the comment below to the struct and make them doc comments

crates/storage/src/tables/lean/mod.rs
```diff
@@ -0,0 +1,4 @@
+pub mod lean_block;
+pub mod lean_state;
+pub mod slot_index;
+pub mod state_root_index;
```
KolbyML: image

Should we have table's/field's for the rest of these ^ as well?

varun-doshi: Made tables/field for `known_votes`, `safe_target`
I dont think we need to store `genesis_hash` and `num_validators`. These are constants I believe so should be fine to hold in this struct itself.

I'd like to wait and decide the best way to store `new_votes` and `head` while integrating the DB in the next PR

Issue #594, opened 2025-06-18
gossipsub validate and forward: sync_committee_contribution_and_proof


unnawut:
### Describe the feature

Validate each received `sync_committee_contribution_and_proof ` message and forward it upon successful validation.

#### Acceptance criteria:
- Each received `sync_committee_contribution_and_proof ` gossipsub message is validated according to the following specs:
  - https://ethereum.github.io/consensus-specs/specs/altair/p2p-interface/#sync_committee_contribution_and_proof
  - https://ethereum.github.io/consensus-specs/specs/deneb/p2p-interface/#beacon_aggregate_and_proof
  - https://ethereum.github.io/consensus-specs/specs/electra/p2p-interface/#beacon_aggregate_and_proof
- The message is dropped if it matches at least one `[IGNORE]` or `[REJECT]` rule
- The message is forwarded if it does not match any `[IGNORE]` or `[REJECT]` rule

jveer634:
Assign me this one please

PR #775, opened 2025-09-05
feat: implemenation of gossipsub validate and forward: sync_committee_contribution_and_proof

jveer634:
### What was wrong?

fixes: #594

 

### To-Do

Once verify the validation logic followed for the `check_sync_committee` mainly

 
### DIFF
--- a/crates/networking/manager/src/gossipsub/validate/sync_committee_contribution_and_proof.rs
+++ b/crates/networking/manager/src/gossipsub/validate/sync_committee_contribution_and_proof.rs
@@ -1,5 +1,6 @@
+use alloy_primitives::FixedBytes;
 use anyhow::anyhow;
-use ream_bls::{PublicKey, traits::Verifiable};
+use ream_bls::{BLSSignature, PublicKey, traits::Verifiable};
 use ream_chain_beacon::beacon_chain::BeaconChain;
 use ream_consensus_beacon::electra::beacon_state::BeaconState;
 use ream_consensus_misc::{
@@ -9,7 +10,7 @@ use ream_consensus_misc::{
     },
 };
 use ream_storage::{
-    cache::{CachedDB, SyncCommitteeContribution, SyncCommitteeKey},
+    cache::{CacheSyncCommitteeContribution, CachedDB, SyncCommitteeKey},
     tables::table::Table,
 };
 use ream_validator_beacon::{
@@ -17,7 +18,9 @@ use ream_validator_beacon::{
         DOMAIN_CONTRIBUTION_AND_PROOF, DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF,
         SYNC_COMMITTEE_SUBNET_COUNT,
     },
-    contribution_and_proof::SignedContributionAndProof,
+    contribution_and_proof::{
+        ContributionAndProof, SignedContributionAndProof, SyncCommitteeContribution,
+    },
     sync_committee::{SyncAggregatorSelectionData, is_sync_committee_aggregator},
 };
 
@@ -107,7 +110,7 @@ pub async fn validate_sync_committee_contribution_and_proof(
 
     // [IGNORE] if a valid sync committee contribution with equal slot, beacon_block_root and
     // subcommittee_index has already been seen.
-    let sync_contribution = SyncCommitteeContribution {
+    let sync_contribution = CacheSyncCommitteeContribution {
         slot: contribution.slot,
         beacon_block_root: contribution.beacon_block_root,
         subcommittee_index: contribution.subcommittee_index,
@@ -133,7 +136,6 @@ pub async fn validate_sync_committee_contribution_and_proof(
     // [IGNORE] if a valid sync committee contribution has already been seen from the
     // aggregator with index contribution_and_proof.aggregator_index for the slot contribution.slot
     // and subcommittee index contribution.subcommittee_index.
-
     let sync_committee_key = SyncCommitteeKey {
         subnet_id: contribution.subcommittee_index,
         slot: contribution.slot,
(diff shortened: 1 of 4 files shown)


crates/networking/manager/src/gossipsub/validate/sync_committee_contribution_and_proof.rs
```diff
@@ -0,0 +1,267 @@
...
+    let domain = compute_domain(
+        DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF,
+        Some(fork_version),
+        None,
+    );
+
+    let signing_root = compute_signing_root(selection_data, domain);
+
+    Ok(signature.verify(validator_pubkey, signing_root.as_slice())?)
+}
```
shariqnaiyer: No need to make functions for these, if we were to use it in multiple places, we shall refactor it. 

For now we can just do it in the validate function.

crates/networking/manager/src/gossipsub/validate/sync_committee_contribution_and_proof.rs
```diff
@@ -0,0 +1,267 @@
...
+    fork_version: FixedBytes<4>,
+    selection_data: &SyncAggregatorSelectionData,
+    signature: &BLSSignature,
+    validator_pubkey: &PublicKey,
+) -> anyhow::Result<bool> {
+    let domain = compute_domain(
+        DOMAIN_SYNC_COMMITTEE_SELECTION_PROOF,
+        Some(fork_version),
+        None,
+    );
```
shariqnaiyer:
```rust
    let domain = state.get_domain(DOMAIN_BEACON_ATTESTER, Some(attestation.data.target.epoch));
    let signing_root = compute_signing_root(&attestation.data, domain);

    let signature_valid = attestation
        .signature
        .verify(&validator.public_key, signing_root.as_slice())?;
```

Use get_domain instead of compute_domain as it handles the fork version for you. 

Refer to the sync_commitee.rs or beacon_attestation.rs to see how the signature validity is checked in other places.

shariqnaiyer: I also think we don't need to define domain and signing_root as variables and we can incline the entire function like 

```rust
    let signature_valid = attestation
        .signature
        .verify(&validator.public_key, compute_signing_root(&attestation.data,  state.get_domain(DOMAIN_BEACON_ATTESTER, Some(attestation.data.target.epoch))).as_slice())?;
```

Let me know if that works

crates/networking/manager/src/gossipsub/validate/sync_committee_contribution_and_proof.rs
jveer634: Have done for some quality code. Else it became one big chunk of code that is hard to understand.

shariqnaiyer: That's good to look out for!

However as
1. You've clearly commented what you're trying to establish where you've called these functions.
2. These functions for verification are only being used once.


As other validate functions have just did it within the function and it can be converted to a one liner like this below
```rust
    let signature_valid = attestation
        .signature
        .verify(&validator.public_key, compute_signing_root(&attestation.data,  state.get_domain(DOMAIN_BEACON_ATTESTER, Some(attestation.data.target.epoch))).as_slice())?;
```

I believe it is much cleaner to and reduces around 60-70 lines and is still pretty understandable.

crates/networking/manager/src/gossipsub/validate/sync_committee_contribution_and_proof.rs
```diff
@@ -0,0 +1,238 @@
...
+    let current_epoch = state.get_current_epoch();
+    let previous_epoch = state.get_previous_epoch();
+
+    // [IGNORE] the epoch of contribution.slot is either the current or previous epoch (with a
+    // MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance)
+    if attestation_epoch != current_epoch && attestation_epoch != previous_epoch {
+        return Ok(ValidationResult::Ignore(
+            "Contribution is from a epoch too far in the past".to_string(),
+        ));
+    }
```
shariqnaiyer: This should all be under one comment right? Why are we repeating the comment

crates/networking/manager/src/gossipsub/validate/sync_committee_contribution_and_proof.rs
```diff
@@ -0,0 +1,238 @@
...
+        return Ok(ValidationResult::Reject(
+            "The contribution has too many participants".to_string(),
+        ));
+    }
+
+    // [REJECT] if is_sync_committee_aggregator(contribution_and_proof.selection_proof) is false
+    if is_sync_committee_aggregator(&contribution_and_proof.selection_proof) {
+        return Ok(ValidationResult::Reject(
+            "The selection proof is not a valid aggregator".to_string(),
+        ));
```
shariqnaiyer, line 75: I believe you meant to reject when false. You are rejecting when true (docs say to reject when true so you probably have to fix the comment)

crates/networking/manager/src/gossipsub/validate/sync_committee_contribution_and_proof.rs
```diff
@@ -0,0 +1,238 @@
...
+        &state.validators[contribution_and_proof.aggregator_index as usize].public_key;
+
+    let sync_committee_validators =
+        get_sync_subcommittee_pubkeys(&state, contribution.subcommittee_index);
+
+    if sync_committee_validators.contains(validator_pubkey) {
+        return Ok(ValidationResult::Reject(
+            "The aggregator is in the subcommittee".to_string(),
+        ));
+    }
```
shariqnaiyer, line 94: I will have to see the logic again but is the logic flipped here too? Do we have to check for "not in"?

shariqnaiyer: Maybe the comment here is wrong

crates/networking/manager/src/gossipsub/validate/sync_committee_contribution_and_proof.rs
shariqnaiyer: ```[REJECT] contribution_and_proof.selection_proof selects the validator as an aggregator for the slot -- i.e. is_sync_committee_aggregator(contribution_and_proof.selection_proof) returns True.```

I think the comment here is wrong too.

jveer634:
> @jveer634 Please ensure the comments are inline with the P2P docs and what you're commenting is what you're implementing. Please ping me when the above are fixed and I will take a look. Looking good for the most part.

I have changed the comments a bit for easier understanding. Rather than directly adding them from the docs.

jveer634:
@shariqnaiyer have a look  now. It should be good

shariqnaiyer:
@KolbyML Looks good to me

crates/networking/manager/src/gossipsub/validate/sync_committee_contribution_and_proof.rs
```diff
@@ -0,0 +1,230 @@
...
+    if !is_sync_committee_aggregator(&contribution_and_proof.selection_proof) {
+        return Ok(ValidationResult::Reject(
+            "The selection proof is not a valid aggregator".to_string(),
+        ));
+    }
+
+    // [REJECT] if the validator with index contribution_and_proof.aggregator_index is not in the
+    // sync committee for the epoch of contribution.slot
+    let validator_pubkey =
+        &state.validators[contribution_and_proof.aggregator_index as usize].public_key;
```
KolbyML: we shouldn't be doing raw parses in code which runs remote payloads, as they could cause us to panic

KolbyML (changes requested):

Issue #778, opened 2025-09-05
Can't run multiple lean nodes on the same machine after #772


unnawut:
### Describe the bug

When running multiple lean nodes on the same machine, I'm getting:

```
==> devnet/logs/node1.log <==

thread 'main' panicked at bin/ream/src/main.rs:87:41:
unable to init Ream Database: Redb(DatabaseAlreadyOpen)
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: core::result::Result<T,E>::expect
   4: ream::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

==> devnet/logs/node2.log <==

thread 'main' panicked at bin/ream/src/main.rs:87:41:
unable to init Ream Database: Redb(DatabaseAlreadyOpen)
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: core::result::Result<T,E>::expect
   4: ream::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

==> devnet/logs/node3.log <==

thread 'main' panicked at bin/ream/src/main.rs:87:41:
unable to init Ream Database: Redb(DatabaseAlreadyOpen)
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: core::result::Result<T,E>::expect
   4: ream::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

But at the same time, if I try to configure the data dir separately...

```
==> devnet/logs/node0.log <==
error: unexpected argument '--data-dir' found

Usage: ream lean_node --network <NETWORK> --validator-registry-path <VALIDATOR_REGISTRY_PATH>

For more information, try '--help'.

==> devnet/logs/node1.log <==
error: unexpected argument '--data-dir' found

Usage: ream lean_node --network <NETWORK> --validator-registry-path <VALIDATOR_REGISTRY_PATH>

For more information, try '--help'.

==> devnet/logs/node2.log <==
error: unexpected argument '--data-dir' found

Usage: ream lean_node --network <NETWORK> --validator-registry-path <VALIDATOR_REGISTRY_PATH>

For more information, try '--help'.

==> devnet/logs/node3.log <==
[...]

PR #779, opened 2025-09-06
feat: bash script to run lean nodes

varun-doshi:
## What was wrong?

After https://github.com/ReamLabs/ream/issues/778 I realized its a bit of a headache to run multiple lean nodes because we have to change the `data-dir` `socket-port` and `http-port` for each. So I created this bash script to run `n` nodes one after the other.
In case we already have a convinient way to do this that I'm unaware of, let me know and I'll close this PR.

## How was it fixed?

The script runs `n` nodes. Currenly it makes all the nodes connect to the first one(by using --bootnodes flag). Ideally, all nodes should connect to all previous nodes. But we can change that later, should not be a big deal.
Also this script is modelled for a macOS machine, have not had the chance to test in other OS.
This creates separate log files for each node, I think those need to be prettified(not sure how so will do a bit of research) or if its an overkill I can remove that.
P.S. it may need executable permission to run. Something like `chmod +x lean_run.sh` depending on your settings.

The reason I've added a `println` is because the script is unable to read `info!` logs(let me know if there is a way). It needs to detect the `local_peer_id` in order to pass it to the next node.

### DIFF
--- a/crates/networking/p2p/src/network/lean/mod.rs
+++ b/crates/networking/p2p/src/network/lean/mod.rs
@@ -182,6 +182,7 @@ impl LeanNetworkService {
         ));
         multi_addr.push(Protocol::QuicV1);
         info!("Listening on {multi_addr:?}");
+        println!("local_peer_id={:?}", lean_network_service.local_peer_id());
 
         lean_network_service
             .swarm
--- a/lean_run.sh
+++ b/lean_run.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+
+read -p "Enter number of nodes: " n
+
+if ! [[ "$n" =~ ^[0-9]+$ ]] || [ "$n" -lt 1 ]; then
+  echo "Please enter a positive integer"
+  exit 1
+fi
+
+# Open first node in a new terminal window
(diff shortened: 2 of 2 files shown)


KolbyML (changes requested):
I don't think devnet testnet infrastructure like this should exist within this repo. if you want share infrastructure like this I would recommend making a new github repo specifically for devnets or testnets.

crates/networking/p2p/src/network/lean/mod.rs
```diff
@@ -182,6 +182,7 @@ impl LeanNetworkService {
         ));
         multi_addr.push(Protocol::QuicV1);
         info!("Listening on {multi_addr:?}");
+        println!("local_peer_id={:?}", lean_network_service.local_peer_id());
```
KolbyML, line 185: The peer_id would be predefined by specifying a secret key manually no? `--secret_key_path`

syjn99:
> I don't think devnet testnet infrastructure like this should exist within this repo.

Agreed to this point.

KolbyML:
Closing this PR as devnet/testnet infrastructure like this shouldn't be in this repo

Issue #776, opened 2025-09-05
Implement `/lean/v0/node/version` endpoint


syjn99:
### Describe the feature

https://github.com/ReamLabs/ream/blob/master/crates/rpc/beacon/src/handlers/version.rs

We can make `version.rs` common, and then add an endpoint (`/lean/v0/node/version`) for lean node.

PR #780, opened 2025-09-07
feat(lean): implement `version` endpoint

Kayden-ML:
### What was wrong?

Fixes: https://github.com/ReamLabs/ream/issues/776

### How was it fixed?

Makes `version.rs` common, and then adds the endpoint (/lean/v0/node/version) for the lean node


KolbyML (approved):
:shipit: looks good

PR #781, opened 2025-09-07
feat: add cli command to generate network key

### LINKED ISSUE #765: Generate network key before running a lean node
### Describe the feature

Thanks to #754, we can provide a path (`--secret-key-path`) that has a secret key encoded with protobuf. I would like to suggest to make a way to 1) generate `secp256k1` keypair and 2) save the encoded secret key in the given path in CLI before running `lean_node`.

By having this functionality, we are able to calculate ENR before running a node.

### Additional context

_No response_

shariqnaiyer:
### What was wrong?

Fixes #765 

### How was it fixed?

Added a cli command to generate a secp256k1 keypair.

### To-Do

 
### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5185,6 +5185,7 @@ dependencies = [
  "clap",
  "discv5",
  "hashbrown 0.15.3",
+ "libp2p-identity",
  "prometheus_exporter",
  "rand 0.9.2",
  "rand_chacha 0.9.0",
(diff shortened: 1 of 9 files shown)


KolbyML:
update our cli from `--secret-key-path` to `--private-key-path`, and name this cli commannd generate private key

You can write in the cli's description what it is used for, but we don't need to make up new terminology like network key

KolbyML (approved):
:shipit: looks good

PR #783, opened 2025-09-10
feat: lean state rpc

varun-doshi:
### How was it fixed?

Adds RPC to fetch a `LeanState` for a give `State ID`

### DIFF
--- a/crates/rpc/lean/src/handlers/state.rs
+++ b/crates/rpc/lean/src/handlers/state.rs
@@ -4,7 +4,6 @@ use actix_web::{
 };
 use ream_api_types_common::{error::ApiError, id::ID};
 use ream_chain_lean::lean_chain::LeanChainReader;
-use ream_consensus_lean::state::LeanState;
 use ream_storage::tables::{field::Field, table::Table};
 
 // GET /lean/v0/states/{state_id}
@@ -13,17 +12,9 @@ pub async fn get_state(
     state_id: Path<ID>,
     lean_chain: Data<LeanChainReader>,
 ) -> Result<impl Responder, ApiError> {
-    Ok(HttpResponse::Ok().json(get_state_from_id(state_id.into_inner(), lean_chain).await?))
-}
-
-// Retrieve a state from the lean chain by its state ID.
-pub async fn get_state_from_id(
-    state_id: ID,
-    lean_chain: Data<LeanChainReader>,
-) -> Result<LeanState, ApiError> {
     let lean_chain = lean_chain.read().await;
 
-    let block_root = match state_id {
+    let block_root = match state_id.into_inner() {
         ID::Finalized => {
             let db = lean_chain.store.lock().await;
             Ok(db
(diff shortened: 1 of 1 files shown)


crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -239,4 +239,8 @@ impl LeanChain {
             .find(|block| block.slot == slot)
             .cloned()
     }
+
+    pub fn get_state_by_block(&self, block_hash: B256) -> Option<LeanState> {
```
KolbyML: this function isn't doing what the name suggestions

KolbyML (changes requested):
.

crates/rpc/lean/src/handlers/state.rs
```diff
@@ -0,0 +1,53 @@
...
+use ream_chain_lean::lean_chain::LeanChainReader;
+use ream_consensus_lean::state::LeanState;
+use tree_hash::TreeHash;
+
+// GET /lean/v0/states/{block_id}
+#[get("/states/{block_id}")]
+pub async fn get_state(
+    block_id: Path<ID>,
+    lean_chain: Data<LeanChainReader>,
+) -> Result<impl Responder, ApiError> {
```
syjn99:
```suggestion
// GET /lean/v0/states/{state_id}
#[get("/states/{state_id}")]
pub async fn get_state(
    state_id: Path<ID>,
    lean_chain: Data<LeanChainReader>,
) -> Result<impl Responder, ApiError> {
```

Ref: https://ethereum.github.io/beacon-APIs/#/Debug/getStateV2

varun-doshi: This code is not meant for State ID since we are not storing a StateID -> State mapping yet
We currently only store BlockID -> State in the `LeanChain` struct.
```
/// {block_hash: post_state} for all blocks that we know about.
    pub post_states: HashMap<B256, LeanState>,
```

StateID -> State can be added easily when I do the DB integration.
Correct me if Im wrong @syjn99

syjn99: Yes, you're right. So I think this work can be deferred after DB integration is done.

varun-doshi: @syjn99 to confirm, do you mean this PR would be blocked till after DB integration or do you mean we can merge this now and later add the `get state from state`?

syjn99: @varun-doshi ah yes, sorry for late reply 🙏 It seems quite odd for me to search a state with block hash.

varun-doshi: Done!

crates/rpc/lean/src/handlers/state.rs
```diff
@@ -0,0 +1,47 @@
...
+};
+use ream_api_types_common::{error::ApiError, id::ID};
+use ream_chain_lean::lean_chain::LeanChainReader;
+use ream_consensus_lean::state::LeanState;
+use ream_storage::tables::table::Table;
+
+// GET /lean/v0/states/{state_id}
+#[get("/states/{state_id}")]
+pub async fn get_state(
+    block_id: Path<ID>,
```
syjn99:
```suggestion
    state_id: Path<ID>,
```

crates/rpc/lean/src/handlers/state.rs
```diff
@@ -0,0 +1,70 @@
...
+use ream_consensus_lean::state::LeanState;
+use ream_storage::tables::{field::Field, table::Table};
+
+// GET /lean/v0/states/{state_id}
+#[get("/states/{state_id}")]
+pub async fn get_state(
+    state_id: Path<ID>,
+    lean_chain: Data<LeanChainReader>,
+) -> Result<impl Responder, ApiError> {
+    Ok(HttpResponse::Ok().json(get_state_from_id(state_id.into_inner(), lean_chain).await?))
```
syjn99: Just a nit, but is there any reason for having a separate function(`get_state_from_id`)?

varun-doshi: I guess it was mostly to keep the code flow similar to the beacon counterparts, but I think in this case we can remove the extra function

PR #786, opened 2025-09-16
feat: Integrate DB to LeanChain

### LINKED ISSUE #777: Integrate DB in Lean Chain working
### Describe the feature

Integrate the DB into the flow of the Lean node and replace the in-memory stores

### Additional context

_No response_

varun-doshi:
Fixes #777 

There's also an issue where justifications and finalisations are jumping i.e. latest finalized block goes from 7 to 11. Stays at 7 for longer than should be and then jumps to 11 a little later. Trying to figure out whats going on here.

I felt the following can stay in memory and the reasoning:

- `new_votes` : these would not be a very long list at any given point since we drain them. Although I think eventually we'll need to put them in db but I'll wait for comments on this.

- `genesis_hash`, `num_validators`: mostly constants set during startup. Since they are read during the lean flow, the db read up latency might not make it worth putting in db.
- `safe_target`, `head`: Updated quite often but these are singular values and the same reasoning as above with the addition of write latency as well


crates/common/consensus/lean/src/state.rs
```diff
@@ -162,11 +162,11 @@ impl LeanState {
 
         // Process slots (including those with no blocks) since block
         self.process_slots(block.slot)
-            .context("Failed to process slots")?;
+            .map_err(|error| anyhow!("Failed to process slots: {error:?}"))?;
 
         // Process block
         self.process_block(block)
-            .context("Failed to process block")?;
+            .map_err(|error| anyhow!("Failed to process block: {error:?}"))?;
```
syjn99: Any reason for this change? `context()` provides similar DX.

crates/common/consensus/lean/src/block.rs
```diff
@@ -17,6 +17,15 @@ pub struct SignedBlock {
...
 }
 
+impl From<Block> for SignedBlock {
+    fn from(block: Block) -> Self {
+        Self {
+            message: block,
+            signature: FixedBytes::default(),
+        }
+    }
+}
```
syjn99: Hmm this seems counter intuitive, `Block -> SignedBlock` isn't one-to-one relation. Maybe we add other method under `SignedBlock` with more proper name? I just want to avoid `From` trait here.

crates/common/chain/lean/src/genesis.rs
```diff
@@ -20,19 +20,19 @@ fn genesis_state(num_validators: u64, genesis_time: u64) -> LeanState {
 /// Setup the genesis block and state for the Lean chain.
 ///
 /// Reference: https://github.com/ethereum/research/blob/d225a6775a9b184b5c1fd6c830cc58a375d9535f/3sf-mini/test_p2p.py#L119-L131
-pub fn setup_genesis() -> (Block, LeanState) {
+pub fn setup_genesis() -> (SignedBlock, LeanState) {
```
syjn99:
```suggestion
pub fn setup_genesis() -> (Block, LeanState) {
```

Could you please wrap `Block` into `SignedBlock` if it is needed? I think this function should return pure `Block`.

crates/rpc/lean/src/handlers/block.rs
```diff
@@ -23,20 +24,35 @@ pub async fn get_block(
 pub async fn get_block_by_id(
     block_id: ID,
     lean_chain: Data<LeanChainReader>,
-) -> Result<Option<Block>, ApiError> {
-    // Obtain read guard first from the reader.
-    let lean_chain = lean_chain.read().await;
+) -> Result<Option<SignedBlock>, ApiError> {
```
syjn99:
```suggestion
) -> Result<Option<Block>, ApiError> {
```

Let's just return `Block` instead of `SignedBlock`

crates/rpc/lean/Cargo.toml
```diff
@@ -11,6 +11,7 @@ version.workspace = true
 
 [dependencies]
 actix-web.workspace = true
+alloy-primitives.workspace = true
```
syjn99: It's okay not to define `alloy-primitives` deps explicitly, let's remove it.

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -47,72 +43,130 @@ pub struct LeanChain {
...
-        let justified_hash = get_latest_justified_hash(&self.post_states)
-            .ok_or_else(|| anyhow!("No justified hash found in post states"))?;
+    pub async fn compute_safe_target(&self) -> anyhow::Result<B256> {
+        let justified_hash = self.latest_justified_hash().await?;
 
         get_fork_choice_head(
-            &self.chain,
+            self.store.clone(),
             &justified_hash,
-            &self.new_votes,
```
syjn99, line 84: Don't we have to use `new_votes` for this case? `get_fork_choice_head` will always fetch data from `KnownVotesTable`.

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -173,70 +221,72 @@ impl LeanChain {
...
+    pub async fn build_vote(&self, slot: u64) -> anyhow::Result<Vote> {
+        let lean_block_provider = {
+            let db = self.store.lock().await;
+            db.lean_block_provider()
+        };
+
+        let lean_state_provider = {
+            let db = self.store.lock().await;
+            db.lean_state_provider()
+        };
```
syjn99: How about taking same approach as `get_fork_choice_head` to acquire lock only once?

crates/storage/src/tables/lean/new_votes.rs
```diff
@@ -0,0 +1,55 @@
...
+/// Key: index (u64, acts like position in an append-only array)
+/// Value: `SignedVote`
+pub(crate) const NEW_VOTES_TABLE: TableDefinition<u64, SSZEncoding<SignedVote>> =
+    TableDefinition::new("new_votes");
+
+pub struct NewVotesTable {
+    pub db: Arc<Database>,
+}
+
+impl NewVotesTable {
```
syjn99: > new_votes : these would not be a very long list at any given point since we drain them. Although I think eventually we'll need to put them in db but I'll wait for comments on this.

^ I think we don't need this table as well.

syjn99 (commented):
Overall good, but I think we should pass different vector of `Vote` for `get_fork_choice_head`. (not impacting the current fork choice code then.

crates/common/consensus/lean/src/state.rs
varun-doshi: context does not actually pass up the `err` it will show as `Failed to process slots` regardless of whatever the actual error is underneath which makes it harder to debug

KolbyML: good catch Varun, we should use the |err| shorthand tho

crates/common/chain/lean/src/lean_chain.rs
varun-doshi: Ahh yes

varun-doshi:
Applied the above suggestions.
Now going back to figuring out why the finalization jump issue is happening

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -150,7 +151,8 @@ impl LeanChain {
     /// Done upon processing new votes or a new block
     pub async fn recompute_head(&mut self) -> anyhow::Result<()> {
         let justified_hash = self.latest_justified_hash().await?;
-        self.head = get_fork_choice_head(self.store.clone(), &justified_hash, 0).await?;
+        self.head =
+            get_fork_choice_head(self.store.clone(), &self.new_votes, &justified_hash, 0).await?;
```
syjn99: This should be `known_votes`.

syjn99 (changes requested):

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -47,72 +43,132 @@ pub struct LeanChain {
...
+        self.recompute_head().await?;
         Ok(())
     }
 
     /// Done upon processing new votes or a new block
-    pub fn recompute_head(&mut self) -> anyhow::Result<()> {
-        let justified_hash = get_latest_justified_hash(&self.post_states)
-            .ok_or_else(|| anyhow!("Failed to get latest_justified_hash from post_states"))?;
-        self.head = get_fork_choice_head(&self.chain, &justified_hash, &self.known_votes, 0)?;
+    pub async fn recompute_head(&mut self) -> anyhow::Result<()> {
```
syjn99, line 144:
```rust
/// Done upon processing new votes or a new block
pub async fn recompute_head(&mut self) -> anyhow::Result<()> {
    let justified_hash = self.latest_justified_hash().await?;
    let known_votes_provider = {
        let db = self.store.lock().await;
        db.known_votes_provider()
    };

    let votes = known_votes_provider.all_sorted_by_slot()?;
    self.head = get_fork_choice_head(self.store.clone(), &votes, &justified_hash, 0).await?;
    info!("new head: {:?}", self.head);
    Ok(())
}
```

This must be enough, tested with my local branch!

crates/common/chain/lean/src/lean_chain.rs
syjn99: Resolving this with code suggestion: https://github.com/ReamLabs/ream/pull/786#discussion_r2372475581

crates/storage/src/tables/lean/new_votes.rs
syjn99, line 1: Could you delete this file if it isn't needed?

crates/storage/src/tables/lean/known_votes.rs
```diff
@@ -0,0 +1,144 @@
...
+/// Key: index (u64, acts like position in an append-only array)
+/// Value: `SignedVote`
+pub(crate) const KNOWN_VOTES_TABLE: TableDefinition<u64, SSZEncoding<SignedVote>> =
+    TableDefinition::new("known_votes");
+
+pub struct KnownVotesTable {
+    pub db: Arc<Database>,
+}
+
+impl KnownVotesTable {
```
syjn99, line 20: Size of this table will monotonically increase since there's no pruning logic here. I think we don't have to add its functionality in this PR, you might add an issue.

crates/storage/src/tables/lean/known_votes.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    /// Returns if there are no known votes
+    pub fn is_empty(&self) -> Result<bool, StoreError> {
+        let read_txn = self.db.begin_read()?;
+        let table = read_txn.open_table(KNOWN_VOTES_TABLE)?;
+        Ok(table.len()? == 0)
+    }
+
+    /// Load all votes sorted by slot
+    pub fn all_sorted_by_slot(&self) -> Result<Vec<SignedVote>, StoreError> {
```
syjn99: We don't have to sort the votes, as `get_fork_choice_head` will handle it.

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -47,72 +42,139 @@ pub struct LeanChain {
 }
 
 impl LeanChain {
-    pub fn new(genesis_block: Block, genesis_state: LeanState, db: LeanDB) -> LeanChain {
-        let genesis_hash = genesis_block.tree_hash_root();
+    pub fn new(genesis_block: SignedBlock, genesis_state: LeanState, db: LeanDB) -> LeanChain {
+        let genesis_block_hash = genesis_block.message.tree_hash_root();
+        let no_of_validators = genesis_state.config.num_validators;
+        // todo: remove this unwrap
```
syjn99:
```suggestion
```

syjn99 (commented):
Could you update the PR description as well?

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -47,72 +42,139 @@ pub struct LeanChain {
...
+        let lean_state_provider = {
+            let db = self.store.lock().await;
+            db.lean_state_provider()
+        };
+
+        let state = lean_state_provider
+            .get(self.head)?
+            .ok_or_else(|| anyhow!("State not found in chain for head: {}", self.head))?;
+
+        Ok(state.latest_justified.root)
```
KolbyML:
```suggestion
    pub async fn latest_justified_hash(&self) -> anyhow::Result<B256> {
        self.store.lock().await.lean_state_provider().get(self.head)?
            .ok_or_else(|| anyhow!("State not found in chain for head: {}", self.head)).map(|state| state.latest_justified.root)
```

we can just inline all of this

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -47,72 +42,139 @@ pub struct LeanChain {
...
-            .map(|state| state.latest_finalized.root)
+    pub async fn get_block_id_by_slot(&self, slot: u64) -> anyhow::Result<B256> {
+        let lean_slot_provider = {
+            let db = self.store.lock().await;
+            db.slot_index_provider()
+        };
+
+        lean_slot_provider
+            .get(slot)?
+            .ok_or_else(|| anyhow!("Block not found in chain for head: {}", self.head))
```
KolbyML, line 81: this can be inlined

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -47,72 +42,139 @@ pub struct LeanChain {
...
+        let lean_state_provider = {
+            let db = self.store.lock().await;
+            db.lean_state_provider()
+        };
+
+        let state = lean_state_provider
+            .get(self.head)?
+            .ok_or_else(|| anyhow!("State not found in chain for head: {}", self.head))?;
+
+        Ok(state.latest_finalized.root)
```
KolbyML: this can be inlined

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -47,72 +42,139 @@ pub struct LeanChain {
...
     }
 
     /// Compute the latest block that the staker is allowed to choose as the target
-    pub fn compute_safe_target(&self) -> anyhow::Result<B256> {
-        let justified_hash = get_latest_justified_hash(&self.post_states)
-            .ok_or_else(|| anyhow!("No justified hash found in post states"))?;
+    pub async fn compute_safe_target(&self) -> anyhow::Result<B256> {
+        let justified_hash = self.latest_justified_hash().await?;
 
         get_fork_choice_head(
```
KolbyML, line 113: It kinda of feels like this function should be a member of this struct, but this isn't the biggest concern for this PR, we can bring it up again possibly in the future.

crates/common/chain/lean/src/service.rs
```diff
@@ -65,9 +66,13 @@ impl LeanChainService {
...
-                            let lean_chain = self.lean_chain.read().await;
-                            let head_state = lean_chain.post_states.get(&lean_chain.head)
-                                .ok_or_else(|| anyhow!("Post state not found for head: {}", lean_chain.head))?;
+                            let (head, store) = {
+                                let lean_chain = self.lean_chain.read().await;
+                                (lean_chain.head, lean_chain.store.clone())
+                            };
+                            let head_state = store.lock().await
+                                .lean_state_provider()
+                                .get(head)?.ok_or_else(|| anyhow!("Post state not found for head: {}", head))?;
```
```suggestion
                                .get(head)?.ok_or_else(|| anyhow!("Post state not found for head: {head}"))?;
```

crates/common/chain/lean/src/service.rs
```diff
@@ -258,18 +279,21 @@ impl LeanChainService {
             // TODO: Validate the signature.
         }
 
-        let lean_chain = self.lean_chain.read().await;
-        let is_known_vote = lean_chain.known_votes.contains(&signed_vote);
-        let is_new_vote = lean_chain.new_votes.contains(&signed_vote);
+        let (lean_block_provider, known_votes_provider) = {
+            let lean_chain = self.lean_chain.read().await;
+            let db = lean_chain.store.lock().await;
```
KolbyML, line 284:
```suggestion
        let (lean_block_provider, known_votes_provider) = {
            let db = self.lean_chain.read().await.store.lock().await;
```

that variable is only used onced, might as well not declare it

crates/common/chain/lean/src/service.rs
```diff
@@ -258,18 +279,21 @@ impl LeanChainService {
...
+            (db.lean_block_provider(), db.known_votes_provider())
+        };
+
+        let is_known_vote = known_votes_provider.contains(&signed_vote)?;
+        let is_new_vote = {
+            let lean_chain = self.lean_chain.read().await;
+            lean_chain.new_votes.contains(&signed_vote)
+        };
```
KolbyML, line 295:
```suggestion
        let is_new_vote = self.lean_chain.read().await.new_votes.contains(&signed_vote);
```

crates/common/consensus/lean/src/block.rs
```diff
@@ -17,6 +17,15 @@ pub struct SignedBlock {
     pub signature: FixedBytes<4000>,
 }
 
+impl SignedBlock {
+    pub fn default_from_block(block: Block) -> Self {
+        Self {
+            message: block,
+            signature: FixedBytes::default(),
+        }
+    }
```
KolbyML: This is a little awkward, as default_from_block isn't descriptive of this containing 0's for the signature, maybe make this conversion a member of the Block struct and call it, `to_signed_block_with_empty_signature()`? that is a bit long, or just inline it since it is only used once with a comment

crates/common/consensus/lean/src/state.rs
unnawut: Interesting! I think it's because of this though, not that it's dropping the underneath error: https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations

> When you print an error object using “{}” or to_string(), only the outermost underlying error or context is printed, not any of the lower level causes. This is exactly as if you had called the Display impl of the error from which you constructed your anyhow::Error.
>
> Failed to read instrs from ./path/to/instrs.json
> To print causes as well using anyhow’s default formatting of causes, use the alternate selector “{:#}”.
>
> Failed to read instrs from ./path/to/instrs.json: No such file or directory (os error 2)

So it might be up to the final error handler to format. But I havn't checked thoroughly yet so I might be wrong.

crates/rpc/lean/src/handlers/block.rs
```diff
@@ -24,19 +25,34 @@ pub async fn get_block_by_id(
...
+            ID::Head => Ok(lean_chain.head),
+            ID::Justified => lean_chain.latest_justified_hash().await.map_err(|err| {
+                ApiError::InternalError(format!("No latest justified hash: {err:?}"))
+            }),
+            ID::Slot(slot) => lean_chain.get_block_id_by_slot(slot).await.map_err(|err| {
+                ApiError::InternalError(format!("No block for slot {slot}: {err:?}"))
+            }),
+            ID::Root(root) => Ok(root),
+        }
+    };
```
KolbyML:
```suggestion
    
    let lean_chain = lean_chain.read().await;
    let block_root = match block_id {
            ID::Finalized => lean_chain.latest_finalized_hash().await.map_err(|err| {
                ApiError::InternalError(format!("No latest finalized hash: {err:?}"))
            }),
            ID::Genesis => Ok(lean_chain.genesis_hash),
            ID::Head => Ok(lean_chain.head),
            ID::Justified => lean_chain.latest_justified_hash().await.map_err(|err| {
                ApiError::InternalError(format!("No latest justified hash: {err:?}"))
            }),
            ID::Slot(slot) => lean_chain.get_block_id_by_slot(slot).await.map_err(|err| {
                ApiError::InternalError(format!("No block for slot {slot}: {err:?}"))
            }),
            ID::Root(root) => Ok(root),
        };
```

the scoping is not needed in this context right?

crates/rpc/lean/src/handlers/block.rs
```diff
@@ -24,19 +25,34 @@ pub async fn get_block_by_id(
...
-        ID::Slot(slot) => lean_chain.get_block_by_slot(slot),
-        ID::Root(root) => lean_chain.get_block_by_root(root),
-    })
+    let db = {
+        let lean_chain = lean_chain.read().await;
+        lean_chain.store.clone()
+    };
+
+    let provider = db.lock().await.lean_block_provider();
+    provider
```
KolbyML:
```suggestion
        lean_chain.read().await.store.lock().await.lean_block_provider()
```

can't we simplify the code to this?

crates/storage/src/tables/lean/lean_block.rs
```diff
@@ -58,3 +58,35 @@ impl Table for LeanBlockTable {
...
+        &self,
+        min_score: u64,
+        vote_weights: &HashMap<B256, u64>,
+    ) -> Result<HashMap<B256, Vec<B256>>, StoreError> {
+        let mut children_map = HashMap::<B256, Vec<B256>>::new();
+        let read_txn = self.db.begin_read()?;
+        let table = read_txn.open_table(LEAN_BLOCK_TABLE)?;
+
+        for entry in table.iter()? {
+            let (hash_enc, block_enc) = entry?;
```
KolbyML: what does enc mean ?, not the most descriptive name

crates/storage/src/tables/lean/lean_state.rs
```diff
@@ -43,3 +43,27 @@ impl Table for LeanStateTable {
...
+
+impl LeanStateTable {
+    pub fn get_latest_justified_hash(&self) -> Result<Option<B256>, StoreError> {
+        let read_txn = self.db.begin_read()?;
+        let table = read_txn.open_table(LEAN_STATE_TABLE)?;
+
+        let mut best: Option<(u64, B256)> = None;
+
+        for entry in table.iter()? {
+            let (_key_enc, state_enc) = entry?;
```
KolbyML:
```suggestion
            let (_, state_enc) = entry?;
```

and same comment about enc

crates/storage/src/tables/lean/lean_state.rs
```diff
@@ -43,3 +43,27 @@ impl Table for LeanStateTable {
         Ok(())
     }
 }
+
+impl LeanStateTable {
+    pub fn get_latest_justified_hash(&self) -> Result<Option<B256>, StoreError> {
+        let read_txn = self.db.begin_read()?;
+        let table = read_txn.open_table(LEAN_STATE_TABLE)?;
+
+        let mut best: Option<(u64, B256)> = None;
```
KolbyML: is this type declaration needed?

crates/storage/src/tables/lean/lean_state.rs
```diff
@@ -43,3 +43,27 @@ impl Table for LeanStateTable {
...
+            let (_key_enc, state_enc) = entry?;
+            let state: LeanState = state_enc.value();
+
+            let slot = state.latest_justified.slot;
+            let root = state.latest_justified.root;
+
+            match best {
+                Some((best_slot, _)) if slot <= best_slot => {}
+                _ => best = Some((slot, root)),
+            }
```
KolbyML:
```suggestion
    if best.map_or(true, |(best_slot, _)| slot > best_slot) {
        best = Some((slot, root));
    }
```

^ this should work

crates/common/chain/lean/src/service.rs
varun-doshi: does not allow

`temporary value dropped while borrowed`

crates/storage/src/tables/lean/known_votes.rs
varun-doshi: Done https://github.com/ReamLabs/ream/issues/805

KolbyML (approved):

Issue #771, opened 2025-09-03
Make `HashSigScheme` configurable & Match with current leanSpec


syjn99:
### Describe the feature

> [!NOTE]
> This issue is **NOT** for PQ Devnet 0, but can be parallelized.

https://github.com/leanEthereum/leanSpec/blob/a2bc45b66b1fa8506dfae54f9966563d1e54101c/src/lean_spec/subspecs/xmss/constants.py#L102-L137

In current leanSpec, you can find configuration sets for production(`PROD_CONFIG`)/testing(`TEST_CONFIG`) respectively.

https://github.com/ReamLabs/ream/blob/8bfe85b495a610d79f3e35e91dfa4c92e9e5a10d/crates/crypto/post_quantum/src/hashsig/mod.rs#L8

We're currently defining our default signature scheme like above. However, I'm quite sure we *won't* use that scheme (`SIGWinternitzLifetime18W4`) as it is far beyond how the latest discussion is going on.

- `PROD_CONFIG` is equal to `SIGTopLevelTargetSumLifetime32Dim64Base8`
- `TEST_CONFIG` doesn't match with existing signature scheme in `hash-sig` but we can just define it.

I would like to suggest that: **we should be able to plug in-and-out which signature scheme to use**. Or at least we should support a very tiny version of signature scheme. I guess #668 partially intended to do this.

syjn99:
@shariqnaiyer You can take this after #765 is finished.

Timosdev99:
Are you still working on this?

shariqnaiyer:
> Are you still working on this?

Yes I shall be making a PR within the next day.

PR #791, opened 2025-09-19
feat: intialize cargo features in post-quantum to use signature schemes based on feature

shariqnaiyer:
### What was wrong?

Fixes: #771 

### How was it fixed?

I initially tried using an enum to select the signature scheme and then trying to use a config.rs file to set a PROD or TEST or some other config. 

I found those to be unclean solutions so I then moved to looking into cargo features. I found that to be a pretty clean solution to making hash-sig configurable. 

I am looking for input for this approach from @syjn99 and @KolbyML. I shall add some more documentation if this is good.

### To-Do

 
### DIFF
--- a/crates/crypto/post_quantum/src/hashsig/scheme.rs
+++ b/crates/crypto/post_quantum/src/hashsig/scheme.rs
@@ -7,8 +7,9 @@ use hashsig::{
     },
 };
 
-/// TEST_CONFIG signature scheme parameters based on leanSpec configuration
-/// Source: https://github.com/leanEthereum/leanSpec/blob/a2bc45b66b1fa8506dfae54f9966563d1e54101c/src/lean_spec/subspecs/xmss/constants.py#L121-L137
+// TEST_CONFIG signature scheme parameters based on leanSpec configuration
+// Source: https://github.com/leanEthereum/leanSpec/blob/a2bc45b66b1fa8506dfae54f9966563d1e54101c/src/lean_spec/subspecs/xmss/constants.py#L121-L137
+
 const LOG_LIFETIME: usize = 8;
 const DIMENSION: usize = 16;
 const BASE: usize = 4;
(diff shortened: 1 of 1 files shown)


crates/crypto/post_quantum/src/hashsig/scheme.rs
```diff
@@ -0,0 +1,50 @@
...
+    MSG_LEN_FE,
+    PARAMETER_LEN,
+    RAND_LEN_FE,
+>;
+type TH = PoseidonTweakHash<PARAMETER_LEN, HASH_LEN_FE, TWEAK_LEN_FE, CAPACITY, DIMENSION>;
+
+#[allow(clippy::upper_case_acronyms)]
+type PRF = ShakePRFtoF<HASH_LEN_FE>;
+
+type IE = TargetSumEncoding<MH, TARGET_SUM>;
```
KolbyML: is it possible to use more descriptive const and type names

shariqnaiyer: Looks better now.

syjn99: IMHO it seems better to follow what `hash-sig` does: https://github.com/b-wagn/hash-sig/blob/84dd4564a17383c154d0e4c25a248c601a1b8914/src/signature/generalized_xmss/instantiations_poseidon_top_level.rs#L11-L47

`crate/crypto/post_quantum/src/hashsig` is totally dependent on `hash-sig` library, which means if we use different notation, it might be confusing.

KolbyML: I don't know why we would follow naming paradigms decided on by researchers (often not known for readable names). what is confusing is seeing FE, what does it stand for Fan Edition? https://www.samsung.com/uk/mobile-phone-buying-guide/what-is-samsung-fe/

I don't see how using more descriptive names would make it more confusing

KolbyML: I don't want to read a math proof etc etc, leave the x,y,z's for the papers

KolbyML: IE = internet explorer?, etc

PR #802, opened 2025-09-21
refactor(account_manager): dedup keystore implementation

### LINKED ISSUE #801: Refactor keystore: Reuse existing keystore code to implement new keystore
### Describe the feature

https://github.com/ReamLabs/ream/blob/master/crates/crypto/keystore/src/keystore.rs

Let's follow this approach for lesser code

### Additional context

_No response_

syjn99:
### What was wrong?

 

Fixes #801.

### How was it fixed?

 

- Use existing code that is more rusty (We don't need explicit validation logic, just use Rust compiler), and dedup newer implementation.
  - Note that the default for `EncryptedKeystore` is still BLS12-381 Keystore.
- Add `crates/crypto/keystore/assets/PostQuantumKeystore.json` for deserialization test.

### Note

Some of my personal opinions, please review this @ch4r10t33r  (~if you want, I can open a PR for it~ -> I left some comments there.)

- We don't need `Full` notation for KDF params.
- Matched with [EIP-2335](https://eips.ethereum.org/EIPS/eip-2335).
  - `ciphertext` -> `message` field
  - Changed `nonce` -> `iv`
- Removed `quantum_secure`: `version` is expressive enough.

### To-Do

 

crates/crypto/keystore/src/keystore.rs
```diff
@@ -94,21 +78,74 @@ impl EncryptedKeystore {
...
+        path: Option<String>,
+    ) -> Self {
+        EncryptedKeystore {
+            crypto: CryptoV5 {
+                kdf: FunctionBlock {
+                    params: KdfParams::Argon2Id {
+                        m: 65536,
+                        t: 4,
+                        p: 2,
+                        salt: rand::random::<[u8; 32]>().to_vec(),
```
ch4r10t33r, line 102: This salt be derived from the `seed_phrase`.

ch4r10t33r: Otherwise how do we make it deterministic? We can endup with a different keystore file for the same seed_phrase.

crates/crypto/keystore/src/keystore.rs
```diff
@@ -94,21 +78,74 @@ impl EncryptedKeystore {
...
+                    },
+                    message: vec![], // Empty message
+                },
+                cipher: FunctionBlock {
+                    params: CipherParams::Aes256Gcm {
+                        iv: rand::random::<[u8; 12]>().to_vec(),
```
ch4r10t33r, line 108: same as above. I can address this in my PR if required.

crates/crypto/keystore/src/keystore.rs
```diff
@@ -94,21 +78,74 @@ impl EncryptedKeystore {
...
+                keytype: FunctionBlock {
+                    params: KeyTypeParams::XmssPoseidon2OtsSeed {
+                        lifetime,
+                        activation_epoch,
+                    },
+                    message: vec![],
+                },
+            },
+            description: description.unwrap_or_default(),
+            public_key: None,
```
ch4r10t33r: Why are we setting the public key as none? Shouldn't this be the root public key?

crates/crypto/keystore/src/keystore.rs
syjn99: No, salt should be purely random value, not deriving from an user input. (And also `from_seed_phrase`  is just a migration from your PR)

crates/crypto/keystore/src/keystore.rs
syjn99: No, IV shouldn't be derived from certain user input.

crates/crypto/keystore/src/keystore.rs
syjn99: Actually this is just a migration from your PR, and `public_key` is not included in your EIP suggestion.

crates/crypto/keystore/src/keystore.rs
syjn99: > Otherwise how do we make it deterministic?

Most of encryption primitives isn't deterministic, that's why we use salt and IV.

ch4r10t33r: My thinking was to re-use the seed_phrase as the salt for all cryptographic operations within the keystore. But, I see your point on using different values.

crates/crypto/keystore/src/keystore.rs
ch4r10t33r: I am wondering if we should include it. We can report value after successful key generation?

ch4r10t33r: However, I think that can done in the next change. This looks good for now.

ch4r10t33r (approved):
Looks good.

PR #804, opened 2025-09-23
chore: add clippy rules for codebase consistency

syjn99:
### What was wrong?

 

Our PR reviews often contain this kind of comment, let Clippy handle it.

### How was it fixed?

 

- Use [clippy lints](https://rust-lang.github.io/rust-clippy/master/index.html) and [lint configurations](https://doc.rust-lang.org/clippy/lint_configuration.html).
- Add [cargo-workspace-lints](https://crates.io/crates/cargo-workspace-lints) on current CI.

#### Rationale

- `let_and_return` ([Link](https://rust-lang.github.io/rust-clippy/master/#let_and_return)): Few cases for assigning variables before returning.
- `uninlined_format_args` ([Link](https://rust-lang.github.io/rust-clippy/master/#uninlined_format_args)): Inline arguments as possible as we can. Note that this rule couldn't detect macros from `tracing`.
- `unnecessary_debug_formatting` ([Link](https://rust-lang.github.io/rust-clippy/master/#unnecessary_debug_formatting)): Use `path.display()` for printing the file path.
- `unwrap_used` ([Link](https://rust-lang.github.io/rust-clippy/master/#unwrap_used)): Stop using `unwrap()` for production code, use `expect()` instead.
- `useless_let_if_seq` ([Link](https://rust-lang.github.io/rust-clippy/master/#useless_let_if_seq)): I think this style is better :)
- `wildcard_imports` ([Link](https://rust-lang.github.io/rust-clippy/master/#wildcard_imports)): This rule reminds me of [this comment](https://github.com/ReamLabs/ream/pull/757#discussion_r2356919550), which is reasonable.

#### Side note

I guess `needless_pass_by_value` ([Link](https://rust-lang.github.io/rust-clippy/master/#needless_pass_by_value)) can be helpful and is often a topic for our PR comments, but this leads so many LOC to satisfy the linter. I'd like the team to share opinions on it.

### To-Do

 

.clippy.toml
```diff
@@ -1,3 +1,4 @@
+allow-mixed-uninlined-format-args = false
 # This feature is in Rust nightly, as of v1.62. The line is added in anticipation.
```
KolbyML, line 3: is it possible to add a new line between this variable and the comment below

clippy.toml
KolbyML, line 1: this file shouldn't be hidden no?

bin/ream/Cargo.toml
```diff
@@ -14,6 +14,9 @@ version.workspace = true
 name = "ream"
 path = "src/main.rs"
 
+[lints]
+workspace = true
```
KolbyML:
```suggestion
lints.workspace = true
```

shouldn't we just inline this, we we will never add another configuration other then workspace = true

syjn99: Resolve this after internal convo

clippy.toml
syjn99: Hmm I don't have strong opinion between `clippy.toml` and `.clippy.toml`, but we at least need a common TOML file for linting rules.

KolbyML: I think it shouldn't be hidden then as our rust fmt config isn't hidden

syjn99:
Converting to draft as I'm adding more rules.

clippy.toml
syjn99: It isn't hidden right now, do you want me to change its name without leading dot? (`clippy.toml`)

unnawut: Yeah the dot prefix makes it hidden on linux and macos. Let's remove the dot prefix!

syjn99: Ah I was confused with gitignore, thanks!

jihoonsong:
This is very nice. This doesn't have to be perfect imo. Once this change is in place, if it feels like over 80 out of 100 to us, that’s good; if it feels over 90, that’s awesome.

KolbyML (approved):
:shipit: looks good, we should use needless_pass_by_value

KolbyML:
actually there is a few cases were it wouldn't make sense and I am not sure if the linter is able to pick that up or not so, never mind

PR #806, opened 2025-09-25
chore: use anyhow context with proper formatting

unnawut:
### What was wrong?

There was a discussion at https://github.com/ReamLabs/ream/pull/786#discussion_r2357723529 that using `context()` is, well... dropping the context. After a bit of digging, I think it's just the formatting choice.

This PR make changes so that the final error is formatted with `{:?}` instead of `{}` so that its contexts are propagated and formatted properly without having to use `map_err()`.

### How was it fixed?

From [`anyhow` docs](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations):

> When you print an error object using “{}” or to_string(), only the outermost underlying error or context is printed, not any of the lower level causes.
> ```
> Failed to read instrs from ./path/to/instrs.json
> ```

**Using `{:#}` can achieve the same result as `map_err(|err| anyhow!("Failed to create clock interval: {err}"))`**, but a lot cleaner:

> To print causes as well using anyhow’s default formatting of causes, use the alternate selector “{:#}”.
> ```
> Failed to read instrs from ./path/to/instrs.json: No such file or directory (os error 2)
> ```

But I opted for a bit more readability with `{:?}`:

> The Debug format “{:?}” includes your backtrace if one was captured. Note that this is the representation you get by default if you return an error from `fn main` instead of printing it explicitly yourself.
> ```
> Error: Failed to read instrs from ./path/to/instrs.json
> 
> Caused by:
>     No such file or directory (os error 2)
> ```

This way we're properly propagating the error struct up the call stack rather than stringifying and recreating the error object at each level.

### Examples

Before:
```
thread 'tokio-runtime-worker' panicked at bin/ream/src/main.rs:236:13:
Chain service exited with error: Failed to create clock interval: Genesis time is 1757356226 but should be greater than 1758782173: second time provided was later than self
```

After:
```
thread 'tokio-runtime-worker' panicked at bin/ream/src/main.rs:236:13:
Chain service exited with error: Failed to create clock interval

Caused by:
    Genesis time is 1758782999 but should be greater than 1758783005: second time provided was later than self
```

Before:
```
2025-09-25T06:58:21.013786Z ERROR ream_chain_lean::service: Failed to handle produce block message: Failed to process slots: State slot must be less than block slot

thread 'tokio-runtime-worker' panicked at /Users/unnawut/Development/repo/ream/crates/common/validator/lean/src/service.rs:74:58:
Failed to receive block from LeanChainService: RecvError(())
```

After:
```
2025-09-25T06:51:38.015937Z ERROR ream_chain_lean::service: Failed to handle produce block message: Failed to process slots

Caused by:
    State slot must be less than block slot

thread 'tokio-runtime-worker' panicked at /Users/unnawut/Development/repo/ream/crates/common/validator/lean/src/service.rs:74:58:
Failed to receive block from LeanChainService: RecvError(())
```

### To-Do

 
### DIFF
--- a/crates/common/chain/lean/src/service.rs
+++ b/crates/common/chain/lean/src/service.rs
@@ -56,8 +56,8 @@ impl LeanChainService {
 
         let mut tick_count = 0u64;
 
-        let mut interval = create_lean_clock_interval()
-            .context("Failed to create clock interval")?;
+        let mut interval =
+            create_lean_clock_interval().context("Failed to create clock interval")?;
 
         loop {
             tokio::select! {
--- a/crates/common/consensus/lean/src/state.rs
+++ b/crates/common/consensus/lean/src/state.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 
 use alloy_primitives::B256;
-use anyhow::{Context, anyhow,ensure};
+use anyhow::{Context, anyhow, ensure};
 use itertools::Itertools;
 use ream_consensus_misc::constants::lean::{MAX_HISTORICAL_BLOCK_HASHES, VALIDATOR_REGISTRY_LIMIT};
 use ream_metrics::{FINALIZED_SLOT, HEAD_SLOT, JUSTIFIED_SLOT, set_int_gauge_vec};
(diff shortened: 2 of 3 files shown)


KolbyML (approved):
I haven't fully looked into context yet, but if you say this is good I am down for it, although personally I will avoid context in the future, as I think map_err is more readable, and simplier as the documentation is better for it

Issue #787, opened 2025-09-17
Double-check implementation with current `leanSpec`.


syjn99:
### Describe the feature

Let's have a look for our current implementation with current `leanSpec`.

https://github.com/leanEthereum/leanSpec/tree/main/docs/client

syjn99:
Let me check this after #786 is merged

PR #811, opened 2025-09-25
feat(lean): add `latest_justified` and `latest_finalized` Field & match function names with `leanSpec`

syjn99:
### What was wrong?

 

Part of #787.

Current implementations of `latest_justified_hash` and `latest_finalized_hash` are actually returning the value from `head` state, which deviates from the spec.

Also, I added/modified some function names like `update_safe_target` to match with current `leanSpec`. (https://github.com/leanEthereum/leanSpec)

### How was it fixed?

 

- Add two `Field` for lean DB: `latest_finalized` and `latest_justified`.
- Refactor existing functions a bit.

### Note

You can run a local ephemery devnet with this command:

```bash
cargo run --release -- --ephemeral lean_node \
    --network ephemery \
    --validator-registry-path ./bin/ream/assets/lean/validator_registry.yml
```

### Next work will be...

Removing `dependencies` in  `LeanChainService`.

### To-Do

 

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -49,8 +52,14 @@ impl LeanChain {
             .insert(genesis_block_hash, genesis_block)
             .expect("Failed to insert genesis block");
         db.lean_state_provider()
-            .insert(genesis_block_hash, genesis_state)
+            .insert(genesis_block_hash, genesis_state.clone())
```
varun-doshi: I think we can avoid the state clone if we rearrange and put it at the end. I dont have a strong opinion on this because genesis state is small but you can make the change if you want to

varun-doshi (approved):
lgtm

crates/common/chain/lean/src/lean_chain.rs
KolbyML: Good catch we should do your suggestion before merging, I would probably move

        db.latest_finalized_provider()
            .insert(genesis_state.latest_finalized.clone())
            .expect("Failed to insert latest finalized checkpoint");
        db.latest_justified_provider
        
        above both the insesrt block and state, to keep them together

syjn99: Resolved! Thanks!

PR #819, opened 2025-09-29
chore: improve logging & apply verbosity 

syjn99:
### What was wrong?

 

- Logging isn't clear and verbose. In Lean development, we mostly see the slot number for justification/finalization.
- Verbosity wasn't applied even though we all have it in subcommand.

### How was it fixed?

 

image

- Used structured logging with consistency. 
  - I didn't use structured manner for validator (`by Validator N`) for readability.
- Add `verbosity` as a common flag for all subcommands.

### To-Do

 
### DIFF
--- a/bin/ream/src/cli/mod.rs
+++ b/bin/ream/src/cli/mod.rs
@@ -90,6 +90,8 @@ mod tests {
     fn test_cli_lean_node_command() {
         let cli = Cli::parse_from([
             "program",
+            "--verbosity",
+            "5",
             "lean_node",
             "--network",
             "./assets/lean/sample_spec.yml",
@@ -97,9 +99,20 @@ mod tests {
             "./assets/lean/validator_registry.yml",
         ]);
 
+        assert_eq!(cli.verbosity.directive(), "trace");
+
         match cli.command {
             Commands::LeanNode(config) => {
-                // assert_eq!(config.network, "./assets/lean/sample_spec.yml");
+                assert_eq!(
+                    config.validator_registry_path.to_str().unwrap(),
+                    "./assets/lean/validator_registry.yml"
+                );
+
+                // Verify the network spec was loaded from the YAML file (sample_spec.yml)
+                assert_eq!(config.network.seconds_per_slot, 12);
+                assert_eq!(config.network.genesis_time, 0);
+                assert_eq!(config.network.justification_lookback_slots, 3);
+                assert_eq!(config.network.num_validators, 4);
             }
             _ => unreachable!("This test should only validate the lean node cli"),
         }
@@ -109,6 +122,8 @@ mod tests {
     fn test_cli_beacon_node_command() {
         let cli = Cli::parse_from([
             "program",
+            "--verbosity",
+            "2",
             "beacon_node",
             "--socket-address",
             "127.0.0.1",
@@ -117,6 +132,9 @@ mod tests {
             "--discovery-port",
             "9002",
         ]);
+
+        // Verify verbosity level (2 = warn)
+        assert_eq!(cli.verbosity.directive(), "warn");
 
         match cli.command {
             Commands::BeaconNode(config) => {
@@ -136,6 +154,8 @@ mod tests {
     fn test_cli_validator_node_command() {
         let cli = Cli::parse_from([
             "program",
+            "--verbosity",
+            "3",
             "validator_node",
             "--beacon-api-endpoint",
             "http://localhost:5052",
@@ -148,6 +168,9 @@ mod tests {
             "--password",
             "𝔱𝔢𝔰𝔱𝔭𝔞𝔰𝔰𝔴𝔬𝔯𝔡🔑", // Taken directly from EIP-2335's test keystores
         ]);
+
+        // Verify verbosity level (3 = info)
+        assert_eq!(cli.verbosity.directive(), "info");
 
         match cli.command {
             Commands::ValidatorNode(config) => {
@@ -165,6 +188,8 @@ mod tests {
     fn test_cli_account_manager_command() {
         let cli = Cli::parse_from([
             "program",
+            "--verbosity",
+            "6",
             "account_manager",
             "--lifetime",
             "30",
@@ -176,6 +201,9 @@ mod tests {
             "5",
         ]);
 
+        // Verify verbosity level (6+ maps to trace)
+        assert_eq!(cli.verbosity.directive(), "trace");
+
         match cli.command {
             Commands::AccountManager(config) => {
                 assert_eq!(config.lifetime, 30);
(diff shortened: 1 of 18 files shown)


bin/ream/src/cli/mod.rs
```diff
@@ -15,12 +16,16 @@ use ream_node::version::FULL_VERSION;
...
-    validator_node::ValidatorNodeConfig, voluntary_exit::VoluntaryExitConfig,
+    validator_node::ValidatorNodeConfig, verbosity::Verbosity, voluntary_exit::VoluntaryExitConfig,
 };
 
 #[derive(Debug, Parser)]
 #[command(author, version = FULL_VERSION, about, long_about = None)]
 pub struct Cli {
+    /// Verbosity level (1=error, 2=warn, 3=info, 4=debug, 5=trace)
+    #[arg(short, long, default_value = "3")]
+    pub verbosity: Verbosity,
```
KolbyML, line 31: isn't it possible to add a value constraint here, so only 1 to 5 can be submitted?

bin/ream/src/main.rs
```diff
@@ -75,17 +75,20 @@ pub const APP_NAME: &str = "ream";
...
+    let cli = Cli::parse();
+
+    // Set the default log level based on verbosity flag or RUST_LOG env var
     let rust_log = env::var(EnvFilter::DEFAULT_ENV).unwrap_or_default();
     let env_filter = match rust_log.is_empty() {
-        true => EnvFilter::builder().parse_lossy("info,actix_server=warn,discv5=error"),
+        true => EnvFilter::builder().parse_lossy(format!(
+            "{},actix_server=warn,discv5=error",
+            cli.verbosity.directive()
+        )),
```
KolbyML: ``actix_server=warn,discv5=error`` should be moved to the directive code


```suggestion
        true => EnvFilter::builder().parse_lossy(cli.verbosity.directive()),
```

bin/ream/src/cli/verbosity.rs
```diff
@@ -0,0 +1,28 @@
+#[derive(Debug, Clone, Copy)]
+pub struct Verbosity(pub u8);
```
KolbyML: I think it would make sense to use an enum here, as we only have 5 possible options

bin/ream/src/cli/verbosity.rs
```diff
@@ -0,0 +1,28 @@
...
+impl Verbosity {
+    pub fn directive(&self) -> &str {
+        match self.0 - 1 {
+            0 => "error",
+            1 => "warn",
+            2 => "info",
+            3 => "debug",
+            _ => "trace",
+        }
+    }
```
KolbyML, line 20: put ``,actix_server=warn,discv5=error`` somehwere here

crates/common/chain/lean/src/service.rs
```diff
@@ -125,11 +137,18 @@ impl LeanChainService {
...
-                                signed_vote.message.source,
-                                signed_vote.message.target
+                            );
+                            debug!(
+                                slot = signed_vote.message.slot,
+                                head = ?signed_vote.message.head,
+                                source = ?signed_vote.message.source,
+                                target = ?signed_vote.message.target,
+                                "Vote details",
                             );
```
KolbyML: Image

I don't like this style of redundent logging, this is done a few times in this PR

```rust
        if enabled!(Level::DEBUG) {
            debug!(
                "Successfully offered to {}/{} peers. Content key: {}. Declined: {:?}. Failed: {:?}",
                self.accepted ,
                self.total,
                self.content_key.to_hex(),
                self.declined,
                self.failed,
            );
        } else {
            info!(
                "Successfully offered to {}/{} peers. Content key: {}. Decline reasons: {}. Failed: {}.",
                self.accepted ,
                self.total,
                self.content_key.to_hex(),
                self.decline_reason_summary(),
                self.failed.len(),
            );
        }
```

do something like this, so that we can reduce the noise even further

KolbyML (changes requested):
left some feedback

crates/common/chain/lean/src/service.rs
syjn99: Oh I like this

bin/ream/src/cli/verbosity.rs
```diff
@@ -0,0 +1,42 @@
+// Fixed verbosity levels for specific crates to reduce log noise
+const ACTIX_SERVER_DIRECTIVE: &str = "actix_server=warn";
+const DISCV5_DIRECTIVE: &str = "discv5=error";
```
KolbyML: i don't think these need to be constants to be honest, they are only use in one spot

bin/ream/src/cli/verbosity.rs
```diff
@@ -0,0 +1,42 @@
...
+impl Verbosity {
+    pub fn directive(&self) -> String {
+        let directive = match self {
+            Verbosity::Error => "error",
+            Verbosity::Warn => "warn",
+            Verbosity::Info => "info",
+            Verbosity::Debug => "debug",
+            Verbosity::Trace => "trace",
+        };
+        format!("{directive},{ACTIX_SERVER_DIRECTIVE},{DISCV5_DIRECTIVE}")
```
KolbyML:
```suggestion
    pub fn directive(&self) -> String {
        match self {
            Verbosity::Error => "error,actix_server=warn,discv5=error",
            Verbosity::Warn => "warn,actix_server=warn,discv5=error",
            Verbosity::Info => "info,actix_server=warn,discv5=error",
            Verbosity::Debug => "debug",
            Verbosity::Trace => "trace",
        }.to_string()
```

we will want to fine tune this over time, but generally it would be good like this ^, using constants doesn't make sense in this case

bin/ream/src/cli/verbosity.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    Ok(match level {
+        1 => Verbosity::Error,
+        2 => Verbosity::Warn,
+        3 => Verbosity::Info,
+        4 => Verbosity::Debug,
+        5 => Verbosity::Trace,
+        _ => unreachable!(),
+    })
```
KolbyML:
```suggestion
    match level {
        1 => Verbosity::Error,
        2 => Verbosity::Warn,
        3 => Verbosity::Info,
        4 => Verbosity::Debug,
        5 => Verbosity::Trace,
        _ => Err(format!("verbosity must be between 1 and 5, got {level}")),
    }.map(Ok)
```

we can avoid the unreachable and simplify the logic, and unreachable should always contain a "message", but we don't need it

If the map doesn't work just wraap them in Ok manually

KolbyML (approved):
:shipit: looks great

PR #823, opened 2025-10-01
feat(lean): remove `dependencies` in `LeanChainService` & implement `on_attestation` of fork choice specification 

### LINKED ISSUE #787: Double-check implementation with current `leanSpec`.

syjn99:
### What was wrong?

 

Fixes #787.

Our codebase is mostly came from 3sf-mini, which uses `dependencies` as a queue. However, this implementation is not intuitive as well as complex for real-world devnets.

### How was it fixed?

 

This PR mainly implements `on_attestation` ([Link](https://github.com/leanEthereum/leanSpec/blob/ee16b19825a1f358b00a6fc2d7847be549daa03b/docs/client/forkchoice.md?plain=1#L279-L312)) function. I did some optimization regarding database transaction using `batch_insert` (thanks to @varun-doshi), so the current code quite deviates from Python spec.

Also, `handle_process_{block|vote}` are now just a wrapper for future p2p validation. By implementing `on_block` and `on_attestation_*` under `LeanChain`, the function becomes much simpler because it removes tackling with synchronization stuffs (e.g., `lock()`...).

### To-Do

 
### DIFF
--- a/crates/common/chain/lean/src/lean_chain.rs
+++ b/crates/common/chain/lean/src/lean_chain.rs
@@ -339,4 +339,74 @@ impl LeanChain {
             source,
         })
     }
+
+    /// Validates and processes a new attestation (a signed vote), updating the store's
+    /// latest votes.
+    ///
+    /// See lean specification:
+    /// <https://github.com/leanEthereum/leanSpec/blob/ee16b19825a1f358b00a6fc2d7847be549daa03b/docs/client/forkchoice.md?plain=1#L279-L312>
+    pub async fn on_attestation(
+        &mut self,
+        signed_vote: SignedVote,
+        is_from_block: bool,
+    ) -> anyhow::Result<()> {
+        let validator_id = signed_vote.validator_id;
+
+        if is_from_block {
+            let latest_known_votes_provider = {
+                let db = self.store.lock().await;
+                db.latest_known_votes_provider()
+            };
+
+            // The below logic mirrors the Python spec:
+            //
+            // ```python
+            // # update latest known votes if this is latest
+            // latest_vote = store.latest_known_votes.get(validator_id)
+            // if latest_vote is None or latest_vote.slot < vote.slot:
+            //     store.latest_known_votes[validator_id] = vote
+            // ```
+            if latest_known_votes_provider
+                .get(validator_id)?
+                .map_or(true, |latest_vote| {
+                    latest_vote.message.slot < signed_vote.message.slot
+                })
+            {
+                latest_known_votes_provider.insert(validator_id, signed_vote.clone())?;
+            }
+
+            // The below logic mirrors the Python spec:
+            // ```python
+            // # clear from new votes if this is latest
+            // latest_vote = store.latest_new_votes.get(validator_id)
+            // if latest_vote is not None and latest_vote.slot < vote.slot:
+            //     del store.latest_new_votes[validator_id]
+            // ```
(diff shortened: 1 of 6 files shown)


crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -339,4 +339,125 @@ impl LeanChain {
...
+                // Filter for batch insertion.
+                //
+                // This logic mirrors the Python spec:
+                //
+                // ```python
+                // # update latest known votes if this is latest
+                // latest_vote = store.latest_known_votes.get(validator_id)
+                // if latest_vote is None or latest_vote.slot < vote.slot:
+                //     store.latest_known_votes[validator_id] = vote
+                // ```
```
KolbyML: at what point do these comments make the code harder to read, as time goes on our code should* deviate more and more from the spec in terms of how it is written, these large multiline comments end up being a distraction over time no?, a comment at the top of the file or function, should be adequate, if not overkill in most cases

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -339,4 +339,125 @@ impl LeanChain {
...
+                    signed_block.message.parent_root
+                )
+            })?;
+        state.state_transition(&signed_block, true, true)?;
+
+        lean_block_provider.insert(block_hash, signed_block.clone())?;
+        latest_justified_provider.insert(state.latest_justified.clone())?;
+        lean_state_provider.insert(block_hash, state)?;
+        self.on_attestation_from_block(signed_block.message.body.attestations)
+            .await?;
```
KolbyML:
```suggestion
        let attestations = signed_block.message.body.attestations;
        lean_block_provider.insert(block_hash, signed_block)?;
        latest_justified_provider.insert(state.latest_justified.clone())?;
        lean_state_provider.insert(block_hash, state)?;
        self.on_attestation_from_block(attestations)
            .await?;
```

can't we avoid cloning the whole block here?

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -339,4 +339,125 @@ impl LeanChain {
...
+    pub fn on_attestation_from_gossip(&mut self, signed_vote: SignedVote) {
+        let validator_id = signed_vote.validator_id;
+
+        // The below logic mirrors the Python spec:
+        // ```
+        // # update latest new votes if this is the latest
+        // latest_vote = store.latest_new_votes.get(validator_id)
+        // if latest_vote is None or latest_vote.slot < vote.slot:
+        //     store.latest_new_votes[validator_id] = vote
+        // ```
```
KolbyML: i think this comment isn't needed

crates/common/chain/lean/src/lean_chain.rs
syjn99: Yes, makes sense. I'll remove bunch of python code blocks then. Thanks!

crates/common/chain/lean/src/lean_chain.rs
syjn99: https://github.com/ReamLabs/ream/pull/823/commits/780408a4ed526b0808691da54204e83336300374

Fixed to only clone `attesations`!

KolbyML (approved):
:shipit: looks good

PR #825, opened 2025-10-03
chore(lean): add alias (`node-key`) for `private-key-path`

syjn99:
### What was wrong?

 

> so as per our discussions 3 args are standard:
> 
> 1. —node-id (zeam_0, ream_0...)
> 2. —node_key : path to private key for enr/libp2p
> 3. —data-dir: path to data directory

We need an alias for `private-key-path` flag for better interop

### How was it fixed?

 

Use `alias` in `clap`

### To-Do

 

varun-doshi (approved):
lgtm

PR #828, opened 2025-10-05
fix(lean/p2p): increase `max_established_per_peer` into 2

syjn99:
### What was wrong?

 

When a node tries to establish a duplicated connection, the existing connection got disconnected, leading some p2p issues in pq devnet0. We're using same `nodes.yaml`, which means there should be at least two duplicated connection here.

### How was it fixed?

 

Increase `max_established_per_peer` value into 2.

### To-Do

 

KolbyML (approved):
:shipit: looks good

Issue #816, opened 2025-09-29
Add tests about state in `leanSpec`


syjn99:
### Describe the feature

https://github.com/leanEthereum/leanSpec/blob/main/tests/lean_spec/subspecs/containers/test_state.py

We now have tests written in Python. Adding those tests in our codebase will be promising as well.

shariqnaiyer:
Hey guys I've already been working on this. Hadn't made an issue yet. I'll be making a PR in the next couple days. 

Can I take this issue?

PR #830, opened 2025-10-05
feat: add specTest for state container

shariqnaiyer:
### What was wrong?

Fixes: #816 

### How was it fixed?

All tests were implemented based on [LeanSpec](https://github.com/leanEthereum/leanSpec/blob/main/tests/lean_spec/subspecs/containers/test_state.py)

I was going to do this in a new package under tests but I saw that O had done a lot of these in lean/src/state.rs already so I believe that would be a good place to put the rest as they are related to functions and structs defined in that file.

### To-Do

 

crates/common/consensus/lean/src/state.rs
```diff
@@ -652,4 +652,440 @@ mod test {
...
+        // History should include the parent's root at index 0
+        assert_eq!(genesis_state.historical_block_hashes.len(), 1);
+        assert_eq!(
+            genesis_state.historical_block_hashes[0],
+            genesis_header_root
+        );
+
+        // Slot 0 should be marked justified
+        assert_eq!(genesis_state.justified_slots.len(), 1);
+        assert!(genesis_state.justified_slots[0]);
```
Kayden-ML: Seems like there is a issue indexing like this, I believe this would work.

```suggestion
        assert!(genesis_state.justified_slots.get(0).unwrap_or(false));
```

crates/common/consensus/lean/src/state.rs
```diff
@@ -652,4 +652,440 @@ mod test {
...
+            };
+            votes.push(vote).unwrap();
+        }
+
+        // Process attestations
+        state.process_attestations(&votes).unwrap();
+
+        // The target (slot 4) should now be justified
+        assert_eq!(state.latest_justified, checkpoint4);
+        assert!(state.justified_slots[4]);
```
Kayden-ML: Similar thing here

PR #832, opened 2025-10-05
fix(lean): consensus issues (use `Bitlist` for `justification_slots`, etc.)

syjn99:
### What was wrong?

 

Fixing multiple issues found in interop

### How was it fixed?

 

1. genesis generation logic
2. use `Bitlist` for `justification_slots`
3. correctly calculate slot (not to start with `1`)

### Reproducing

Use https://github.com/ReamLabs/local-pq-devnet/tree/zeam-integration

### To-Do

 
### DIFF
--- a/crates/common/consensus/lean/src/state.rs
+++ b/crates/common/consensus/lean/src/state.rs
@@ -51,7 +51,10 @@ impl LeanState {
                 genesis_time,
             },
             slot: 0,
-            latest_block_header: BlockHeader::default(),
+            latest_block_header: BlockHeader {
+                body_root: BlockBody::default().tree_hash_root(),
+                ..BlockHeader::default()
+            },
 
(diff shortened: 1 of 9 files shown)


crates/common/consensus/lean/src/state.rs
```diff
@@ -115,7 +116,9 @@ impl LeanState {
             // If the length is incorrect, the constructed bitlist will be corrupt.
             ensure!(
                 justifications_for_root.len() == self.config.num_validators as usize,
-                "Justifications length does not match validator registry limit"
+                "Justifications length ({}) does not match validator registry limit ({})",
```
unnawut:
```suggestion
                "Justifications length ({}) does not match validators length ({})",
```

unnawut (approved):
Lgtm with just 1 nit

PR #843, opened 2025-10-15
feat(lean): add `Validator` Container & add `validators` field in `LeanState`

### LINKED ISSUE #841: Add `Validator` Container for Lean
### Describe the feature

```python
"""Validator container for the Lean Ethereum consensus specification."""

from lean_spec.types import Bytes52, Container


class Validator(Container):
    """Represents a validator's static metadata."""

    pubkey: Bytes52
    """XMSS one-time signature public key."""
```

### Additional context

_No response_

syjn99:
### What was wrong?

 

Fix #841 

### How was it fixed?

 

I tried a couple of way to make our `PublicKey` compatible to all the traits we need (e.g., `Debug`, `Clone`, `ssz::Encode`, `tree_hash::TreeHash`...) but it was hard. (`SignatureScheme::PublicKey` only implements `serde::{Serialize, Deserialize}`.)

So treat `PublicKey` as a raw bytes and convert it to the one at `hash-sig` if we need.

### To-Do

 
### DIFF
--- a/crates/crypto/post_quantum/src/hashsig/public_key.rs
+++ b/crates/crypto/post_quantum/src/hashsig/public_key.rs
@@ -24,19 +24,20 @@ const BINCODE_CONFIG: bincode::config::Configuration<LittleEndian, Fixint, NoLim
 
 /// Wrapper around the `GeneralizedXMSSPublicKey` from the hashsig crate.
 ///
-/// With current signature parameters, the serialized public key is 52 bytes
-/// Public key is composed of:
+/// With current signature parameters, the serialized public key is 52 bytes:
+/// - Public key consists of:
 /// - the root of the merkle tree (an array of 8 finite field elements),
 /// - a parameter for the tweakable hash (an array of 5 finite field elements).
-/// Each KoalaBear finite field element is 32-bit, so the total size is:
-/// 32 bits * (8 + 5) = 416 bits = 52 bytes
+/// - One KoalaBear finite field element is 32 bits (4 bytes).
+/// - The total size is 52 bytes.
 ///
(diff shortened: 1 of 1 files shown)


crates/crypto/post_quantum/src/hashsig/public_key.rs
```diff
@@ -1,17 +1,90 @@
...
+        }
+    }
+
+    /// Convert back to the original `GeneralizedXMSSPublicKey` type from the hashsig crate.
+    /// To use this public key for signature verification.
+    pub fn to_hash_sig_public_key(&self) -> anyhow::Result<HashSigPublicKey> {
+        bincode::serde::decode_from_slice(&self.inner, BINCODE_CONFIG)
+            .map(|(value, _)| value)
+            .map_err(|err| anyhow::anyhow!("Failed to decode public key: {}", err))
+    }
```
KolbyML, line 67: shouldn't we use the From/TryFrom traits to implement these?

syjn99: Good point, that was one that I was suffering from.

```
/// NOTE: [SignatureScheme::PublicKey] is a Rust trait that only implements [serde::Serialize] and
/// [serde::Deserialize]. So it's impossible to implement [From] or [Into] traits for it.
```

I kept facing compiler issue for implementing `From` trait of `SignatureScheme::PublicKey` which is also a trait. Any advices to make this clearer?

KolbyML: Ok I think we can just merge this for now then

PR #861, opened 2025-10-17
fix: getrandom target error in risc0 and sp1 workflow

shariqnaiyer:
### What was wrong?

risc0 and sp1 use unsupported targets for getrandom and cause an error in our worflow

### How was it fixed?

https://github.com/risc0/risc0/issues/634
https://docs.rs/getrandom/latest/getrandom/#opt-in-backends

The workaround as suggested by the above issue is to use a rust flag to tell getrandom a that custom backend will be used.

```
Opt-in backends can be enabled using the getrandom_backend configuration flag. The flag can be set either by specifying the rustflags field in [.cargo/config.toml](https://doc.rust-lang.org/cargo/reference/config.html):

# It's recommended to set the flag on a per-target basis:
[target.wasm32-unknown-unknown]
rustflags = ['--cfg', 'getrandom_backend="wasm_js"']
```

We can put in the config.toml so devs don't have to add it with environment variables.

### To-Do

 
### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,7 +1,2 @@
-# Custom getrandom backend for RISC Zero zkVM target
-[target.riscv32im-risc0-zkvm-elf]
-rustflags = ['--cfg', 'getrandom_backend="custom"']
-
-# Custom getrandom backend for SP1 zkVM target
-[target.riscv32im-succinct-zkvm-elf]
+[target.'cfg(target_os = "zkvm")']
 rustflags = ['--cfg', 'getrandom_backend="custom"']


syjn99 (commented):
Great! We might use `.cargo/config.toml` for applying this at all zkVM target compilation. This `config.toml` works for me:

```toml
[target.'cfg(target_os = "zkvm")']
rustflags = ["--cfg", "getrandom_backend=\"custom\""]
```

.cargo/config.toml
```diff
@@ -0,0 +1,7 @@
+# Custom getrandom backend for RISC Zero zkVM target
+[target.riscv32im-risc0-zkvm-elf]
+rustflags = ['--cfg', 'getrandom_backend="custom"']
+
+# Custom getrandom backend for SP1 zkVM target
+[target.riscv32im-succinct-zkvm-elf]
+rustflags = ['--cfg', 'getrandom_backend="custom"']
```
KolbyML: you should do the suggestion by @syjn99 we want to support all zkVM's, not target them individually

KolbyML:
@shariqnaiyer before we merge this, we should get the issues opened on hashsig/plonky3/ream first

shariqnaiyer:
The following issues were made to handle this upstream:

Ream: https://github.com/ReamLabs/ream/issues/862
Hash-sig: https://github.com/b-wagn/hash-sig/issues/90
Plonky3: https://github.com/Plonky3/Plonky3/issues/1089

PR #869, opened 2025-10-21
feat: add shadow network simulator support

turuslan:
### What was wrong?
Couldn't run ream in [shadow network simulator](https://github.com/shadow/shadow).
Shadow doesn't implement some `setsockopt` options.

### How was it fixed?
Patch `quinn-udp` to avoid unsupported `setsockopt` by using [fallback.rs](https://github.com/quinn-rs/quinn/blob/main/quinn-udp/src/fallback.rs#L19-L25) instead of [unix.rs](https://github.com/quinn-rs/quinn/blob/fc5e219fd9c9418c874f3db4e6d26d9cc32cea67/quinn-udp/src/unix.rs#L86-L194).

`quinn-udp` is transitive dependency
`ream -> libp2p -> libp2p-quic -> quinn -> quinn-udp`
so `feature = "shadow"` was added to every crate in that chain.

### To-Do

 
### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,9 +99,9 @@ itertools = "0.14"
 jsonwebtoken = "9.3.1"
 kzg = { git = "https://github.com/grandinetech/rust-kzg" }
 lazy_static = "1.5.0"
-libp2p = { version = "0.55", default-features = false, features = ["identify", "yamux", "noise", "dns", "serde", "tcp", "tokio", "plaintext", "secp256k1", "macros", "ecdsa", "metrics", "quic", "upnp", "gossipsub", "ping"] }
+libp2p = { version = "0.55", default-features = false, features = ["identify", "yamux", "noise", "dns", "serde", "tcp", "tokio", "plaintext", "secp256k1", "macros", "ecdsa", "metrics", "quic", "upnp", "gossipsub", "ping", "shadow"] }
 libp2p-identity = "0.2.12"
-libp2p-mplex = "0.43.1"
+libp2p-mplex = "0.43.0"
 lru = "0.16.2"
 parking_lot = "0.12.5"
 prometheus_exporter = { git = "https://github.com/AlexanderThaller/prometheus_exporter", rev = "c49efe614486f998b20eb410ae0caf3e904cf540" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -168,3 +168,10 @@ ream-validator-lean = { path = "crates/common/validator/lean" }
 
 [patch.crates-io]
 ethereum_hashing = { git = "https://github.com/ReamLabs/ethereum_hashing.git" }
+
+# https://github.com/shadow/shadow
+libp2p = { git = "https://github.com/qdrvm/rust-libp2p.git", branch = "feature/shadow-0.55.0" }
```
syjn99, line 173: I think we can figure out a better way, not patching the full `libp2p` with forked one. Also it seems we only need to patch `quinn-udp` but it's a transitive dep so I should check.

@KolbyML Is it possible to patch **only** when a specific rust flag is enabled? For example in this case, `--feature shadow`.

Cargo.toml
```diff
@@ -99,9 +99,9 @@ itertools = "0.14"
 jsonwebtoken = "9.3.1"
 kzg = { git = "https://github.com/grandinetech/rust-kzg" }
 lazy_static = "1.5.0"
-libp2p = { version = "0.55", default-features = false, features = ["identify", "yamux", "noise", "dns", "serde", "tcp", "tokio", "plaintext", "secp256k1", "macros", "ecdsa", "metrics", "quic", "upnp", "gossipsub", "ping"] }
+libp2p = { version = "0.55", default-features = false, features = ["identify", "yamux", "noise", "dns", "serde", "tcp", "tokio", "plaintext", "secp256k1", "macros", "ecdsa", "metrics", "quic", "upnp", "gossipsub", "ping", "shadow"] }
 libp2p-identity = "0.2.12"
-libp2p-mplex = "0.43.1"
+libp2p-mplex = "0.43.0"
```
KolbyML, line 104: I don't think we should downgrade our packages

KolbyML (changes requested):
I am not a fan of adding unmaintained dependencies to our Cargo.toml

Regardless if we were to do something like this we would instead

```diff
diff --git a/.cargo/config-shadow.toml b/.cargo/config-shadow.toml
new file mode 100644
index 0000000..cb8b6ca
--- /dev/null
+++ b/.cargo/config-shadow.toml
@@ -0,0 +1,7 @@
+[patch.crates-io]
+# https://github.com/shadow/shadow
+libp2p = { git = "https://github.com/qdrvm/rust-libp2p.git", branch = "feature/shadow-0.55.0" }
+libp2p-mplex = { git = "https://github.com/qdrvm/rust-libp2p.git", branch = "feature/shadow-0.55.0" }
+libp2p-quic = { git = "https://github.com/qdrvm/rust-libp2p.git", branch = "feature/shadow-0.55.0" }
+quinn = { git = "https://github.com/qdrvm/quinn.git", branch = "feature/shadow-0.5.11" }
+quinn-udp = { git = "https://github.com/qdrvm/quinn.git", branch = "feature/shadow-0.5.11" }
```
and build Ream with ``cargo build --config .cargo/config-shadow.toml`` to include these patches.

But since we will be upgrading to libp2p version 0.56 shortly, I don't think we should merge this.

Even then once version 0.57 came out it would break again. I don't want to maintain this so it would probably be removed later.

So maybe it would be best to keep this on your local testing branch of ream?

PR #873, opened 2025-10-27
feat: use state validators and rm in config

### LINKED ISSUE #872: Remove num_validators from config
### Describe the feature

Ref: https://github.com/leanEthereum/leanSpec/pull/84

### Additional context

_No response_

varun-doshi:
### What was wrong?

Fixes #872 
### How was it fixed?

 
Use `validator.len()` from lean state and remove `no_of_validators` from config

### DIFF
--- a/crates/common/consensus/lean/src/validator.rs
+++ b/crates/common/consensus/lean/src/validator.rs
@@ -14,7 +14,7 @@ impl Validator {
     pub fn generate_default_validators(number_of_validators: usize) -> Vec<Validator> {
         (0..number_of_validators)
             .map(|_| Validator {
-                public_key: PublicKey::from_bytes(&[0_u8; 52]),
+                public_key: PublicKey::from(&[0_u8; 52][..]),
             })
             .collect()
     }
--- a/crates/crypto/post_quantum/src/hashsig/public_key.rs
+++ b/crates/crypto/post_quantum/src/hashsig/public_key.rs
@@ -43,13 +43,15 @@ pub struct PublicKey {
     inner: FixedVector<u8, U52>,
 }
 
-impl PublicKey {
-    pub fn from_bytes(bytes: &[u8]) -> Self {
+impl From<&[u8]> for PublicKey {
+    fn from(value: &[u8]) -> Self {
         Self {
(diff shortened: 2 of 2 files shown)


crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -48,7 +48,7 @@ pub struct LeanChain {
 impl LeanChain {
     pub fn new(genesis_block: SignedBlock, genesis_state: LeanState, db: LeanDB) -> LeanChain {
         let genesis_block_hash = genesis_block.message.tree_hash_root();
-        let no_of_validators = genesis_state.config.num_validators;
+        let no_of_validators = genesis_state.validators.len();
```
shariqnaiyer: just a thought, views on calling this variable `number_of_validators`?

crates/common/consensus/lean/src/validator.rs
```diff
@@ -9,3 +9,13 @@ pub struct Validator {
...
+
+impl Validator {
+    pub fn new_from_keys(number_of_validators: usize) -> Vec<Validator> {
+        (0..number_of_validators)
+            .map(|_| Validator {
+                public_key: PublicKey::from_bytes(&[0_u8; 52]),
+            })
+            .collect()
+    }
+}
```
shariqnaiyer: I think `new_from_keys` is a misleading name from this function as usually `from_x (x: type x) -> y`

If this was taking in public keys that would make more sense but it seems we could name it with more context here.

crates/crypto/post_quantum/src/hashsig/public_key.rs
```diff
@@ -44,6 +44,12 @@ pub struct PublicKey {
 }
 
 impl PublicKey {
+    pub fn from_bytes(bytes: &[u8]) -> Self {
+        Self {
+            inner: FixedVector::from(bytes.to_vec()),
+        }
+    }
```
KolbyML: to_bytes is fine, but shouldn't this be done using the from trait?

KolbyML (commented):
most of the PR looks fine just have a question

Issue #846, opened 2025-10-15
Add functionality for storing/recovering `hash-sig` Keypair for Devnet-1


syjn99:
### Describe the feature

Although https://github.com/ReamLabs/ream/issues/792 suggests the endgame (e.g., encryption), I would say we only need a very limited set of functionality for keystore in Devnet-1.

My proposal:
- We want to provide a file that contains "encoded" keypair (Public key + Secret Key).
- Also, as one validator might contain multiple validators, we should b able to provide multiple file paths that contain encoded keypair with a validator index.

Note that `SecretKey` in `hash-sig` doesn't have an API like `to_public_key()`.

shariqnaiyer:
@syjn99 I'd love to work on this.

PR #875, opened 2025-10-31
feat: Add functionality for storing/recovering hash-sig Keypair for D…

Kayden-ML:
### What was wrong?

Fixes: https://github.com/ReamLabs/ream/issues/846

### How was it fixed?

Added functionality for storing/recovering hash-sig Keypair

### DIFF
--- a/crates/crypto/keystore/Cargo.toml
+++ b/crates/crypto/keystore/Cargo.toml
@@ -17,7 +17,6 @@ hashsig.workspace = true
 rand.workspace = true
 serde.workspace = true
 serde_json.workspace = true
-serde_yaml.workspace = true
 sha2.workspace = true
 ssz_types.workspace = true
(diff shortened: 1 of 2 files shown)


crates/crypto/keystore/Cargo.toml
```diff
@@ -13,15 +13,18 @@ version.workspace = true
 aes.workspace = true
 alloy-primitives.workspace = true
 anyhow.workspace = true
+hashsig.workspace = true
 rand.workspace = true
 serde.workspace = true
 serde_json.workspace = true
+serde_yaml.workspace = true
```
shariqnaiyer: This is not being used in this crate right now. I think it is save to remove.

Kayden-ML: ya i had to update udeps is all

shariqnaiyer:
looks good 👍

Issue #847, opened 2025-10-15
Update newly added STF logic for Devnet-1


syjn99:
### Describe the feature

https://github.com/leanEthereum/leanSpec/pull/67

This PR will add new STF and fork choice logic. Currently blocked because #840 will add necessary containers for this issue. (And the spec isn't fixed yet.)

Kayden-ML:
Picking this back up 👍

PR #877, opened 2025-11-01
feat(devnet1): update containers for block

### LINKED ISSUE #856: Update containers for SignedBlock and Block
### Describe the feature

New spec PR has changes to the `SignedBlock` and `Block` containers. 

### Additional context

_No response_

varun-doshi:
### What was wrong?

Fixes #856 
### How was it fixed?

 

Updates `Block` `SignedBlockWithAttestation` `BlockWithAttestation` types and apply necessary changes throughout codebase 
Ref: https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/containers/block/block.py

`propose_block` now returns `<(Block, Vec<FixedBytes<4000>>)`

### DIFF
--- a/crates/common/chain/lean/src/lean_chain.rs
+++ b/crates/common/chain/lean/src/lean_chain.rs
@@ -285,7 +285,6 @@ impl LeanChain {
             slot,
             proposer_index: slot % lean_network_spec().num_validators,
             parent_root: head,
-            // Diverged from Python implementation: Using `B256::ZERO` instead of `None`)
             state_root: B256::ZERO,
             body: BlockBody::default(),
         };
@@ -306,12 +305,18 @@ impl LeanChain {
             let mut new_attestations_to_add = Vec::new();
             let mut new_signatures_to_add = Vec::new();
 
-            for (_, attestation) in latest_known_attestation_provider.get_all_attestations()? {
-                if attestation.message.source() == state.latest_justified
-                    && !new_block.body.attestations.contains(&attestation.message)
+            for signed_attestation in latest_known_attestation_provider
+                .get_all_attestations()?
+                .values()
+            {
+                if signed_attestation.message.source() == state.latest_justified
+                    && !new_block
+                        .body
+                        .attestations
+                        .contains(&signed_attestation.message)
                 {
-                    new_attestations_to_add.push(attestation.message);
-                    new_signatures_to_add.push(attestation.signature);
+                    new_attestations_to_add.push(signed_attestation.message.clone());
+                    new_signatures_to_add.push(signed_attestation.signature);
                 }
             }
 
@@ -381,9 +386,10 @@ impl LeanChain {
     /// <https://github.com/leanEthereum/leanSpec/blob/ee16b19825a1f358b00a6fc2d7847be549daa03b/docs/client/forkchoice.md?plain=1#L314-L342>
     pub async fn on_block(
         &mut self,
-        signed_block: SignedBlockWithAttestation,
+        signed_block_with_attestation: SignedBlockWithAttestation,
     ) -> anyhow::Result<()> {
-        let block_hash = signed_block.message.block.tree_hash_root();
+        let block = &signed_block_with_attestation.message.block;
+        let block_hash = signed_block_with_attestation.message.block.tree_hash_root();
 
         let (lean_block_provider, latest_justified_provider, lean_state_provider) = {
             let db = self.store.lock().await;
@@ -399,26 +405,25 @@ impl LeanChain {
             return Ok(());
         }
 
-        let mut state = lean_state_provider
-            .get(signed_block.message.block.parent_root)?
-            .ok_or_else(|| {
-                anyhow!(
-                    "Parent state not found for block: {block_hash}, parent: {}",
-                    signed_block.message.block.parent_root
-                )
-            })?;
+        let mut state = lean_state_provider.get(block.parent_root)?.ok_or_else(|| {
+            anyhow!(
+                "Parent state not found for block: {block_hash}, parent: {}",
+                block.parent_root
+            )
+        })?;
 
         // TODO: Add signature validation once spec is complete.
-        state.state_transition(&signed_block.message.block, true, true)?;
+        // Tracking issue: https://github.com/ReamLabs/ream/issues/881
+        state.state_transition(block, true, true)?;
 
         let mut signed_attestations = vec![];
-        for attestation in &signed_block.message.block.body.attestations {
+        for attestation in &block.body.attestations {
             signed_attestations.push(SignedAttestation {
                 message: attestation.clone(),
                 signature: FixedBytes::<4000>::default(),
             });
         }
-        lean_block_provider.insert(block_hash, signed_block)?;
+        lean_block_provider.insert(block_hash, signed_block_with_attestation)?;
         latest_justified_provider.insert(state.latest_justified)?;
         lean_state_provider.insert(block_hash, state)?;
         self.on_attestation_from_block(signed_attestations).await?;
--- a/crates/common/chain/lean/src/messages.rs
+++ b/crates/common/chain/lean/src/messages.rs
@@ -34,7 +34,7 @@ pub enum LeanChainServiceMessage {
         sender: oneshot::Sender<AttestationData>,
     },
     ProcessBlock {
-        signed_block: Box<SignedBlockWithAttestation>,
+        signed_block_with_attestation: Box<SignedBlockWithAttestation>,
         is_trusted: bool,
         need_gossip: bool,
     },
--- a/crates/common/chain/lean/src/service.rs
+++ b/crates/common/chain/lean/src/service.rs
@@ -106,31 +106,31 @@ impl LeanChainService {
                                 error!("Failed to handle build attestation data message: {err:?}");
                             }
                         }
-                        LeanChainServiceMessage::ProcessBlock { signed_block, is_trusted, need_gossip } => {
+                        LeanChainServiceMessage::ProcessBlock { signed_block_with_attestation, is_trusted, need_gossip } => {
                             if enabled!(Level::DEBUG) {
                                 debug!(
-                                    slot = signed_block.message.block.slot,
-                                    block_root = ?signed_block.message.block.tree_hash_root(),
-                                    parent_root = ?signed_block.message.block.parent_root,
-                                    state_root = ?signed_block.message.block.state_root,
-                                    attestations_length = signed_block.message.block.body.attestations.len(),
+                                    slot = signed_block_with_attestation.message.block.slot,
+                                    block_root = ?signed_block_with_attestation.message.block.tree_hash_root(),
+                                    parent_root = ?signed_block_with_attestation.message.block.parent_root,
+                                    state_root = ?signed_block_with_attestation.message.block.state_root,
+                                    attestations_length = signed_block_with_attestation.message.block.body.attestations.len(),
                                     "Processing block built by Validator {}",
-                                    signed_block.message.block.proposer_index,
+                                    signed_block_with_attestation.message.block.proposer_index,
                                 );
                             } else {
                                 info!(
-                                    slot = signed_block.message.block.slot,
-                                    block_root = ?signed_block.message.block.tree_hash_root(),
+                                    slot = signed_block_with_attestation.message.block.slot,
+                                    block_root = ?signed_block_with_attestation.message.block.tree_hash_root(),
                                     "Processing block built by Validator {}",
-                                    signed_block.message.block.proposer_index,
+                                    signed_block_with_attestation.message.block.proposer_index,
                                 );
                             }
 
-                            if let Err(err) = self.handle_process_block(*signed_block.clone(), is_trusted).await {
+                            if let Err(err) = self.handle_process_block(*signed_block_with_attestation.clone(), is_trusted).await {
                                 warn!("Failed to handle process block message: {err:?}");
                             }
 
-                            if need_gossip && let Err(err) = self.outbound_gossip.send(LeanP2PRequest::GossipBlock(signed_block)) {
+                            if need_gossip && let Err(err) = self.outbound_gossip.send(LeanP2PRequest::GossipBlock(signed_block_with_attestation)) {
                                 warn!("Failed to send item to outbound gossip channel: {err:?}");
                             }
                         }
@@ -205,7 +205,7 @@ impl LeanChainService {
 
     async fn handle_process_block(
         &mut self,
-        signed_block: SignedBlockWithAttestation,
+        signed_block_with_attestation: SignedBlockWithAttestation,
         is_trusted: bool,
     ) -> anyhow::Result<()> {
         if !is_trusted {
@@ -215,7 +215,7 @@ impl LeanChainService {
         self.lean_chain
             .write()
             .await
-            .on_block(signed_block.clone())
+            .on_block(signed_block_with_attestation.clone())
             .await?;
 
         Ok(())
--- a/crates/common/consensus/lean/src/attestation.rs
+++ b/crates/common/consensus/lean/src/attestation.rs
@@ -7,9 +7,7 @@ use tree_hash_derive::TreeHash;
 use crate::checkpoint::Checkpoint;
 
 /// Attestation content describing the validator's observed chain view.
-#[derive(
-    Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash,
-)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
 pub struct AttestationData {
     pub slot: u64,
     pub head: Checkpoint,
@@ -18,9 +16,7 @@ pub struct AttestationData {
 }
 
 /// Validator specific attestation wrapping shared attestation data.
-#[derive(
-    Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash,
-)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
 pub struct Attestation {
     pub validator_id: u64,
     pub data: AttestationData,
--- a/crates/common/consensus/lean/src/block.rs
+++ b/crates/common/consensus/lean/src/block.rs
@@ -36,9 +36,6 @@ pub struct Block {
 }
 
 /// Represents a block header in the Lean chain.
-///
-/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/5da200c13f5eeda0b4139b1d55970d75c011d4b2/src/lean_spec/subspecs/containers/block/block.py#L36)
-/// for detailed protocol information.
 #[derive(
     Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash,
 )]
@@ -63,9 +60,6 @@ impl From<Block> for BlockHeader {
 }
 
 /// Represents the body of a block in the Lean chain.
-///
-/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/5da200c13f5eeda0b4139b1d55970d75c011d4b2/src/lean_spec/subspecs/containers/block/block.py#L20)
-/// for detailed protocol information.
 #[derive(
     Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash,
 )]
--- a/crates/common/validator/lean/src/service.rs
+++ b/crates/common/validator/lean/src/service.rs
@@ -2,8 +2,9 @@ use alloy_primitives::FixedBytes;
 use anyhow::{Context, anyhow};
 use ream_chain_lean::{clock::create_lean_clock_interval, messages::LeanChainServiceMessage};
 use ream_consensus_lean::{
-    attestation::{Attestation, SignedAttestation},
+    attestation::{Attestation, AttestationData, SignedAttestation},
     block::{BlockWithAttestation, SignedBlockWithAttestation},
+    checkpoint::Checkpoint,
 };
 use ream_network_spec::networks::lean_network_spec;
 use ssz_types::VariableList;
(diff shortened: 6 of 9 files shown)


crates/common/consensus/lean/src/block.rs
```diff
@@ -5,21 +5,30 @@ use ssz_types::{VariableList, typenum::U4096};
...
+    pub signature: VariableList<FixedBytes<4000>, U4096>,
+}
+
+/// Bundle containing a block and the proposer's attestation.
+///
+/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/5da200c13f5eeda0b4139b1d55970d75c011d4b2/src/lean_spec/subspecs/containers/block/block.py#L83)
+#[derive(Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode)]
+pub struct BlockWithAttestation {
+    pub block: Block,
+    pub proposer_attestation: Attestation,
```
KolbyML: these shouldn't implement default

crates/common/chain/lean/src/service.rs
```diff
@@ -173,7 +173,7 @@ impl LeanChainService {
         slot: u64,
         response: oneshot::Sender<Block>,
     ) -> anyhow::Result<()> {
-        let new_block = self.lean_chain.write().await.propose_block(slot).await?;
+        let (new_block, _) = self.lean_chain.write().await.propose_block(slot).await?;
```
KolbyML, line 176: if we are not using the signature anymore, why is propose_block returning it?

crates/common/consensus/lean/src/attestation.rs
```diff
@@ -45,7 +49,9 @@ impl Attestation {
 }
 
 /// Validator attestation bundled with its signature.
-#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
+#[derive(
+    Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash,
+)]
 pub struct SignedAttestation {
```
KolbyML, line 49: this shouldn't implement default

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -378,20 +391,29 @@ impl LeanChain {
...
                 anyhow!(
                     "Parent state not found for block: {block_hash}, parent: {}",
-                    signed_block.message.parent_root
+                    signed_block.message.block.parent_root
                 )
             })?;
-        state.state_transition(&signed_block, true, true)?;
 
-        let attestations = signed_block.message.body.attestations.clone();
+        // TODO: Add `is_valid_signature` once spec is complete.
```
KolbyML: is_valid_signature for what object?

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -378,20 +391,29 @@ impl LeanChain {
...
+        state.state_transition(&signed_block.message.block, true, true)?;
+
+        let mut signed_attestations = vec![];
+        for attestation in &signed_block.message.block.body.attestations {
+            // let signature = signed_block.signature[index];
```
KolbyML: why is this comment here? what does it mean

crates/common/consensus/lean/src/block.rs
```diff
@@ -5,21 +5,30 @@ use ssz_types::{VariableList, typenum::U4096};
...
 use tree_hash_derive::TreeHash;
 
-use crate::attestation::SignedAttestation;
+use crate::attestation::Attestation;
 
-/// Represents a signed block in the Lean chain.
+/// Envelope carrying a block, an attestation from proposer, and aggregated signatures.
 ///
-/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/main/docs/client/containers.md#signedblock)
+/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/5da200c13f5eeda0b4139b1d55970d75c011d4b2/src/lean_spec/subspecs/containers/block/block.py#L93)
```
KolbyML: we should remove the link to the specification as this is inherently buggy and won't be maintained

crates/common/consensus/lean/src/block.rs
```diff
@@ -5,21 +5,30 @@ use ssz_types::{VariableList, typenum::U4096};
...
-    pub signature: FixedBytes<4000>,
+#[derive(Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode)]
+pub struct SignedBlockWithAttestation {
+    pub message: BlockWithAttestation,
+    pub signature: VariableList<FixedBytes<4000>, U4096>,
+}
+
+/// Bundle containing a block and the proposer's attestation.
+///
+/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/5da200c13f5eeda0b4139b1d55970d75c011d4b2/src/lean_spec/subspecs/containers/block/block.py#L83)
```
KolbyML: we should remove the link to the specification as this is inherently buggy and won't be maintained

crates/common/consensus/lean/src/block.rs
```diff
@@ -5,21 +5,30 @@ use ssz_types::{VariableList, typenum::U4096};
...
+#[derive(Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode)]
+pub struct BlockWithAttestation {
+    pub block: Block,
+    pub proposer_attestation: Attestation,
 }
 
 /// Represents a block in the Lean chain.
 ///
-/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/main/docs/client/containers.md#block)
+/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/5da200c13f5eeda0b4139b1d55970d75c011d4b2/src/lean_spec/subspecs/containers/block/block.py#L64)
```
KolbyML: we should remove the link to the specification as this is inherently buggy and won't be maintained

crates/common/consensus/lean/src/block.rs
```diff
@@ -36,7 +45,7 @@ pub struct Block {
 
 /// Represents a block header in the Lean chain.
 ///
-/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/main/docs/client/containers.md#blockheader)
+/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/5da200c13f5eeda0b4139b1d55970d75c011d4b2/src/lean_spec/subspecs/containers/block/block.py#L36)
```
KolbyML: we should remove the link to the specification as this is inherently buggy and won't be maintained, etc etc

crates/common/consensus/lean/src/state.rs
```diff
@@ -1015,19 +1012,16 @@ mod test {
...
-                        slot: 4,
-                        head: checkpoint4,
-                        target: checkpoint4,
-                        source: genesis_checkpoint,
-                    },
+            let attestations = Attestation {
+                validator_id: i,
+                data: AttestationData {
+                    slot: 4,
+                    head: checkpoint4,
```
KolbyML:
```suggestion
                    head: checkpoint_4,
```

etc, it is weird this is done inconsistently  in the test, apply whereever else this applies in the test

crates/common/validator/lean/src/service.rs
```diff
@@ -74,16 +75,20 @@ impl ValidatorService {
...
-                                    message: new_block,
-                                    signature: FixedBytes::default(),
+                                // TODO: Sign the block with the keystore once spec is finalized.
+                                let signed_block = SignedBlockWithAttestation {
+                                    message: BlockWithAttestation {
+                                        block: new_block,
+                                        proposer_attestation: Attestation::default(),
+
+                                    },
+                                    signature:VariableList::empty(),
```
KolbyML:
```suggestion
                                    signature: VariableList::empty(),
```

how is fmt ci passing

KolbyML (changes requested):
please remove the default trait from all the places you added it, the usage encourages bad practices which can lead to logical bugs in this example

Overall looks good though

crates/common/consensus/lean/src/state.rs
```diff
@@ -303,28 +301,27 @@ impl LeanState {
 
     pub fn process_attestations(
         &mut self,
-        attestations: &VariableList<SignedAttestation, U4096>,
+        attestations: &VariableList<Attestation, U4096>,
```
syjn99:
```suggestion
        attestations: &[Attestation],
```

Can we use like this, so that we don't have to stick to `VariableList` type?

crates/storage/src/tables/lean/lean_block.rs
```diff
@@ -14,9 +14,11 @@ use crate::{
 /// Table definition for the Lean Block table
 ///
 /// Key: block_id
-/// Value: [SignedBlock]
-pub(crate) const LEAN_BLOCK_TABLE: TableDefinition<SSZEncoding<B256>, SSZEncoding<SignedBlock>> =
-    TableDefinition::new("lean_block");
+/// Value: [SignedBlockWithAttestation]
```
syjn99, line 17: Can we make another table for mapping block with proposer attestation, so that this table only manages block hash and the corresponding block? This change leads some unnecessary changes, for example, at `get_fork_choice_head`. I think mapping only block id -> block matches well with current spec (`store.py`) as well.

crates/common/chain/lean/src/service.rs
varun-doshi: This is from the spec itself...it will eventually use the signature once signing and other functions are integrated in the specs

crates/storage/src/tables/lean/lean_block.rs
varun-doshi: Isn't it weird to have two tables holding very similar data with the only difference being a single field?
Difference between `SignedBlockWithAttestation` and `SignedBlock ` is mostly just Attestation

syjn99: hmm rethinking this, I agree with you as a block and an attestation from the proposer are tightly coupled with each other. I'd like to suggest some comments for better readability soon.

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -207,33 +212,33 @@ impl LeanChain {
                 lean_block_provider.get(self.safe_target)?.ok_or_else(|| {
                     anyhow!("Block not found for safe target hash: {}", self.safe_target)
                 })?;
-            if target_block.message.slot > safe_target_block.message.slot {
+            if target_block.message.block.slot > safe_target_block.message.block.slot {
```
syjn99:
```suggestion
            if target_block.slot > safe_target_block.slot {
```

Suggestion for better readability. All variables end with `_block` are actually typed `SignedBlockWithAttestation`, so I think we can access `.message.block` at assigning it. For example, `safe_target_block` can be assigned like:

```rust
let safe_target_block = lean_block_provider
    .get(self.safe_target)?
    .ok_or_else(|| {
        anyhow!("Block not found for safe target hash: {}", self.safe_target)
    })?
    .message
    .block;
```

What do you think? If this seems good, it would be best if we can apply this kind of changes all over this PR as this PR already modifies those lines.

syjn99: The very first [comment](https://github.com/ReamLabs/ream/pull/877#discussion_r2485207771) was from these readability issues that I faced while reviewing this PR. If we determine the type of table value as `SignedBlockWithAttestation`, I believe this can be much intuitive.

varun-doshi: Yeah understood...will make this change

varun-doshi: I've made this change wherever its possible to do so without actually doing a `clone`

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -264,21 +281,19 @@ impl LeanChain {
...
-                // Diverged from Python implementation: Using `B256::ZERO` instead of `None`)
-                state_root: B256::ZERO,
-                body: BlockBody::default(),
-            },
-            signature: FixedBytes::default(),
+        let mut new_block = Block {
+            slot,
+            proposer_index: slot % lean_network_spec().num_validators,
+            parent_root: head,
+            // Diverged from Python implementation: Using `B256::ZERO` instead of `None`)
```
syjn99:
```suggestion
```
This seems outdated at this moment, so we can delete this one as we're changing these lines.

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -287,43 +302,47 @@ impl LeanChain {
...
+            let mut new_signatures_to_add = Vec::new();
+
+            for (_, attestation) in latest_known_attestation_provider.get_all_attestations()? {
+                if attestation.message.source() == state.latest_justified
+                    && !new_block.body.attestations.contains(&attestation.message)
+                {
+                    new_attestations_to_add.push(attestation.message);
+                    new_signatures_to_add.push(attestation.signature);
+                }
+            }
```
syjn99:
```suggestion
            for signed_attestation in latest_known_attestation_provider
                .get_all_attestations()?
                .values()
            {
                if signed_attestation.message.source() == state.latest_justified
                    && !new_block
                        .body
                        .attestations
                        .contains(&signed_attestation.message)
                {
                    new_attestations_to_add.push(signed_attestation.message.clone());
                    new_signatures_to_add.push(signed_attestation.signature);
                }
            }
```

- Using `values()`
- Rename `attestation` -> `signed_attestation`

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -360,8 +379,11 @@ impl LeanChain {
     ///
     /// See lean specification:
     /// <https://github.com/leanEthereum/leanSpec/blob/ee16b19825a1f358b00a6fc2d7847be549daa03b/docs/client/forkchoice.md?plain=1#L314-L342>
-    pub async fn on_block(&mut self, signed_block: SignedBlock) -> anyhow::Result<()> {
-        let block_hash = signed_block.message.tree_hash_root();
+    pub async fn on_block(
+        &mut self,
+        signed_block: SignedBlockWithAttestation,
```
syjn99:
```suggestion
        signed_block_with_attestation: SignedBlockWithAttestation,
```

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -360,8 +379,11 @@ impl LeanChain {
...
+    ) -> anyhow::Result<()> {
+        let block_hash = signed_block.message.block.tree_hash_root();
```
syjn99:
```suggestion
        let block = signed_block.message.block;
        let block_hash = block.tree_hash_root();
```

This will make our life easier for the rest of this function body.

crates/storage/src/tables/lean/lean_block.rs
```diff
@@ -76,13 +78,13 @@ impl LeanBlockTable {
         for entry in table.iter()? {
             let (hash_entry, block_entry) = entry?;
             let hash: B256 = hash_entry.value();
-            let block: SignedBlock = block_entry.value();
+            let block: SignedBlockWithAttestation = block_entry.value();
```
syjn99:
```rust
            let block = block_entry.value().message.block;

            if block.parent_root != B256::ZERO
                && *attestation_weights.get(&hash).unwrap_or(&0) >= min_score
            {
                children_map
                    .entry(block.parent_root)
                    .or_default()
                    .push(hash);
            }
```

Suggestion for this section

crates/common/validator/lean/src/service.rs
```diff
@@ -74,16 +75,20 @@ impl ValidatorService {
                                     keystore.validator_id,
                                 );
 
-                                // TODO: Sign the block with the keystore.
-                                let signed_block = SignedBlock {
-                                    message: new_block,
-                                    signature: FixedBytes::default(),
+                                // TODO: Sign the block with the keystore once spec is finalized.
+                                let signed_block = SignedBlockWithAttestation {
```
syjn99:
```suggestion
                                let signed_block_with_attestation = SignedBlockWithAttestation {
```

crates/common/chain/lean/src/messages.rs
```diff
@@ -34,12 +34,12 @@ pub enum LeanChainServiceMessage {
         sender: oneshot::Sender<AttestationData>,
     },
     ProcessBlock {
-        signed_block: SignedBlock,
+        signed_block: Box<SignedBlockWithAttestation>,
```
syjn99:
```suggestion
        signed_block_with_attestation: Box<SignedBlockWithAttestation>,
```

crates/common/chain/lean/src/service.rs
```diff
@@ -205,7 +205,7 @@ impl LeanChainService {
 
     async fn handle_process_block(
         &mut self,
-        signed_block: SignedBlock,
+        signed_block: SignedBlockWithAttestation,
```
syjn99:
```suggestion
        signed_block_with_attestation: SignedBlockWithAttestation,
```

syjn99 (commented):
Only code styling comments, I think we're almost ready to go. Thanks!

crates/common/chain/lean/src/lean_chain.rs
varun-doshi: Did not go this route of looping because of the clone but sure

crates/common/chain/lean/src/lean_chain.rs
varun-doshi:
```
let block = signed_block.message.block.clone();
```
This will need to be cloned and hence did not do this but sure

syjn99:
```rust
    pub async fn on_block(
        &mut self,
        signed_block: SignedBlockWithAttestation,
    ) -> anyhow::Result<()> {
        let block = &signed_block.message.block;
        let block_hash = block.tree_hash_root();

        let (lean_block_provider, latest_justified_provider, lean_state_provider) = {
            let db = self.store.lock().await;
            (
                db.lean_block_provider(),
                db.latest_justified_provider(),
                db.lean_state_provider(),
            )
        };

        // If the block is already known, ignore it
        if lean_block_provider.contains_key(block_hash) {
            return Ok(());
        }

        let mut state = lean_state_provider.get(block.parent_root)?.ok_or_else(|| {
            anyhow!(
                "Parent state not found for block: {block_hash}, parent: {}",
                block.parent_root
            )
        })?;

        // TODO: Add signature validation once spec is complete.
        state.state_transition(&block, true, true)?;

        let mut signed_attestations = vec![];
        for attestation in &block.body.attestations {
            signed_attestations.push(SignedAttestation {
                message: attestation.clone(),
                signature: FixedBytes::<4000>::default(),
            });
        }
        lean_block_provider.insert(block_hash, signed_block)?;
        latest_justified_provider.insert(state.latest_justified)?;
        lean_state_provider.insert(block_hash, state)?;
        self.on_attestation_from_block(signed_attestations).await?;
        self.update_head().await?;

        Ok(())
    }
```

My bad, we might get the pointer from the `signed_block` like this.

crates/common/chain/lean/src/service.rs
KolbyML: We want to be spec compliant, not a 1 for 1 port of the spec from Python to Rust. "eventually" doesn't mean anything till the work is done

I don't think we should add dead code, which may never be used. We can add it back when it has a purpose.

This change was out of scope of the issue you made, and your PR title/description regardless

crates/common/chain/lean/src/lean_chain.rs
```diff
@@ -377,21 +405,27 @@ impl LeanChain {
...
+        let mut state = lean_state_provider.get(block.parent_root)?.ok_or_else(|| {
+            anyhow!(
+                "Parent state not found for block: {block_hash}, parent: {}",
+                block.parent_root
+            )
+        })?;
 
-        let attestations = signed_block.message.body.attestations.clone();
-        lean_block_provider.insert(block_hash, signed_block)?;
+        // TODO: Add signature validation once spec is complete.
```
KolbyML, line 415: Can we track this in an issue, and link the issue here

crates/common/consensus/lean/src/attestation.rs
```diff
@@ -7,7 +7,9 @@ use tree_hash_derive::TreeHash;
 use crate::checkpoint::Checkpoint;
 
 /// Attestation content describing the validator's observed chain view.
-#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
+#[derive(
+    Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash,
+)]
 pub struct AttestationData {
```
KolbyML: .

crates/common/consensus/lean/src/attestation.rs
```diff
@@ -16,7 +18,9 @@ pub struct AttestationData {
 }
 
 /// Validator specific attestation wrapping shared attestation data.
-#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash)]
+#[derive(
+    Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash,
+)]
 pub struct Attestation {
```
KolbyML: .

crates/common/consensus/lean/src/block.rs
```diff
@@ -63,7 +64,7 @@ impl From<Block> for BlockHeader {
 
 /// Represents the body of a block in the Lean chain.
 ///
-/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/main/docs/client/containers.md#blockbody)
+/// See the [Lean specification](https://github.com/leanEthereum/leanSpec/blob/5da200c13f5eeda0b4139b1d55970d75c011d4b2/src/lean_spec/subspecs/containers/block/block.py#L20)
 /// for detailed protocol information.
 #[derive(
     Debug, Default, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash,
```
KolbyML: .

KolbyML (commented):
This PR does a lot more then what is described in the PR's issue/title/description. The issue was to just update the containers, not the logic which was under this issue https://github.com/ReamLabs/ream/issues/847

^ which is pretty misleading, a PR's title/description/issue should encapsulate the work I would think

It isn't clear which my previous review was fully seen or not, so I will keep this look short

crates/common/chain/lean/src/lean_chain.rs
varun-doshi: https://github.com/ReamLabs/ream/issues/881

KolbyML: I meant a link in the code comment, but not a blocker

varun-doshi: Have added a link in the code comment

Issue #883, opened 2025-11-05
Update process_attestations() to match devnet-1 Spec


Kayden-ML:
### Describe the feature

Update process_attestations() to match devnet-1 Spec 

https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/containers/state/state.py

PR #884, opened 2025-11-05
feat: Update process_attestations() to match devnet-1 Spec

Kayden-ML:
### What was wrong?

 Fixes: https://github.com/ReamLabs/ream/issues/883

### How was it fixed?

I updated the process_attestations() to match devnet-1 Spec https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/containers/state/state.py

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5499,7 +5499,6 @@ dependencies = [
  "ethereum_ssz",
  "ethereum_ssz_derive",
  "itertools 0.14.0",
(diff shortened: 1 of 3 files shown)


KolbyML (approved):
:shipit: looks good. I am not sure if the old metrics make sense in terms of the new spec, can you make an issue and link to this PR in the description. Call the issue something like "Update process_attestations() metrics for devnet-1"

Issue #887, opened 2025-11-06
Update STF logic to latest spec


Kayden-ML:
### Describe the feature

https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/containers/state/state.py#L68

PR #889, opened 2025-11-06
feat: Update newly added STF logic for Devnet-1

Kayden-ML:
### What was wrong?

Fixes: https://github.com/ReamLabs/ream/issues/887

### How was it fixed?

I updated our STF to match the latest spec


KolbyML (approved):
:shipit: looks good

PR #893, opened 2025-11-07
feat: add lean head table to the database

Kayden-ML:
### What was wrong?

I'm currently implementing the updated fork choice logic and a few new types are introduced so I'm making tables 

### How was it fixed?

adds lean head table to the database


KolbyML (approved):
:shipit: looks good

PR #896, opened 2025-11-08
feat: reduce code duplication for database table trait

Kayden-ML:
### What was wrong?

Every time we added a new database table we copy and pasted the same boilerplate code

### How was it fixed?

Implemented default methods in trait so we don't have to copy and paste code


KolbyML (approved):
:shipit: looks good

PR #897, opened 2025-11-09
refactor: reduce code duplication for database field table

Kayden-ML:
### What was wrong?

Every time we added a new database field we copy and pasted the same boilerplate code

### How was it fixed?

Implemented default methods in trait so we don't have to copy and paste code


KolbyML:
> @KolbyML asking for my knowledge is this a "feat" or a "refactor"?

this is a refactor

KolbyML (approved):
:shipit: looks good

PR #899, opened 2025-11-09
fix: fixed deadlock issues

Kayden-ML:
### What was wrong?

image

### How was it fixed?

limited database lock scope
image


KolbyML (approved):
:shipit: looks good

PR #901, opened 2025-11-09
refactor: rename Leanchain to Store and move it to the lean fork choice 

Kayden-ML:
### What was wrong?

Our struct called LeanChain is just Store from the Spec. Store is apart of the fork choice so it shouldn't be in the chain crate

### How was it fixed?

renamed Leanchain to Store and move it to the lean fork choice


KolbyML (approved):
:shipit: looks good

Issue #741, opened 2025-08-25
Implement Lean ReqResp Initial Handshake


KolbyML:
### Describe the feature

```
Clients SHOULD immediately disconnect from one another following the handshake
above under the following conditions:

1. If the (`finalized_root`, `finalized_epoch`) shared by the peer is not in the
   client's chain at the expected epoch. For example, if Peer 1 sends (root,
   epoch) of (A, 5) and Peer 2 sends (B, 3) but Peer 1 has root C at epoch 3,
   then Peer 1 would disconnect because it knows that their chains are
   irreparably disjoint.

Once the handshake completes, the client with the lower `finalized_epoch` or
`head_slot` (if the clients have equal `finalized_epoch`s) SHOULD request blocks
from its counterparty via the `LeanBlocksByRoot` request.

*Note*: Under abnormal network condition or after some rounds of
`LeanBlocksByRoot` requests, the client might need to send `Status` request
again to learn if the peer has a higher head. Implementers are free to implement
such behavior in their own way.

```

https://github.com/leanEthereum/leanSpec/pull/8/files#diff-80d51cf6fd9015f791df4db30f7101a089f95f4bfe1ece79ddce8ee7665c76cfR176-R192

PR #906, opened 2025-11-12
feat: add devnet configuration

KolbyML:
### What was wrong?

We want to start implementing features for devnet 2 in parallel of devnet 1, for example https://github.com/ReamLabs/ream/issues/741
### How was it fixed?

adding the configuration


Kayden-ML (approved):
Looks good 👍

shariqnaiyer (approved):

PR #916, opened 2025-11-13
feat: use bytes in signature container

varun-doshi:
### What was wrong?

 
It would be difficult to use the existing `Signature` in Ream as internally it was [this](https://github.com/b-wagn/hash-sig/blob/23517cf35a4806c93a864f55ace5c4e4a67a08b3/src/signature/generalized_xmss.rs#L40)
This does not derive most of the macros we need (encode/decode, treehash). Hence, we need a compatible type for signature which we can serialize/deserialize from the hash-sig type

### How was it fixed?

 

Instead use a `FixedBytes` type which converts to the correct signature. 
Added a test for serialization roundtrip

### To-Do

 
### DIFF
--- a/crates/crypto/post_quantum/src/hashsig/signature.rs
+++ b/crates/crypto/post_quantum/src/hashsig/signature.rs
@@ -1,4 +1,3 @@
-use crate::hashsig::{HashSigScheme, public_key::PublicKey};
 use alloy_primitives::FixedBytes;
 use bincode::config::{Fixint, LittleEndian, NoLimit};
 use hashsig::{MESSAGE_LENGTH, signature::SignatureScheme};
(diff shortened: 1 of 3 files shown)


crates/crypto/post_quantum/src/hashsig/signature.rs
```diff
@@ -1,18 +1,74 @@
...
 
 type HashSigSignature = <HashSigScheme as SignatureScheme>::Signature;
 
+// NOTE: `GeneralizedXMSSSignature` doesn't implement methods like `to_bytes`,
+// which means we need to use bincode to serialize/deserialize it.
+// However, using bincode's default config (little-endian + variable int encoding)
+// add extra bytes to the serialized output, which is not what we want.
+// Thus, define a custom configuration for bincode here.
+const BINCODE_CONFIG: bincode::config::Configuration<LittleEndian, Fixint, NoLimit> =
+    bincode::config::standard().with_fixed_int_encoding();
```
KolbyML, line 19: this is already delcared in `private_key.rs` and `public_key.rs` could you remove it from both those files, and put it in the mod.rs so it can be reused

KolbyML (approved):
:shipit: looks good, I will just fix it, would be nice to get this in

Issue #923, opened 2025-11-17
Add Lean ReqResp handshake status handshake


KolbyML:
### Describe the feature

Implements the initial tracking of the status, and handle the status ping/pongs

### Additional context

[](url)

Issue #925, opened 2025-11-17
Add Lean ReqResp Handshake disconnect if peer is on disjoint chain


KolbyML:
### Describe the feature

Clients SHOULD immediately disconnect from one another following the handshake
above under the following conditions:

1. If the (`finalized_root`, `finalized_epoch`) shared by the peer is not in the
   client's chain at the expected epoch. For example, if Peer 1 sends (root,
   epoch) of (A, 5) and Peer 2 sends (B, 3) but Peer 1 has root C at epoch 3,
   then Peer 1 would disconnect because it knows that their chains are
   irreparably disjoint.

Issue #924, opened 2025-11-17
Add Lean ReqResp Initial Handshake, request missing blocks case


KolbyML:
### Describe the feature

Once the handshake completes, the client with the lower `finalized_epoch` or
`head_slot` (if the clients have equal `finalized_epoch`s) SHOULD request blocks
from its counterparty via the `LeanBlocksByRoot` request.

KolbyML:
With syncing I consider this closed

PR #926, opened 2025-11-17
feat: Add Lean ReqResp handshake status handshake

KolbyML:
### What was wrong?

Fixes https://github.com/ReamLabs/ream/issues/923

### How was it fixed?

- add a `NetworkState` struct for holding data needed for p2p functions like ping/pongs using status messages, and peer tables
- enable sending a status message on dialing a per once connected if devnet2 is enabled

future work
- https://github.com/ReamLabs/ream/issues/925
- https://github.com/ReamLabs/ream/issues/924


bin/ream/src/main.rs
```diff
@@ -185,37 +186,25 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor, ream_
...
-
     // Initialize the services that will run in the lean node.
     let (chain_sender, chain_receiver) = mpsc::unbounded_channel::<LeanChainServiceMessage>();
     let (outbound_p2p_sender, outbound_p2p_receiver) = mpsc::unbounded_channel::<LeanP2PRequest>();
 
-    let chain_service =
-        LeanChainService::new(lean_chain_writer, chain_receiver, outbound_p2p_sender).await;
+    // Initialize the lean network service
+
+    let head_block_hash = lean_db
```
varun-doshi, line 195: Just a small doubt here; would this not fail on starting up a new node?
Because there will not be any data in the db when starting a fresh node

Am I missing something here?

KolbyML: you are right I will fix this, I will make an issue to add a regresion test to prevent this from happening in the future

varun-doshi: I was just about to open a small Pr to fix...i can do that if you dont mind
You may add the regression test issue

varun-doshi: https://github.com/ReamLabs/ream/pull/927

PR #927, opened 2025-11-17
fix: use default for network service

varun-doshi:
### What was wrong?

 
image

### How was it fixed?

 
The node was failing on startup because there is no data in the DB when running a fresh node. 
This PR initiates the network service manager with default Checkpoints on startup which is fine I think since there is no `finalized` root or  `head` root on startup.
The node is running after this fix and also finalizing:
image

### To-Do

 

KolbyML:
Hi Varun I made a PR which fixes it in a different way and added a regression test https://github.com/ReamLabs/ream/pull/928

I don't see your message till I finished

This is a awkward situation. I do prefer my PR because
- I think the solution is better structured
- I added a regression test

PR #934, opened 2025-11-18
refactor: remove redundant lean_/beacon_ prefix on db providers

KolbyML:
### What was wrong?

We have LeanDB and BeaconDB, the lean_/beacon_ prefix on the providers is redundant
### How was it fixed?

remove the redundant prefix


Kayden-ML (approved):
Looks good 👍

PR #938, opened 2025-11-18
feat: add lean metrics

### LINKED ISSUE #836: Add initial leanMetrics tracking
### Describe the feature

https://github.com/leanEthereum/leanMetrics/pull/4/files
https://github.com/leanEthereum/leanMetrics/pull/3/files

The above draft PR's by Katya propose a set of metrics that we should track for the PQ interop. 

We want to start tracking these metrics in our Lean Client. 

### Additional context

_No response_

shariqnaiyer:
### What was wrong?

Fixes #836 

### How was it fixed?

Added the lean metrics tracking as per the spec and tested it with Katya's Dashboard.

I have additionally included Katya's lean metrics dashboard for anyone to test.

### To-Do

 

crates/common/consensus/lean/src/state.rs
```diff
@@ -284,11 +304,15 @@ impl LeanState {
...
 
     pub fn process_attestations(&mut self, attestations: &[Attestation]) -> anyhow::Result<()> {
+        let timer = start_timer_vec(&STATE_TRANSITION_ATTESTATIONS_PROCESSING_TIME, &[]);
+
         // get justifications, justified slots and historical block hashes are
         // already up to date as per the processing in process_block_header
         let mut justifications_map = self.get_justifications()?;
 
         for attestation in attestations {
+            let validation_timer = start_timer_vec(&ATTESTATION_VALIDATION_TIME, &[]);
```
varun-doshi: `validate_attestation_timer`

crates/common/consensus/lean/src/state.rs
```diff
@@ -284,11 +304,15 @@ impl LeanState {
     }
 
     pub fn process_attestations(&mut self, attestations: &[Attestation]) -> anyhow::Result<()> {
+        let timer = start_timer_vec(&STATE_TRANSITION_ATTESTATIONS_PROCESSING_TIME, &[]);
```
varun-doshi: rename to `attestation_processing_timer`

crates/common/consensus/lean/src/state.rs
```diff
@@ -160,18 +166,24 @@ impl LeanState {
         block: &Block,
         valid_signatures: bool,
     ) -> anyhow::Result<()> {
+        let timer = start_timer_vec(&STATE_TRANSITION_TIME, &[]);
```
varun-doshi: rename to `stf_timer` or `state_transition_timer`

unnawut: Imho I'm okay with `timer` though because it's in the scope of `state_transition()`, there's no other ambiguity.

Same opinion with the rest of the renaming suggestions.

crates/common/consensus/lean/src/state.rs
```diff
@@ -284,11 +304,15 @@ impl LeanState {
     }
 
     pub fn process_attestations(&mut self, attestations: &[Attestation]) -> anyhow::Result<()> {
```
unnawut, line 223: We're adding metrics to `state.process_attestations()` but I think leanMetrics want attestation valid/invalid on fork choice's on validate attestation. I believe state operations happen a lot more often e.g. when building a block, while fork choice on_attestation happens only when receiving a new block, so the number might be skewed.

Can you double check with leanMetrics again?

crates/common/consensus/lean/src/state.rs
KolbyML: We should use `timer`, unless there is 2 timers which elipse each other for example

x_t = start_timer()
y_timer = start_timer()
stop_timer(y_timer)
stop_timer(x_timer)

^ this is the case where we should have `state_transition_timer`, if the code is simple `timer` is fine

crates/common/consensus/lean/src/state.rs
```diff
@@ -304,6 +328,8 @@ impl LeanState {
                     "Skipping attestations by Validator {}",
                     attestation.validator_id,
                 );
+                inc_int_counter_vec(&ATTESTATIONS_INVALID_TOTAL, &[]);
+                stop_timer(validation_timer);
```
KolbyML: we don't need to call stop_timer here https://docs.rs/prometheus/0.13.4/prometheus/struct.HistogramTimer.html ``This timer can be stopped and observed at most once, either **automatically (when it goes out of scope)** or manually``

Same for the other few places below

crates/common/fork_choice/lean/src/store.rs
```diff
@@ -536,7 +552,9 @@ impl Store {
         let proposer_attestation = &signed_block_with_attestation.message.proposer_attestation;
         let block_root = block.tree_hash_root();
 
+        // If the block is already known, ignore it
         if block_provider.get(block_root)?.is_some() {
+            stop_timer(block_processing_timer);
```
KolbyML: we don't need this stop timer here, read comment above

crates/common/metrics/src/lib.rs
```diff
@@ -58,3 +139,15 @@ pub fn start_timer_vec(histogram_vec: &HistogramVec, label_values: &[&str]) -> H
 pub fn stop_timer(timer: HistogramTimer) {
     timer.observe_duration()
 }
+
+/// Create a new counter metric
+pub fn create_int_counter_vec(name: &str, help: &str, label_names: &[&str]) -> IntCounterVec {
+    let registry = default_registry();
+    register_int_counter_vec_with_registry!(name, help, label_names, registry)
+        .expect("failed to create int counter vec")
+}
```
KolbyML: we don't need this function, just inline it

KolbyML (commented):
here is some feedback

crates/common/consensus/lean/src/state.rs
shariqnaiyer: I believe you are correct as it explicitly says validate attestation so I have moved it there in the recent commit.

PR #939, opened 2025-11-19
feat: add DiscardOnDropHistogramTimer

KolbyML:
### What was wrong?

In certain situations we don't want to record the metrics timer, if the timer is dropped
### How was it fixed?

add DiscardOnDropHistogramTimer


Kayden-ML (approved):
Looks good 👍

PR #943, opened 2025-11-20
feat: use leanSig instead of hashSig

varun-doshi:
### What was wrong?

 
Hash Sig was moved to the official Lean Ethereum org here
https://github.com/leanEthereum/leanSig

We should use this as our hash sig instead of the earlier one
### How was it fixed?

 
Use the new hash sig library

### To-Do

 

KolbyML (approved):
I rebased the PR and I think the CI will pass and we can merge this

PR #962, opened 2025-11-24
feat: Add Lean ReqResp Handshake disconnect if peer is on disjoint chain

KolbyML:
### What was wrong?

Resolves 
- https://github.com/ReamLabs/ream/issues/925

### How was it fixed?

Check if we have the slot indexed, if it is check if the block matches, if assume the other node is canonical (this is a bit naive, but we should over engineer this this early on), if the slot is less then 5 consider the node canonical as a special case


Kayden-ML (approved):
Looks good 👍

PR #966, opened 2025-11-25
fix: code was panicking with the error

Kayden-ML:
### What was wrong?

docker: Error response from daemon: pull access denied for my-app, repository does not exist or may require 'docker login'

### How was it fixed?

fixes error that was happening with our get_current_slot()


KolbyML (approved):
:shipit: looks good

Issue #976, opened 2025-11-27
Test that Ream generates the same genesis state root as Zeam


KolbyML:
### Describe the feature

https://github.com/leanEthereum/leanSpec/pull/195

PR #919, opened 2025-11-15
test: add lean spec test for fork choice

shariqnaiyer:
### What was wrong?

Fixes part of #867 

### How was it fixed?

To run the tests go to lean-spec-test directory and run `make test`

This PR is also a partial PR as it doesn't run the state transition tests. I believe those are simpler and I plan to make a PR for that after this one.

### To-Do

 
- [x] I have read [CONTRIBUTING.md](https://github.com/ReamLabs/ream/blob/master/CONTRIBUTING.md).
- [x] This PR title follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

PR #978, opened 2025-11-27
fix: invalid genesis root + add genesis tests

KolbyML:
### What was wrong?

fixes https://github.com/ReamLabs/ream/issues/976

### How was it fixed?

Ok I refound out why we were generating the wrong state root. It was because we were using the wrong limit for the VariableList of valiators in the state. Originally @shariqnaiyer found this bug, but we didn't merge it because it was in a bigger PR integrating fork_choice tests the data was generated without signatures https://github.com/ReamLabs/ream/pull/919/files

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5433,6 +5433,7 @@ dependencies = [
  "ream-fork-choice-lean",
  "ream-metrics",
  "ream-network-spec",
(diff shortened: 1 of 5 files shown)


Kayden-ML (approved):
Looks good 👍

Issue #979, opened 2025-11-27
Add Swatinem/rust-cache


KolbyML:
### Describe the feature

Add Swatinem/rust-cache speed up ci 🥺

PR #980, opened 2025-11-27
ci: add Swatinem/rust-cache

KolbyML:
### What was wrong?

fixes https://github.com/ReamLabs/ream/issues/979

CI takes longer then I prefer
 

### How was it fixed?

add Swatinem/rust-cache


Kayden-ML (approved):
Looks good 👍

PR #992, opened 2025-11-28
feat(breaking): remove bincode and use ssz instead

KolbyML:
### What was wrong?

We are migrating from bincode to SSZ depends

### How was it fixed?

replacing bincode with ssz alternatives


shariqnaiyer (approved):
Looks good 👍

Issue #993, opened 2025-11-28
Pull ephemery() config's public keys from config.yaml


KolbyML:
### Describe the feature

Pull ephemery() config's public keys from config.yaml

PR #994, opened 2025-11-28
feat: update network config to read validators from config, instead of hardcoded

KolbyML:
### What was wrong?

fixes https://github.com/ReamLabs/ream/issues/993

It was annoying to need to manually update the hardcoded emphemy validators each time
### How was it fixed?

read from the config.yaml file in assets


Kayden-ML (approved):
Looks good 👍

Issue #914, opened 2025-11-13
Implement fork choice tests from test_validator.py


shariqnaiyer:
### Describe the feature

https://github.com/leanEthereum/leanSpec/blob/main/tests/lean_spec/subspecs/forkchoice/test_validator.py

varun-doshi:
Completed by https://github.com/ReamLabs/ream/pull/1023, https://github.com/ReamLabs/ream/pull/1008 and https://github.com/ReamLabs/ream/pull/986

PR #1008, opened 2025-11-29
feat(tests): Add leanspec attestation tests

varun-doshi:
### What was wrong?
Part of https://github.com/ReamLabs/ream/issues/914

How was it fixed?

Adds tests for attestaions in lean

### To-Do

 
### DIFF
--- a/crates/common/fork_choice/lean/src/store.rs
+++ b/crates/common/fork_choice/lean/src/store.rs
@@ -765,11 +765,9 @@ mod tests {
         let (genesis_block, genesis_state) =
             setup_genesis(0, generate_default_validators(no_of_validators));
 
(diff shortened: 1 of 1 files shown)


KolbyML:
> Note: although most of the times slot and validator_idx are used only once, but having them in a variable seems easier to read and follow than floating numbers in the code

If the variable is used once, I don't think wee need it, I am going to clean up this PR quick and merge it

KolbyML:
> Note: although most of the times slot and validator_idx are used only once, but having them in a variable seems easier to read and follow than floating numbers in the code

also we should either use i or validator_id or validator_index or index

KolbyML (approved):
:shipit: looks good

KolbyML:
Also for test pr's we should use the `test:` tag

PR #1022, opened 2025-12-03
feat: use ssz binary files instead of json files for storing private keys

KolbyML:
### What was wrong?

Hash-sig-cli is updating to support private keys stored as in ssz binaries rather then json files https://github.com/blockblaz/hash-sig-cli/pull/28

### How was it fixed?

read/write ssz binaries to files, instead of json for private keys


Kayden-ML (approved):
Looks good 👍

PR #1026, opened 2025-12-05
feat: implement validator/beacon_committee_subscriptions

akronim26:
### What was wrong?

Implementation of `validator/beacon_committee_subscriptions` was absent

### How was it fixed?

- While implementing `validator/beacon_committee_subscriptions`, I noticed that the networking layer currently only maintains static node-level attestation subnets via `DiscoveryConfig`, and does not yet implement dynamic validator attestation subnet subscriptions. Therefore, this endpoint currently performs committee validation, subnet computation, and grouping. 

Could you confirm whether this endpoint should remain as it is for now, or if you prefer that we extend the networking layer in this issue as well?

### To-Do

 
- [X] This PR title follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

### DIFF
--- a/crates/rpc/beacon/src/handlers/validator.rs
+++ b/crates/rpc/beacon/src/handlers/validator.rs
@@ -665,7 +665,7 @@ pub async fn post_beacon_committee_subscriptions(
 
         subnet_to_subscriptions
             .entry(subnet_id)
-            .or_insert_with(Vec::new)
+            .or_default()
             .push(sub);
     }
 


crates/rpc/beacon/src/handlers/validator.rs
```diff
@@ -621,3 +627,52 @@ pub async fn post_aggregate_and_proofs_v2(
...
+            compute_subnet_for_attestation(sub.committees_at_slot, sub.slot, sub.committee_index);
+
+        subnet_to_subscriptions
+            .entry(subnet_id)
+            .or_default()
+            .push(sub);
+    }
+
+    // TODO
+    // add support for attestation subnet subscriptions for validators
```
KolbyML, line 673: Can you make an issue for this. I can assign you if you want

KolbyML (approved):
:shipit: looks good

crates/rpc/beacon/src/handlers/validator.rs
akronim26: Yeah, sure! @KolbyML, I created issue - #1043 for the same

Issue #915, opened 2025-11-13
Implement gossipsub testing as per leanSpec tests


shariqnaiyer:
### Describe the feature

https://github.com/leanEthereum/leanSpec/blob/main/tests/lean_spec/subspecs/networking/test_gossipsub.py

PR #1042, opened 2025-12-09
test: Add gossipsub tests according to leanspec

Dsorken:
### What was wrong?

https://github.com/ReamLabs/ream/issues/915

### How was it fixed?

Implemented the gossipsub tests in accordance with leanspec, added helper function in the lean network module to match the beacon equivalent

Some tests or test aspects were not relevant in the context of how Ream is setup and would be trivial and/or would be redundantly testing the libp2p library:

##### test_message_id_computation
The test assumes that snappy decompress is done on a message to message basis, but in Ream it is done at a much higher level before the messages are formed, which adds certain guarantees of id computation that the specs do not assume exist, the parts to test for this specific behavior are omitted

##### test_message_id_caching
This is handled in libp2p which does not explicitly cache messages, the specific part about caching is omitted

##### test_realistic_blockchain_scenarios
This test primarily tests for different results from using or not using snappy, but again assumes snappy is done a message to message basis which is not done in Ream. However it would be trivial to verify that hashing compressed vs non compressed data will produce different results

### To-Do

 
### DIFF
--- a/crates/common/network_spec/src/networks/lean.rs
+++ b/crates/common/network_spec/src/networks/lean.rs
@@ -24,7 +24,9 @@ pub fn set_lean_network_spec(network_spec: Arc<LeanNetworkSpec>) {
         );
     }
     HAS_NETWORK_SPEC_BEEN_INITIALIZED.call_once(|| {
-        set_lean_network_spec(network_spec);
+        LEAN_NETWORK_SPEC
+            .set(network_spec)
+            .expect("LeanNetworkSpec should be set only once at the start of the application");
     });
 }
 


crates/common/network_spec/src/networks/lean.rs
```diff
@@ -138,3 +141,6 @@ impl<'de> Deserialize<'de> for DiscardUnknown {
         Ok(DiscardUnknown)
     }
 }
+
+pub static TEST: LazyLock<Arc<LeanNetworkSpec>> =
+    LazyLock::new(|| LeanNetworkSpec::ephemery().into());
```
KolbyML, line 146: Since this is used nowhere else, y ou could honestly just inline it in the function

KolbyML (approved):
:shipit: looks good

PR #1050, opened 2025-12-11
feat: if service panics stop the node + clean up interval panic log

KolbyML:
### What was wrong?

- when one of ream's services paniced, the node would keep on running like nothing has ever happened
- our  logs weren't intuitive

### How was it fixed?

- remove a few panics, and handle if a service quits
- update log message


Kayden-ML (approved):
Looks good 👍

PR #1064, opened 2025-12-15
feat(api): implement "/eth/v3/validator/blocks/{slot}"

akronim26:
### What was wrong?

 Implementation of endpoint `/eth/v3/validator/blocks/{slot}` was absent.

### How was it fixed?

This PR implements the `/eth/v3/validator/blocks/{slot}` endpoint. Although, as per the current implementation, fields such as `attestations`, `deposits`, and `sync_aggregate` are populated by default values due to absence of their fetching logic in the `OperationPool`. If this approach is acceptable, I can create follow-up issues to track these missing pieces and implement them.

### To-Do

 
- [X] This PR title follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

### DIFF
--- a/crates/rpc/beacon/src/handlers/validator.rs
+++ b/crates/rpc/beacon/src/handlers/validator.rs
@@ -1050,14 +1050,14 @@ fn calculate_consensus_block_value(
                 .collect();
 
             for index in slashed_indices {
-                if let Some(validator) = state.validators.get(index as usize) {
-                    if validator.is_slashable_validator(current_epoch) {
-                        total_reward = total_reward.saturating_add(
-                            validator
-                                .effective_balance
-                                .saturating_div(WHISTLEBLOWER_REWARD_QUOTIENT),
-                        );
-                    }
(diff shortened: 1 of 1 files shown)


crates/rpc/beacon/src/handlers/validator.rs
```diff
@@ -925,3 +958,541 @@ pub async fn post_contribution_and_proofs(
...
+        blob_gas_used: local_payload_v4.execution_payload.blob_gas_used,
+        excess_blob_gas: local_payload_v4.execution_payload.excess_blob_gas,
+    };
+
+    let blob_kzg_commitments: Vec<KZGCommitment> = local_payload_v4
+        .blobs_bundle
+        .commitments
+        .iter()
+        .map(|c| {
+            let vec = c.clone().to_vec();
```
shariqnaiyer: We want to avoid using shorthands like 'c'. This should be expanded to 'commitment'. Please fix this in other places across this PR as well.

KolbyML (approved):
:shipit: looks good

PR #1073, opened 2025-12-16
fix: add Err variable name check in make fmt

### LINKED ISSUE #1071: Add `Err(e)` -> `Err(err)` to `make fmt`, like we have for `map_err(e)` to `map_err(err)`. `make fmt` is used in CI so that is all that is needed
### Describe the feature

Add `Err(e)` -> `Err(err)` to `make fmt`, like we have for `map_err(e)` to `map_err(err)`. `make fmt` is used in CI so that is all that is needed

### Additional context

_No response_

aditya172926:
### What was wrong?

 
closes #1071 
Err(e) to Err(err)

### How was it fixed?

 
Added the regex check in make fmt

The check would consider the syntax like `Err(e)`, `Err(error)`, `Err(xyz)` and similar as violations and recommend the correct form.

So syntax like this which can be used in enums, or return types are not considered violations
examples that are not considered violations
```
            return Err(ValidatorError::RequestFailed {
                status_code: response.status(),
            });
```
```
enum {
Err(Xyz)
}
```

### To-Do

 
### DIFF
--- a/crates/common/polynomial_commitments/src/trusted_setup.rs
+++ b/crates/common/polynomial_commitments/src/trusted_setup.rs
@@ -29,7 +29,7 @@ macro_rules! impl_settings {
 
                 match output {
                     Ok(settings) => settings,
-                    Err(error) => panic!("failed to load kzg trusted setup: {error}"),
+                    Err(err) => panic!("failed to load kzg trusted setup: {err}"),
                 }
             })
         }
--- a/crates/common/validator/beacon/src/validator.rs
+++ b/crates/common/validator/beacon/src/validator.rs
@@ -275,18 +275,18 @@ impl ValidatorService {
     // Runs at the start of every slot
     pub async fn on_slot(&mut self, slot: u64) {
         info!("Current Slot: {slot}");
-        if let Err(sync_error) = self.prepare_sync_infos(slot - 1).await {
-            warn!("Could not prepare the sync infos: {sync_error:?}");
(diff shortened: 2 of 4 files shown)


KolbyML:
> The CI exits like this after the make fmt check
> 
> ```
> cargo +nightly fmt --all
> Invalid Err variable naming found:
> ./crates/rpc/beacon/src/handlers/validator.rs:945:            Err(message) => {
> ./crates/rpc/beacon/src/handlers/block.rs:414:        Err(e) => {
> ./crates/common/polynomial_commitments/src/trusted_setup.rs:32:                    Err(error) => panic!("failed to load kzg trusted setup: {error}"),
> ./crates/common/validator/beacon/src/validator.rs:278:        if let Err(sync_error) = self.prepare_sync_infos(slot - 1).await {
> ./crates/common/validator/beacon/src/validator.rs:280:        } else if let Err(sync_error) = self.process_normal_sync_infos(slot - 1).await {
> ./crates/common/validator/beacon/src/validator.rs:288:        if let Err(sync_error) = self.process_aggregator_sync_infos(slot - 1).await {
> ./crates/common/validator/beacon/src/validator.rs:491:                        Err(signing_error) => Some(Err(anyhow!(
> 
> make: *** [Makefile:51: fmt] Error 1
> Only this form is allowed:
>     Err(err)
> ```

can you fix ci now?

KolbyML (approved):
:shipit: looks good

PR #1080, opened 2025-12-17
docs: clarify parent block validation for data column sidecar

### LINKED ISSUE #1061: Implement parent block validation for data column sidecar
## Summary

Add validation to reject data column sidecars whose parent block fails validation.

## Context

In `crates/networking/manager/src/gossipsub/validate/data_column_sidecar.rs`, the `validate_data_column_sidecar_full` function is missing a validation step from the consensus specs.

Per the [PeerDAS networking spec](https://ethereum.github.io/consensus-specs/specs/fulu/p2p-interface/#data_column_sidecar_subnet_id), data column sidecars should be rejected if:
> The sidecar's block's parent (defined by `block_header.parent_root`) passes validation.

## Current behavior

The code currently checks if the parent block exists and has a lower slot, but does not fully validate that the parent block itself passes all validation rules.

## Expected behavior

> TODO understand if parent block seen is enough, since previous blocks might be already validated before adding to the db.

Implement the full parent block validation check and reject sidecars with invalid parent blocks.

## Location
  https://github.com/ReamLabs/ream/blob/c7925b2e1d7b08d6788e8162cf2b466d5cf685bc/crates/networking/manager/src/gossipsub/validate/data_column_sidecar.rs#L62-L63

zemse:
### What was wrong?

"TODO comment" for parent block validation in data column sidecar gossip validation. Closes #1061

### How was it fixed?

Documented that checking parent block existence in `block_provider` is sufficient since blocks are only stored after passing validation.

### To-Do

 
### DIFF
--- a/crates/networking/manager/src/gossipsub/validate/data_column_sidecar.rs
+++ b/crates/networking/manager/src/gossipsub/validate/data_column_sidecar.rs
@@ -58,9 +58,9 @@ pub async fn validate_data_column_sidecar_full(
         ));
     }
 
-    // TODO [REJECT] The sidecar's block's parent (defined by block_header.parent_root) passes
-    // validation.
-
+    // The sidecar's block's parent (defined by block_header.parent_root) passes validation.
+    // Blocks are only added to block_provider after passing full validation (ST func)
+    // so existence in the store implies validation passed. If parent is not in store, we IGNORE.
     let Some(parent_block) = store.db.block_provider().get(header.parent_root)? else {
         return Ok(ValidationResult::Ignore(
             "Parent block not seen".to_string(),


crates/networking/manager/src/gossipsub/validate/data_column_sidecar.rs
```diff
@@ -58,9 +58,9 @@ pub async fn validate_data_column_sidecar_full(
         ));
     }
 
-    // TODO [REJECT] The sidecar's block's parent (defined by block_header.parent_root) passes
-    // validation.
-
+    // The sidecar's block's parent (defined by block_header.parent_root) passes validation.
+    // Blocks are only added to block_provider after passing full validation (ST func)
```
KolbyML, line 62: In the future, please don't use the shorthand ST for state transition 👍

KolbyML (approved):

PR #1085, opened 2025-12-18
feat(lean): Update Block Production

### LINKED ISSUE #1025: Update Block Production
### Describe the feature

Context: https://github.com/leanEthereum/leanSpec/pull/218

### Additional context

_No response_

varun-doshi:
### What was wrong?

 
Fixes #1025 

### How was it fixed?

 
Context for this change: https://github.com/leanEthereum/leanSpec/pull/218

### To-Do

 
### DIFF
--- a/crates/common/fork_choice/lean/src/store.rs
+++ b/crates/common/fork_choice/lean/src/store.rs
@@ -3,8 +3,8 @@ use std::{collections::HashMap, sync::Arc};
 use alloy_primitives::B256;
 use anyhow::{anyhow, ensure};
 use ream_consensus_lean::{
-    attestation::{AttestationData, SignedAttestation},
-    block::{BlockWithSignatures, SignedBlockWithAttestation},
+    attestation::{Attestation, AttestationData, SignedAttestation},
+    block::{Block, BlockBody, BlockWithSignatures, SignedBlockWithAttestation},
     checkpoint::Checkpoint,
     state::LeanState,
     validator::is_proposer,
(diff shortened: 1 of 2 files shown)


crates/storage/src/tables/lean/lean_block.rs
```diff
@@ -115,4 +115,15 @@ impl LeanBlockTable {
...
+    pub fn get_block_roots(&self) -> Result<Vec<B256>, StoreError> {
+        let read_txn = self.db.begin_read()?;
+        let table = read_txn.open_table(Self::TABLE_DEFINITION)?;
+        let mut block_roots = Vec::new();
+        for entry in table.iter()? {
+            let (key, _) = entry?;
+            block_roots.push(key.value());
+        }
+        Ok(block_roots)
+    }
```
KolbyML: We probably shouldn't be appending every block root in the database to a vector

varun-doshi: This is the issue we spoke about in the previous PR...I cannot call db functions from inside state due to circular dependency

KolbyML: Put the build function in the Store store struct then, the State shouldn't have access to storage

varun-doshi: updated as per your suggestion...let me know if this is what you meant

KolbyML: Could you put it in the `pub struct Store {` instead of the `pub struct LeanDB {` struct

PR #1099, opened 2025-12-28
feat: remove --devnet flag and use feature flags instead

KolbyML:
### What was wrong?

using a flag to configure devnet's is too much work for supporting devnet's which we will be throwing away a lot of legacy code after we are done with the old devnet
### How was it fixed?

use feature flags


Kayden-ML (approved):
Looks good 👍

Issue #886, opened 2025-11-05
Count down until genesis


syjn99:
### Describe the feature

Running `lean_node` seems stuck until the genesis timestamp reaches, so I think we might add a tiny countdown logs until there.

mw2000:
is this still being worked on? happy to take this on

syjn99:
Closed by #1104

PR #1104, opened 2026-01-03
feat: add countdown log till genesis

MuhtasimTanmoy:
### What was wrong?
https://github.com/ReamLabs/ream/issues/886

### How was it fixed?
https://github.com/ReamLabs/ream/pull/1104

### To-Do
Review

### DIFF
--- a/bin/ream/src/main.rs
+++ b/bin/ream/src/main.rs
@@ -90,9 +90,9 @@ use ream_validator_lean::{
     registry::load_validator_registry, service::ValidatorService as LeanValidatorService,
 };
 use ssz_types::VariableList;
-use tokio::time;
 use tokio::{
     sync::{broadcast, mpsc},
+    time,
     time::Instant,
 };
 use tracing::{error, info};
@@ -718,7 +718,7 @@ pub async fn countdown_for_genesis() {
         let genesis = lean_network_spec().genesis_time;
 
         if now >= genesis {
-            // Only log the "Genesis reached" message if we are starting within 
+            // Only log the "Genesis reached" message if we are starting within
             // a small 2-second window of the actual event.
             if now <= genesis + 2 {
                 info!("Genesis reached! Starting services...");


bin/ream/src/main.rs
```diff
@@ -703,6 +706,34 @@ pub async fn run_generate_private_key(config: GeneratePrivateKeyConfig) {
     process::exit(0);
 }
 
+// Countdown logs until the genesis timestamp reaches
+pub async fn countdown_for_genesis() {
+    loop {
+        let now = SystemTime::now()
+                      .duration_since(UNIX_EPOCH)
+                      .expect("System time is before UNIX epoch")
+                      .as_secs();
```
KolbyML: Isn't this `expect()` comment misleading?, I think we should avoid expect here, handle the exception intentionally and safely

MuhtasimTanmoy: Referenced from here.
https://github.com/ReamLabs/ream/blob/d131153f887202192543cb362cdd6c700a79bcf1/crates/common/network_spec/src/networks/lean.rs#L84

KolbyML: Yeah I will be removing those in a PR I am working on

MuhtasimTanmoy:
```rust
  let now = SystemTime::now()
                      .duration_since(UNIX_EPOCH)
                      .unwrap_or(Duration::MAX)// remaining <= 0 will be true if error
                      .as_secs();
        let remaining = lean_network_spec().genesis_time - now;

        if remaining <= 0 {
            info!("Genesis reached! Starting services...");
            break;
        }
```

Does this look ok?  Or should safely log error for that specific case?

KolbyML: let remaining = lean_network_spec().genesis_time.saturating_sub(now);

You should do a saturating sub then to avoid underflows

MuhtasimTanmoy:
```rust
let now = SystemTime::now()
                      .duration_since(UNIX_EPOCH)
                      .unwrap_or(Duration::MAX)
                      .as_secs();
let remaining = lean_network_spec().genesis_time.saturating_sub(now);

if remaining == 0 {
    info!("Genesis reached! Starting services...");
    break;
}
```

If this looks okay, then will update the PR.

KolbyML: Remove the `// remaining <= 0 will be true if error` 

also if `now == Duration::MAX` we should not print anything and just immediately break


So maybe do

```
 let now = match SystemTime::now()
                      .duration_since(UNIX_EPOCH) {
                           Ok(now) => now.as_secs(),
                           // The chain is already past genesis, we don't need a countdown
                           Err(_) => break,
                      };
                      
let remaining = lean_network_spec().genesis_time.saturating_sub(now);

if remaining == 0 {
    info!("Genesis reached! Starting services...");
    break;
}
```

^ so do something like this

KolbyML: I didn't suggest that looking at the full context, so you might need to make further modifications

MuhtasimTanmoy: Thanks.

MuhtasimTanmoy: I tested.

bin/ream/src/main.rs
```diff
@@ -703,6 +708,36 @@ pub async fn run_generate_private_key(config: GeneratePrivateKeyConfig) {
...
+            // The chain is already past genesis, we don't need a countdown
+            Err(_) => break,
+        };
+
+        let remaining = lean_network_spec().genesis_time.saturating_sub(now);
+
+        if remaining == 0 {
+            info!("Genesis reached! Starting services...");
+            break;
+        }
```
```suggestion
let now = SystemTime::now()
    .duration_since(UNIX_EPOCH)
    .unwrap_or(Duration::MAX)
    .as_secs();

let genesis = lean_network_spec().genesis_time;

if now >= genesis {
    if now <= genesis + 2 {
        info!("Genesis reached! Starting services...");
    }
    break;
}
```

After reading the code a bit more, could you do this ^

MuhtasimTanmoy: Generally, `now` will catch up to `genesis` and `now == genesis` should trigger `info!("Genesis reached! Starting services...");`.

Should add any comments on why `2`? 
```rust
 if now >= genesis {
        // Only log the "Genesis reached" message if we are starting within 
        // a small 2-second window of the actual event.
	    if now <= genesis + 2 {
	        info!("Genesis reached! Starting services...");
	    }
	    break;
	}
```

PR #1134, opened 2026-01-13
feat: make pruning code more robust and add p2p ream tests allowing setting connection  topology

KolbyML:
### What was wrong?

Katya reported the prune code could crash the node so I worked on making the code more robust

### How was it fixed?

- add new p2p tests where you can specify connection topoloy 
- make it so node can be stopped if prune fails


Kayden-ML (approved):
Looks good 👍

PR #1141, opened 2026-01-14
refactor: switch duplicate to trace log

KolbyML:
### What was wrong?

Katya says the problem happens for other gossip messages not just block

### How was it fixed?

fixed it


shariqnaiyer (approved):

Kayden-ML (approved):
Looks good 👍

PR #1149, opened 2026-01-17
fix: checkpoint sync APIs

syjn99:
### What was wrong?

 

- `get_finalized_state_ssz` has duplicated logic with `get_state` except content type.
- `/health` endpoint wasn't registered.

### How was it fixed?

 

- Dedup state handler
- Register health endpoint

### Testing

Try those requests:

```bash
curl 127.0.0.1:5052/health | jq
```

```bash
curl 127.0.0.1:5052/lean/v0/states/head -H 'accept: application/octet-stream' > head.ssz
curl 127.0.0.1:5052/lean/v0/states/finalized -H 'accept: application/octet-stream' > finalized.ssz
```

### To-Do

 
### DIFF
--- a/crates/rpc/lean/src/handlers/state.rs
+++ b/crates/rpc/lean/src/handlers/state.rs
@@ -19,12 +19,6 @@ pub async fn get_state(
     state_id: Path<ID>,
     lean_chain: Data<LeanStoreReader>,
 ) -> Result<impl Responder, ApiError> {
-    let content_type = http_request
-        .headers()
-        .get(header::ACCEPT)
-        .map(ContentType::from)
-        .unwrap_or(ContentType::Json);
-
     let lean_chain = lean_chain.read().await;
 
     let block_root = match state_id.into_inner() {
@@ -83,7 +77,7 @@ pub async fn get_state(
         .map_err(|err| ApiError::InternalError(format!("DB error: {err}")))?
         .ok_or_else(|| ApiError::NotFound("Lean state not found".to_string()))?;
 
-    match content_type {
+    match ContentType::from(http_request.headers().get(header::ACCEPT)) {
         ContentType::Ssz => Ok(HttpResponse::Ok()
             .content_type(SSZ_CONTENT_TYPE)
             .body(state.as_ssz_bytes())),


crates/rpc/lean/src/handlers/state.rs
```diff
@@ -10,9 +12,19 @@ use ssz::Encode;
 // GET /lean/v0/states/{state_id}
 #[get("/states/{state_id}")]
 pub async fn get_state(
+    req: HttpRequest,
```
KolbyML:
```suggestion
    http_request: HttpRequest,
```

crates/rpc/lean/src/handlers/state.rs
```diff
@@ -10,9 +12,19 @@ use ssz::Encode;
...
     lean_chain: Data<LeanStoreReader>,
 ) -> Result<impl Responder, ApiError> {
+    // Determine response format based on Accept header.
+    // Default to JSON if not specified.
+    let accept_mime = req
+        .headers()
+        .get(header::ACCEPT)
+        .and_then(|v| v.to_str().ok())
+        .unwrap_or(mime::APPLICATION_JSON.as_ref());
+    let is_ssz = accept_mime.contains(mime::APPLICATION_OCTET_STREAM.as_ref());
```
KolbyML: the comment is somewhat redundent, and could you refactor/use https://github.com/ReamLabs/ream/blob/f4a4ad008defb0773e58e60f195d75a0a545636e/crates/common/validator/beacon/src/beacon_api_client/http_client.rs#L10-L23

We only support ssz/json, so it would be nice to use the type instead of doing it raw here, we can avoid using mime then

crates/rpc/lean/src/handlers/state.rs
```diff
@@ -64,38 +76,20 @@ pub async fn get_state(
...
+        return Ok(HttpResponse::Ok()
+            .content_type(mime::APPLICATION_OCTET_STREAM)
+            .body(state.as_ssz_bytes()));
+    }
+
     Ok(HttpResponse::Ok()
-        .content_type("application/octet-stream")
-        .body(state.as_ssz_bytes()))
+        .content_type(mime::APPLICATION_JSON)
+        .json(state))
```
KolbyML: we can just match here after the change is done

```rust
Ok(match content_type {
    Json => {etc}
    Ssz => {etc}
})
```

KolbyML (commented):
here is some feedback

syjn99:
@KolbyML Resolved the comments! I also made `ContentType` as common, so please take a look at whether this is desired one.

crates/common/api_types/common/src/content_type.rs
```diff
@@ -0,0 +1,36 @@
...
+            .map(|s| {
+                if s.contains(SSZ_CONTENT_TYPE) {
+                    ContentType::Ssz
+                } else {
+                    ContentType::Json
+                }
+            })
+            .unwrap_or(ContentType::Json)
+    }
+}
```
KolbyML:
```suggestion
impl From<Option<&ActixHeaderValue>> for ContentType {
    fn from(header_value: Option<&ActixHeaderValue>) -> Self {
        match header_value.and_then(|h| h.to_str().ok()) {
            Some(s) if s.split(';').any(|p| p.trim() == SSZ_CONTENT_TYPE) => ContentType::Ssz,
            _ => ContentType::Json,
        }
    }
}
```

What do you think of this? then we can simplify the code in the other file, remove the unwrap_or, also the SSZ_CONTENT_TYPE check is less loose then just contains

KolbyML (approved):
This is very nice <3

Kayden-ML:
Nice catch and improvement

PR #1155, opened 2026-01-19
feat(lean): add syncing + responding to BlocksByRoot requests

### LINKED ISSUE #741: Implement Lean ReqResp Initial Handshake

KolbyML:
### What was wrong?

fixes #741

Ream didn't have code to sync the chain, this PR adding syncing to Ream

### How was it fixed?

By adding a syncer, the LeanChainService can be in 2 states Synced and Syncing, if it is in Synced it performs as it did before

If it is in syncing, we have `JobQueue`'s which work on syncing `our HEAD` to the `networks HEAD`. Once a JobQueue collects enough blocks to execute a section, we spawn `ForwardBackgroundSyncer` which executes the chain forward and doesn't block the main `LeanChainService` from handling requests.

There are more details but I will leave that to the PR reviewer to figure out

I also implemented responding to BlockByRoot's requests


Kayden-ML (approved):
Really good work nice job, Looks great

PR #1157, opened 2026-01-20
test: enable p2p + finalization tests for devnet 2

KolbyML:
### What was wrong?

We were lacking signal if we were going in the right direction

### How was it fixed?

Enable the tests, they fail currently but should pass, which means something is wrong


Kayden-ML (approved):
Looks good 👍

PR #1168, opened 2026-01-22
ci: make docker ci upload builds for devnet1/devnet2

KolbyML:
### What was wrong?

We only had docker builds avaliable for devnet1

### How was it fixed?

making it so we build images for devnet1 and devnet2


Kayden-ML (approved):
Looks good 👍

PR #1171, opened 2026-01-22
fix: consider is_complete edgecase in replace_job_with_next_job syncer

KolbyML:
### What was wrong?

there still appears to be a different bug, who knows if it is in the syncer or devnet2 forkchoice

### How was it fixed?

consider is_complete case


Kayden-ML (approved):
Looks good 👍

PR #1187, opened 2026-01-26
refactor: use macro for logging skipped attestations

syjn99:
### What was wrong?

 

`process_attestations` is over 400+ LOC, let's make this smaller.

### How was it fixed?

 

Use a new macro for logging in attestations.

### To-Do

 
### DIFF
--- a/crates/common/consensus/lean/src/utils.rs
+++ b/crates/common/consensus/lean/src/utils.rs
@@ -2,6 +2,29 @@ use ream_post_quantum_crypto::leansig::public_key::PublicKey;
 
 use crate::validator::Validator;
 
+/// Macro to log skipped attestations with feature-specific identifier
+#[macro_export]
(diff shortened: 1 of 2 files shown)


crates/common/consensus/lean/src/utils.rs
```diff
@@ -2,6 +2,29 @@ use ream_post_quantum_crypto::leansig::public_key::PublicKey;
...
+        #[cfg(feature = "devnet2")]
+        info!(
+            reason = $reason,
+            source_slot = $attestation.source().slot,
+            target_slot = $attestation.target().slot,
+            "Skipping attestation: {:?}",
+            $attestation.aggregation_bits,
+        );
+    };
+}
```
KolbyML, line 26: this can be done without macros, hence I think it should be done without macro's. Macro's should only be used if their complexity is justified, which I don't see it in this case

Issue #1108, opened 2026-01-05
Add EF Tests for PeerDAS


shariqnaiyer:
### Describe the feature

https://github.com/paulmillr/eth-vectors/tree/main/peerdas/kzg

PR #1197, opened 2026-02-03
feat: Implement ef peerdas test macro

Kayden-ML:
### What was wrong?

Fixes: https://github.com/ReamLabs/ream/issues/1108

### How was it fixed?

I updated the make file to add the kzg files the same way it grabs the other ef tests then Implemented the macro, two of the functions pass, but the verify_cell_kzg_proof_batch() fails, and we still need to Implement the compute_cells() to Implement the remaining test case

image

image

image


KolbyML (approved):
:shipit: looks good

Issue #1205, opened 2026-02-04
Fix broken state ef-tests


shariqnaiyer:
### Describe the feature

Fix broken state ef-tests

<img width="685" height="139" alt="Image" src="https://github.com/user-attachments/assets/6c9f71a8-07a9-4a75-ace7-5e824c546077" />

PR #1211, opened 2026-02-06
fix: updated tests to fulu to fix the broken state tests

Patchoulis:
### What was wrong?

We were using the old electra tests which caused some of our tests to output as failing instead of the fulu tests. Fixes https://github.com/ReamLabs/ream/issues/1205

### How was it fixed?

I simply changed it to use the fulu folder.

### To-Do

 

Kayden-ML (approved):
Looks good 👍

PR #1215, opened 2026-02-10
test: add SSZ test vectors

unnawut:
### What was wrong?

Implement https://github.com/leanEthereum/leanSpec/pull/380

- Removed `TreeHash` derive from `SignedAttestation` because it's not used and it's in turn requiring `Signature` to implement `TreeHash` as well which requires custom logic but also not used.
- Change `Signature::inner` from `FixedBytes<SIGNATURE_SIZE>` to `LeanSigSignature` so that the final `Signature` scheme type is used directly without needing `FixedBytes` in-between
- Implemented Encode/Decode on Signature replacing the FixedBytes implementation
- Because leanSpec is giving us the expected outputs in json and camelCase format, we have to parse it to Ream/Rust types. So I created `passthrough_conversion`, `simple_conversion`, `nested_conversion`, `custom_conversion` to handle the json camelCase conversion to Ream types.

### To-Do

 
### DIFF
--- a/crates/crypto/post_quantum/src/leansig/signature.rs
+++ b/crates/crypto/post_quantum/src/leansig/signature.rs
@@ -1,15 +1,25 @@
 use leansig::{signature::SignatureScheme, MESSAGE_LENGTH};
 use serde::{Deserialize, Serialize};
-use ssz::{Decode, DecodeError, Encode};
+use ssz::{Decode, Encode};
+use ssz_derive::{Decode, Encode};
 
 use crate::leansig::{public_key::PublicKey, LeanSigScheme};
 
 /// The inner leansig signature type with built-in SSZ support.
 pub type LeanSigSignature = <LeanSigScheme as SignatureScheme>::Signature;
 
+const BLANK_SIGNATURE_SSZ_BYTES: [u8; 40] = [
+    36, 0, 0, 0, // offset_path = 36
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, // rho (28 zeros)
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, //
+    40, 0, 0, 0, // offset_hashes = 40
+    4, 0, 0, 0, // path: empty HashTreeOpening
+];
+
 /// Wrapper around leansig's signature type.
 /// Uses leansig's built-in SSZ encoding for interoperability with other clients.
-#[derive(Clone, Serialize, Deserialize)]
+#[derive(Clone, Serialize, Deserialize, Encode, Decode)]
+#[ssz(struct_behaviour = "transparent")]
 pub struct Signature {
     pub inner: LeanSigSignature,
 }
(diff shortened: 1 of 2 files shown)


crates/crypto/post_quantum/src/leansig/signature.rs
```diff
@@ -1,37 +1,73 @@
...
-type LeanSigSignature = <LeanSigScheme as SignatureScheme>::Signature;
-
-/// Wrapper around a fixed-size serialized hash-based signature.
-#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Encode, Decode, TreeHash, Copy)]
+/// Wrapper around leansig's signature type.
+/// Uses leansig's built-in SSZ encoding for interoperability with other clients.
+#[derive(Clone, Serialize, Deserialize)]
 pub struct Signature {
-    pub inner: FixedBytes<SIGNATURE_SIZE>,
+    pub inner: LeanSigSignature,
```
unnawut, line 24: Not sure if this is the right way to do it, but since `FixedBytes` is how Ream stores it internally and needs conversion to LeanSigSignature to do any operation anyway, so I'm thinking we could just remove this interim FixedBytes and use LeanSigSignature directly

KolbyML: Will LeanSigSignature encode to FixedBytes though is the question? I will read your code, but that would be the concern

unnawut: In case it's useful this is what I recall. I think FixedBytes was introduced to the spec in devnet0 just because leanSig wasn't stable yet so we just used `FixedBytes<SIGNATURE_SIZE>` as the mock Signature type in the specs.

And then in devnet1 we started with [`Bytes3116` in the specs](https://github.com/leanEthereum/leanSpec/blob/050fa4a18881d54d7dc07601fe59e34eb20b9630/src/lean_spec/subspecs/containers/signature.py) even with the actual signatures.

But towards end of devnet1, the `class Signature(Bytes3116)` was removed from the specs by https://github.com/leanEthereum/leanSpec/pull/210 and the canonical Signature type becomes this [`class Signature(Container)`](https://github.com/leanEthereum/leanSpec/blob/main/src/lean_spec/subspecs/xmss/containers.py#L50-L108).

So since then, leanSig and leanSpec uses the `Signature` container type that's variable length, and the fixed bytes is no longer used. And so that's why I'm thinking we could just use `LeanSigSignature` directly without conversion to fixed bytes. But I dunno maybe I'm missing something...

testing/lean-spec-tests/src/types/ssz_test.rs
```diff
@@ -0,0 +1,372 @@
...
+            #[allow(clippy::redundant_closure_call)]
+            fn try_from($val: &$json) -> anyhow::Result<Self> {
+                Ok(Self {
+                    $($field: $conv),+
+                })
+            }
+        }
+    };
+}
+
```
KolbyML: wouldn't the code just be easier to read without macro's? this kind of feels like we are abusing macros here without much benefits.

crates/crypto/post_quantum/src/leansig/signature.rs
```diff
@@ -1,40 +1,87 @@
...
+        ];
+        Self::from_ssz_bytes(&BYTES).expect("blank signature bytes are valid")
     }
 
     /// Create a mock signature for testing purposes.
+    ///
+    /// Note: This generates a real signature which is expensive. Prefer `blank()` when
+    /// you just need a placeholder signature.
     pub fn mock() -> Self {
         use rand::rng;
```
KolbyML, line 58: can you just remove this function as it isn't used

crates/crypto/post_quantum/src/leansig/signature.rs
```diff
@@ -1,40 +1,87 @@
...
-        Self::new(Default::default())
+        // 40 bytes: offset_path(4) + rho(28 zeros) + offset_hashes(4) + path(4)
+        const BYTES: [u8; 40] = [
+            36, 0, 0, 0, // offset_path = 36
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, // rho (28 zeros)
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, //
+            40, 0, 0, 0, // offset_hashes = 40
+            4, 0, 0, 0, // path: empty HashTreeOpening
+        ];
+        Self::from_ssz_bytes(&BYTES).expect("blank signature bytes are valid")
```
KolbyML: is this some universial constant? how did you derive it?

crates/crypto/post_quantum/src/leansig/signature.rs
```diff
@@ -1,40 +1,87 @@
...
-        }
+impl Decode for Signature {
+    fn is_ssz_fixed_len() -> bool {
+        <LeanSigSignature as Decode>::is_ssz_fixed_len()
+    }
+
+    fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, DecodeError> {
+        Ok(Self {
+            inner: LeanSigSignature::from_ssz_bytes(bytes)?,
+        })
```
KolbyML:
```suggestion
/// Wrapper around leansig's signature type.
/// Uses leansig's built-in SSZ encoding for interoperability with other clients.
#[derive(Clone, Serialize, Deserialize, Encode, Decode, PartialEq, Eq)]
#[ssz(struct_behaviour = "transparent")]
pub struct Signature {
    pub inner: LeanSigSignature,
}

impl std::fmt::Debug for Signature {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct("Signature")
            .field("inner", &"<LeanSigSignature>")
            .finish()
    }
}

```

We don't need to implement Encode, Decode, PartialEq ourselves, do we even need to implement Debug?

crates/crypto/post_quantum/src/leansig/signature.rs
KolbyML: Yeah I did some regression testing and this looks fine 👍

Issue #1219, opened 2026-02-11
store: more checkpoint checks and tests


Kayden-ML:
### Describe the feature

https://github.com/leanEthereum/leanSpec/pull/388

PR #1224, opened 2026-02-14
feat: added more checkpoint checks and tests

Kayden-ML:
### What was wrong?

Fixes: https://github.com/ReamLabs/ream/issues/1219

### How was it fixed?

Adds more tests, adds new checks to the validate_attestation() and updates the produce_block_with_signatures() to align with the python logic


KolbyML (approved):
:shipit: looks good

PR #1227, opened 2026-02-17
fix: remove manual re-gossip of attestations received from gossip

shariqnaiyer:
### What was wrong?

Removing manual re-gossip of attestations received from gossip. I suspect this was flooding our send pool:

This also makes our logs massive as attestations have proofs which are printed by libp2p-gossipsub.

image

### How was it fixed?

Remove need gossip from received attestations.

### To-Do

 

Kayden-ML (approved):
Looks good 👍

Issue #1145, opened 2026-01-15
Implement tests for checkpoint sync


Kayden-ML:
### Describe the feature

Implement tests for checkpoint sync 
https://github.com/leanEthereum/leanSpec/pull/279

PR #1236, opened 2026-02-21
feat: added tests for checkpoint_sync

Patchoulis:
### What was wrong?

fixes https://github.com/ReamLabs/ream/issues/1145. Essentially just adds the test within the spec for the checkpoint.

### How was it fixed?

On top of what was just in the spec, I added an extra assertion to ensure that the deserialization actually matches what we expect it to be.

### To-Do

 

crates/common/checkpoint_sync/lean/src/lib.rs
```diff
@@ -41,13 +41,120 @@ impl LeanCheckpointClient {
...
+            std::process::id(),
+            std::thread::current().id()
+        ));
+        std::fs::create_dir_all(&temp_path).expect("Failed to create temp directory");
+        let ream_db = ReamDB::new(temp_path).expect("Failed to init Ream Database");
+        let lean_db = ream_db.init_lean_db().expect("Failed to init lean db");
+
+        Store::get_forkchoice_store(signed_genesis_block, genesis_state, lean_db, Some(0))
+            .expect("Failed to create forkchoice store")
+    }
```
Kayden-ML, line 118:
```suggestion
    use ream_test_utils::store::sample_store;
``

Kayden-ML (changes requested):
Instead of redefining this function you should just be able to import this function

PR #1239, opened 2026-02-23
fix: handle case when tally is equal in common_highest_checkpoint

shariqnaiyer:
### What was wrong?

When comparing checkpoints from peers. We tally peers but in case of a tie, we should choose the checkpoint with the highest slot.

This should fix the late joiner test on Devnet3.

### How was it fixed?

```rust        
let mut highest_tally = 0;
        for (checkpoint, tally) in checkpoint_tally {
            if tally > highest_tally
                || (tally == highest_tally
                    && common_checkpoint
                        .is_some_and(|common_checkpoint| checkpoint.slot > common_checkpoint.slot))
            {
                highest_tally = tally;
                common_checkpoint = Some(checkpoint);
            }
        }

        common_checkpoint
```

I initially expanded the if statement which looked a little ugly to read and found a clean way to do the loop.

Tuples in Rust are compared left to right so it's equivalent to the above loop.

### To-Do

 
### DIFF
--- a/crates/networking/network_state/lean/src/lib.rs
+++ b/crates/networking/network_state/lean/src/lib.rs
@@ -82,8 +82,6 @@ impl NetworkState {
 
     pub fn common_highest_checkpoint(&self) -> Option<Checkpoint> {
         let peer_table = self.peer_table.lock();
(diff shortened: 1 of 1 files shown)


KolbyML:
I need to look more into this

PR #1243, opened 2026-02-25
feat: added checkpoint test implementation

Patchoulis:
### What was wrong?
fixes https://github.com/ReamLabs/ream/issues/1145. Essentially just adds the test within the spec for the checkpoint.

### How was it fixed?
On top of what was just in the spec, I added an extra assertion to ensure that the deserialization actually matches what we expect it to be.

### To-Do

 

KolbyML (approved):
:shipit: looks good

Issue #1242, opened 2026-02-24
Make tables for store fields that where introduced in devnet 3


Kayden-ML:
### Describe the feature

Make tables for store fields that where introduced in devnet 3

PR #1244, opened 2026-02-25
feat: Implement tables for devnet 3 store struct fields

Kayden-ML:
### What was wrong?

Fixes: https://github.com/ReamLabs/ream/issues/1242

### How was it fixed?

Implemented tables for the fields that where introduced on the store struct in devnet 3


KolbyML (approved):
:shipit: looks good

Issue #1231, opened 2026-02-20
Update is_aggregator logic for devnet 3


shariqnaiyer:
### Describe the feature

Update is_aggregator logic for devnet 3

TODO in service.rs

PR #1251, opened 2026-02-27
feat: allow setting is_aggregator from config

shariqnaiyer:
### What was wrong?

Fixes: #1231 
is_aggregator is set from config for all other clients

### How was it fixed?

Set is aggregator in node config instead of hardcoding to validator with id 1 as done previously to pass tests.

### To-Do

 

KolbyML (approved):
:shipit: looks good

Issue #1250, opened 2026-02-27
Add GET fork_choice JSON endpoint for lean


Patchoulis:
### Describe the feature

https://github.com/leanEthereum/leanSpec/pull/418

PR #1252, opened 2026-02-27
feat: added the fork_choice endpoint for lean

Patchoulis:
### What was wrong?

Fixes https://github.com/ReamLabs/ream/issues/1250

### How was it fixed?

Had to add a way to fetch all blocks. Additionally, since the endpoint required a devnet3 method, it's currently locked behind the devnet 3 feature flag.

### What's left to do?

Within the PR for the lean spec, it was mentioned that this was a change requested for Eth Lambda.
This PR is meant to help us eventually interop with Eth Lambda, with a good goal to work towards being to coordinate with them to get interoperability up and running.

Additionally, the next PR would be to add the lean spec tests associated with the endpoint.

### To-Do

 

crates/rpc/lean/src/handlers/fork_choice.rs
```diff
@@ -0,0 +1,88 @@
...
+        .iter()
+        .filter(|&block| block.message.block.slot >= finalized_slot)
+        .map(|block| {
+            let root = block.message.block.tree_hash_root();
+            let weight = weight_map
+                .as_ref()
+                .map(|weight_map| weight_map.get(&root).cloned().unwrap_or(0))
+                .unwrap_or(0);
+            serde_json::json!({
+                "root": format!("0x{}", encode(root)),
```
KolbyML:
```suggestion
                "root": root,
```


You can remove all the format's B256 serializes to 0xxyz properly

crates/storage/src/tables/lean/block.rs
```diff
@@ -155,4 +155,17 @@ impl LeanBlockTable {
...
+        let mut blocks = Vec::new();
+
+        for entry in table.iter()? {
+            let (_, block_entry) = entry?;
+            blocks.push(block_entry.value());
+        }
+
+        Ok(blocks)
+    }
 }
```
KolbyML: We are not putting all the blocks into memory delete this

Patchoulis: Hmm, the endpoint in general does require us to send info for all blocks above a certain slot, but I did modify it to have a filter for minimum slot to help improve the memory footprint. Let me know what you think.

PR #1291, opened 2026-03-12
test: re-add SSZ test vectors

unnawut:
### What was wrong?

Re-adding ssz test vectors after the https://github.com/leanEthereum/leanSpec/pull/435 fix in leanSpec. The vectors should now pass without any signature change in ream (pqinterop decided to update leanSpec to the format all clients are already using).

### How was it fixed?

Almost the same as https://github.com/ReamLabs/ream/pull/1215 but without the signature encoding changes

### To-Do

 
### DIFF
--- a/testing/lean-spec-tests/tests/tests.rs
+++ b/testing/lean-spec-tests/tests/tests.rs
@@ -42,9 +42,12 @@ fn test_all_state_transition_fixtures() {
         Ok(filter) => EnvFilter::builder().parse_lossy(filter),
         Err(_) => EnvFilter::new("info"),
     };
-    let _ = tracing_subscriber::fmt()
+    if let Err(err) = tracing_subscriber::fmt()
         .with_env_filter(env_filter)
-        .try_init();
+        .try_init()
+    {
+        warn!("Failed to initialize tracing subscriber: {err}");
+    }
 
     let fixtures = find_json_files("fixtures/devnet3/state_transition");
 
@@ -103,9 +106,12 @@ fn test_all_ssz_fixtures() {
         Ok(filter) => EnvFilter::builder().parse_lossy(filter),
         Err(_) => EnvFilter::new("info"),
     };
-    let _ = tracing_subscriber::fmt()
+    if let Err(err) = tracing_subscriber::fmt()
         .with_env_filter(env_filter)
-        .try_init();
+        .try_init()
+    {
+        warn!("Failed to initialize tracing subscriber: {err}");
+    }
 
     let fixtures = find_json_files("fixtures/devnet3/ssz/devnet");
 


testing/lean-spec-tests/src/types/ssz_test.rs
```diff
@@ -0,0 +1,519 @@
...
+impl TryFrom<&ValidatorJSON> for Validator {
+    type Error = anyhow::Error;
+
+    fn try_from(value: &ValidatorJSON) -> anyhow::Result<Self> {
+        Ok(Self {
+            public_key: {
+                let bytes = decode_hex(&value.pubkey)?;
+                if bytes.len() != 52 {
+                    return Err(anyhow!("Expected 52-byte pubkey, got {}", bytes.len()));
+                }
```
KolbyML:
```suggestion
                ensure!(bytes.len() == 52, "Expected 52-byte pubkey, got {}", bytes.len()));
         
```

We can use ensure here

testing/lean-spec-tests/src/types/ssz_test.rs
```diff
@@ -0,0 +1,515 @@
...
+        Ok(Self {
+            attestations: VariableList::try_from(
+                value
+                    .attestations
+                    .data
+                    .iter()
+                    .map(TryInto::try_into)
+                    .collect::<Result<Vec<_>, _>>()?,
+            )
+            .map_err(|e| anyhow!("{e}"))?,
```
KolbyML:
```suggestion
            .map_err(|err| anyhow!("Failed to xyz {err}"))?,
```
- we shouldn't just do a blind map of an error to anyhow, as it could make things harder to debug in the future
- we should use err to be consistent in the codebase

testing/lean-spec-tests/src/types/ssz_test.rs
```diff
@@ -0,0 +1,515 @@
...
+        Ok(Self {
+            config: (&value.config).try_into()?,
+            slot: value.slot,
+            latest_block_header: (&value.latest_block_header).try_into()?,
+            latest_justified: value.latest_justified,
+            latest_finalized: value.latest_finalized,
+            historical_block_hashes: VariableList::try_from(
+                value.historical_block_hashes.data.clone(),
+            )
+            .map_err(|e| anyhow!("{e}"))?,
```
KolbyML: apply this to all other places

testing/lean-spec-tests/tests/tests.rs
```diff
@@ -92,3 +95,70 @@ fn test_all_state_transition_fixtures() {
...
+#[test]
+fn test_all_ssz_fixtures() {
+    // Initialize tracing subscriber for test output
+    let env_filter = match env::var(EnvFilter::DEFAULT_ENV) {
+        Ok(filter) => EnvFilter::builder().parse_lossy(filter),
+        Err(_) => EnvFilter::new("info"),
+    };
+    let _ = tracing_subscriber::fmt()
+        .with_env_filter(env_filter)
+        .try_init();
```
KolbyML: I would do .expect or unwrap here, just to not hide the possible failure

testing/lean-spec-tests/tests/tests.rs
```diff
@@ -92,3 +95,70 @@ fn test_all_state_transition_fixtures() {
...
+                    total_tests += 1;
+                    info!("Starting test: {}", test_name);
+                    match run_ssz_test(test_name, test) {
+                        Ok(true) => {
+                            passed += 1;
+                            info!("PASSED: {}", test_name);
+                        }
+                        Ok(false) => {
+                            skipped += 1;
+                            info!("SKIPPED: {}", test_name);
```
KolbyML: inline these {test_name} etc

testing/lean-spec-tests/tests/tests.rs
unnawut: Already spoken in chat but typing down for reference:

expect/unwrap would panic when tracing is already setup by other tests. This is the error: 

> called `Result::unwrap()` on an `Err` value: SetGlobalDefaultError("a global default trace dispatcher has already been set")

testing/lean-spec-tests/tests/tests.rs
unnawut: copied over from the existing tests 😅 so I've updated those ones as well

unnawut:
There are some remaining clippy errors but it was from a previous PR so I'll go ahead just merge this one. Thanks for the reviews!

PR #1347, opened 2026-04-18
fix: suppress redundant log

Dsorken:
### What was wrong?

In devnet runs with 7 or less nodes the send queue full log will still show as per defined behavior in slow peer handling. There is another log added that is not spammed in the same way to indicate the same behavior that already exists.

### How was it fixed?

Suppress the redundant lib_p2p log.

### To-Do

 

KolbyML (approved):
:shipit: looks good

PR #1351, opened 2026-04-25
fix: use only "new" pool for safe target

Kayden-ML:
### How was it fixed?

https://github.com/leanEthereum/leanSpec/pull/680


KolbyML (approved):
:shipit: looks good

PR #1369, opened 2026-05-10
fix: disable default lean spec test features

KolbyML:
### What was wrong?

fix docker build

### How was it fixed?

do it


Kayden-ML (approved):
Looks good 👍

PR #1373, opened 2026-05-11
fix: run lean spec tests against lstar fixtures

KolbyML:
### What was wrong?

lean spec tests aren't running

### How was it fixed?

fix the issue


shariqnaiyer (approved):

Kayden-ML (approved):
Looks good 👍

PR #1374, opened 2026-05-11
feat: move --aggregate-subnet-ids assert into cli parser

KolbyML:
### What was wrong?

we should put as much validation into the cli parser as we can

### How was it fixed?

do it


Kayden-ML (approved):
Looks good 👍

PR #1379, opened 2026-05-12
fix: finalized endpoint + leanSpecTests ci

KolbyML:
### What was wrong?

Fix issues by the last PR

### How was it fixed?

fix them


Kayden-ML (approved):
Looks good 👍

PR #1381, opened 2026-05-13
fix: align lean blocks by range with spec

KolbyML:
### What was wrong?

We had some mistakes in our blocks by range implementation it wasn't aligned with the spec

### How was it fixed?

fix it


Kayden-ML (approved):
Looks Good 👍

PR #1390, opened 2026-05-20
feat: update lean multisig commit to latest 5eba3b1

Kayden-ML:
### What was wrong?

updates lean multisig commit to latest 5eba3b1


shariqnaiyer (approved):
Looks good 👍

PR #1420, opened 2026-06-09
chore: Updates lean multisig commit for devnet 5

Kayden-ML:
### How was it fixed?

updates lean multisig commit for devnet 5


shariqnaiyer:
There is a more recent commit addressed in a different PR.

PR #1425, opened 2026-06-12
chore: separating the test vectors for devnet4 and devnet5

Sahilgill24:
### What was wrong?

When running the Makefile commands for the test vectors in testing/lean-spec-tests, the same fixtures folder is used for devnet4 and devnet5, which would lead to re-downloading if we want to switch from devnet4 to devnet5 and back to devnet4 

### How was it fixed?

It was fixed by creating a split in the fixtures folder and adding the test vectors for both devnet4 and devnet5 in fixtures/{network} folders and changing the Makefile and some part accordingly. 

### To-Do


KolbyML (approved):
:shipit: looks good

PR #1433, opened 2026-06-16
fix: report fork_choice validator_count as null when head state missing

### LINKED ISSUE #1429: Report fork-choice validator_count as null (not 0) when the head state is missing in fork choice API
### Describe the feature

It's addressed in this PR.

https://github.com/leanEthereum/leanSpec/pull/1104

### Additional context

_No response_

zemse:
Fixes #1429.

When the node does not have the head state, the `/fork_choice` API used to return `validator_count: 0`. But `0` is also a real count, so a reader could not tell "no validators" apart from "state is missing".

Now it returns `validator_count: null` when the head state is missing. `null` means "unknown".

This matches leanSpec PR leanEthereum/leanSpec#1104. The null case is already covered by leanSpec's server test, and the happy path by the apitest conformance run.

### DIFF
--- a/crates/rpc/lean/src/handlers/fork_choice.rs
+++ b/crates/rpc/lean/src/handlers/fork_choice.rs
@@ -44,9 +44,7 @@ pub async fn get_fork_choice_tree(
         .get(head_root)
         .map_err(|err| ApiError::InternalError(format!("Unable to get head state: {err:?}")))?;
 
-    let validator_count = head_state
-        .map(|state| state.validators.len() as u64)
-        .unwrap_or(0);
+    let validator_count = head_state.map(|state| state.validators.len() as u64);
 
     let blocks = db
         .block_provider()


testing/lean-spec-tests/src/fork_choice.rs
zemse, line 1: fmt task was failing, if this fix should not be in this PR I can remove it

zemse: that fmt fix was done in recent commit on master hence removing the commit

zemse:
I've removed the tests since the checks are done in the lean api spec tests already.

Null path check: https://github.com/leanEthereum/leanSpec/blob/03e946a7e49db9d449a7e094b0071f3b4c372419/tests/node/api/test_server.py#L192

shariqnaiyer (approved):
Looks good!

Issue #1430, opened 2026-06-16
Clean up the aggregator API as Lean admin aggregator toggle is a no-op


shariqnaiyer:
### Describe the feature

Details can be found in this PR:
https://github.com/leanEthereum/leanSpec/pull/1104

perfogic:
Interested

shariqnaiyer:
@perfogic Assigned

perfogic:
TLDR for this `good first issue`:
### On `leanSpec` repository:
- Before the [leanSpec update](https://github.com/leanEthereum/leanSpec/pull/1104), the variable is stored seperate on `SyncService` and `NetworkService`.
```
POST /lean/v0/admin/aggregator
        |
        v
handle_toggle() -------> toggle is_aggregator field.
        |
        v
request.app["aggregator_controller"]
        |
        v
AggregatorController.set_enabled(enabled)
        |
        +------------------------------+
        |                              |
        v                              v
sync_service.is_aggregator     network_service.is_aggregator
```
So once `is_aggregator` is called from `AggregatorController`, it trigger both updates on `SyncService` and `NetworkService`.
Basicallly, the state is fragmented into two services. What to notice here is that `NetworkService` don't use 
- After the merge, `is_aggregator` is handled mainly on`SyncService`:
```
POST /lean/v0/admin/aggregator
        |
        v
ApiHandlers.aggregator_toggle()
        |
        v
ApiContext.require_aggregator_role_control()
        |
        v
sync_service.is_aggregator
```
`AggregatorController` is removed here.

perfogic:
The contribution is in PR #1442

PR #1434, opened 2026-06-17
fix: remove aggregator controller and share aggregator state wit…

perfogic:
### What was wrong?

Cleans up https://github.com/ReamLabs/ream/issues/1430.

- `POST /lean/v0/admin/aggregator` updated `NetworkState.is_aggregator`, but the actual aggregation paths in `LeanChainService` used their own boot-time `is_aggregator` value. 
=> This made the admin API misleading because toggling it did not affect chain-service aggregation behavior.

### How was it fixed?

- Deleted `AggregatorController` and stopped routing the admin endpoint through a separate controller layer.
- Removed `is_aggregator` from `NetworkState`; network state should not own the aggregator role.
- Added `AggregatorState` as the shared runtime flag for the node's aggregator role.
- Passed that same state into both the RPC handlers and `LeanChainService`.
- Changed the chain service to read the shared flag when deciding whether to aggregate during ticks or gossip attestation handling.
- Left prover setup and subnet subscriptions as boot-time behavior for now.

### To-Do


perfogic:
### Illustration about how this PR works
- Before this PR, the admin API and the chain service used different aggregator role states (`is_aggregator`).
```text
For Admin API:
  POST /lean/v0/admin/aggregator
          |
          v
  RPC aggregator handler
          |
          v
  AggregatorController.set_enabled(enabled)
          |
          v
  NetworkState.is_aggregator
          |
          v
  GET /lean/v0/admin/aggregator
 For Chain Service:
   config.is_aggregator
          |
          v
  LeanChainService.is_aggregator
          |
          +--> tick_interval(..., is_aggregator)
          |
          +--> on_gossip_attestation(..., is_aggregator)
```
=> This lead to the case that **aggregator role was enabled while the chain service still behaved as a non-aggregator.**
- After this PR, both RPC and chain-service aggregation checks share the same live role state, which is `AggregatorState`:
```
config.is_aggregator
        |
        v
AggregatorState
        |
        +------------------------------+
        |                              |
        v                              v
RPC admin aggregator handlers   LeanChainService
GET/POST /admin/aggregator      tick_interval(...)
                                on_gossip_attestation(...)
```

perfogic:
This PR is moved to #1442

Issue #1443, opened 2026-06-18
Enforce distinct-attestation-data cap in state transition


Kayden-ML:
### Describe the feature

The following leanSpec PR needs to be implemented in Ream, please give it a read understand it's effects and make a PR implementing the relevant portions in Ream:

https://github.com/leanEthereum/leanSpec/pull/1102/changes

shariqnaiyer:
Converting code from Python to Rust is trivial and is a waste of the core teams time. I am actively assigning these issues to 3rd party contributors so we can build a contributor friendly culture.

@Kayden-ML Please do the same. Do let me know if agree.

@Sahilgill24 @perfogic @tosynthegeek and some others have been quite eager to take these up.

Sahilgill24:
@Kayden-ML @shariqnaiyer  would love to take this one up if possible :) , would be helpful in getting more familiar with the ream codebase

Kayden-ML:
> [@Kayden-ML](https://github.com/Kayden-ML) [@shariqnaiyer](https://github.com/shariqnaiyer) would love to take this one up if possible :) , would be helpful in getting more familiar with the ream codebase

Assigned

shariqnaiyer:
@Sahilgill24 Please ensure you understand what you're implementing and can explain it's relevance in your EPF and Ream updates

PR #1446, opened 2026-06-18
fix: Enforcing distinct attestation data cap in state transition

Sahilgill24:
### What was wrong?

#1443 

### How was it fixed?

Earlier the distinct-data cap was being enforced in the fork choice in the `crates/common/fork_choice/lean/src/store.rs`, the bound check is now moved to the attestation processing step directly to state transition in `crates/common/consensus/lean/src/state.rs`. 

currently also implemented a unit test `process_attestations_rejects_over_distinct_data_cap` to verify the procedure, the test can be removed as the fixtures are updated. 

image

### Update

As the unit test was a duplicate, it has been removed 
### To-Do

 
### DIFF
--- a/crates/common/consensus/lean/src/state.rs
+++ b/crates/common/consensus/lean/src/state.rs
@@ -3,6 +3,7 @@ use std::collections::{HashMap, HashSet};
 use alloy_primitives::B256;
 use anyhow::{Context, anyhow, ensure};
 use itertools::Itertools;
+use ream_consensus_misc::constants::lean::MAX_ATTESTATIONS_DATA;
 use ream_metrics::{
     FINALIZED_SLOT, JUSTIFIED_SLOT, STATE_TRANSITION_ATTESTATIONS_PROCESSED_TOTAL,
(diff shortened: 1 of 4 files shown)


crates/common/consensus/lean/src/state.rs
```diff
@@ -278,6 +279,17 @@ impl LeanState {
     ) -> anyhow::Result<()> {
         let timer = start_timer(&STATE_TRANSITION_ATTESTATIONS_PROCESSING_TIME, &[]);
 
+        let mut distinct_data_roots = HashSet::new();
```
shariqnaiyer, line 282: do we need the explicit type set here

crates/common/consensus/lean/src/state.rs
```diff
@@ -1229,4 +1241,45 @@ mod test {
         assert!(result.is_err());
         assert!(result.unwrap_err().to_string().contains("state root"));
     }
+
+    #[test]
```
shariqnaiyer: feel free to remove the test please ensure it is covered in spec tests

shariqnaiyer (approved):

PR #1447, opened 2026-06-18
perf: reduce allocations in lean block import and head computation

zemse:
### What was wrong?

Two hot paths in the lean node allocated far more than necessary on every block and tick:

- `update_head` and `produce_block_with_signatures` iterated the entire
  `latest_known_aggregated_payloads` table and SSZ-decoded every `Vec<PayloadProof>`
  (the large post-quantum signatures) on each call. The only consumer reads just the
  validator id and attestation data root from each key, so the proof values were
  decoded and thrown away.
- `on_block` cloned the whole `SignedBlock`, carrying those same large signatures,
  just to hand it to the block store, which then cloned it again for its cache.

### How was it fixed?

- Added `iter_keys()` and `entry_count()` to the payloads table so callers read the
  keys and the entry count without decoding the proof values, and changed
  `extract_attestations_from_aggregated_payloads` to take `&[SignatureKey]`. The
  previous `is_empty()` skip is preserved by construction: both insert sites only ever
  store non-empty proof lists.
- Added `LeanBlockTable::insert_ref` so `on_block` passes the block by reference and
  redb serializes it in place; the owned `insert` now delegates to it.

No behavior change. Measured on a single-node devnet-4 (60s): `produce_block_with_signatures`
~22 MB -> ~40 KB per call, `update_head` ~15 MB -> ~4 MB per call, total process
allocation churn ~8.4 GB -> ~7.3 GB.

### To-Do

### DIFF
--- a/crates/common/fork_choice/lean/src/store.rs
+++ b/crates/common/fork_choice/lean/src/store.rs
@@ -380,12 +380,13 @@ impl Store {
         let latest_justified_root = latest_justified_provider.get()?.root;
 
         let attestations = {
-            let new_payloads = latest_new_aggregated_payloads_provider
+            let new_payload_keys = latest_new_aggregated_payloads_provider
                 .iter()?
                 .into_iter()
-                .collect();
+                .map(|(signature_key, _proofs)| signature_key)
+                .collect::<Vec<_>>();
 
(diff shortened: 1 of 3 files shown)


KolbyML:
Please rebase

KolbyML (approved):
:shipit: looks good

PR #1456, opened 2026-06-21
fix: correct field swap in recover_matrix (#1449)

### LINKED ISSUE #1449: Fix matrix reconstruction and complete matrix spec tests
### Describe the feature

Fix field swap bug in `matrix_entry` and complete two unit tests that would have caught it:

[Das core spec tests](https://github.com/ethereum/consensus-specs/blob/master/tests/core/pyspec/eth_consensus_specs/test/fulu/unittests/das/test_das.py)
[Polynomial commitments spec tests](https://github.com/ethereum/consensus-specs/tree/master/tests/core/pyspec/eth_consensus_specs/test/fulu/unittests/polynomial_commitments)

### Additional context

_No response_

tosynthegeek:
### What was wrong?

Fixes #1449 

`recover_matrix` was reconstructing `MatrixEntry` with `column_index` and `row_index` swapped

 

### How was it fixed?

Fix the swap and add the `test_compute_matrix` and `test_recover_matrix` unit tests from the spec's [das-core unittests](https://github.com/ethereum/consensus-specs/blob/master/tests/core/pyspec/eth_consensus_specs/test/fulu/unittests/das/test_das.py) that were missing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6281,6 +6281,7 @@ dependencies = [
  "ethereum_ssz",
  "ethereum_ssz_derive",
  "itertools 0.14.0",
+ "rand 0.9.2",
  "ream-bls",
  "ream-consensus-misc",
  "ream-execution-engine",
--- a/crates/common/consensus/beacon/src/matrix_entry.rs
+++ b/crates/common/consensus/beacon/src/matrix_entry.rs
@@ -8,7 +8,7 @@ use ssz_types::FixedVector;
 
 use crate::data_column_sidecar::Cell;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct MatrixEntry {
     cell: Cell,
     #[allow(dead_code)]
@@ -60,8 +60,8 @@ pub fn recover_matrix(
             matrix.push(MatrixEntry {
                 cell,
                 kzg_proof,
-                column_index: blob_index,
-                row_index: cell_index as u64,
+                column_index: cell_index as u64,
+                row_index: blob_index,
             });
         }
     }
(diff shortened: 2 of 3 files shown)


crates/common/consensus/beacon/src/matrix_entry.rs
```diff
@@ -60,8 +60,8 @@ pub fn recover_matrix(
             matrix.push(MatrixEntry {
                 cell,
                 kzg_proof,
-                column_index: blob_index,
-                row_index: cell_index as u64,
+                column_index: cell_index as u64,
+                row_index: blob_index,
```
perfogic, line 64: Good catch here, bro 💯 . `blob index` is the row index

Cargo.lock
```diff
@@ -6281,6 +6281,7 @@ dependencies = [
  "ethereum_ssz",
  "ethereum_ssz_derive",
  "itertools 0.14.0",
+ "rand 0.9.2",
```
perfogic, line 6284: this one i think you should move it to [dev-dependencies], because it is used in test only.
Follow the pattern of `Cargo.toml` in `fork_choice/lean`

Cargo.toml
```diff
@@ -57,14 +57,7 @@ exclude = ["book/cli"]
...
 edition = "2024"
-keywords = [
-    "ethereum",
-    "beam-chain",
-    "blockchain",
-    "consensus",
-    "protocol",
-    "ream",
-]
+keywords = ["ethereum", "beam-chain", "blockchain", "consensus", "protocol", "ream"]
```
perfogic: This one may be you should keep the format

perfogic (commented):
great work, man. I just drop some comments, hope it helps before the core team review 💯

shariqnaiyer (approved):

PR #1457, opened 2026-06-21
fix: hive client interop

rupam-04:
### What was wrong?

 
In the devnet5 lean client-interop tests, every mixed ream/ethlambda topology where ethlambda was the aggregator failed: the ream nodes panicked and died (RPC went dead and then connection refused on /lean/v0/fork_choice). The container logs revealed that non-aggregator ream nodes never initialized the type-2 aggregation bytecode, so they panicked (call init_aggregation_bytecode() first) the first time they had to verify an aggregate produced by the ethlambda aggregator. Startup only initialized the devnet4 bytecode, a different crate from the devnet5 one that verification actually reads — so for any non-aggregator the type-2 BYTECODE was never set. Aggregators masked the bug because they warm it lazily when producing aggregates while non-aggregators never aggregate.

### How was it fixed?

 
- `bin/ream/src/main.rs`: always initialize the aggregation verifier bytecode at startup for every node (not just aggregators), so any node can verify incoming aggregates. Feature-gated so devnet4 keeps its leanVM setup and devnet5 uses the type-2 setup; aggregators still additionally initialize the prover side.
- `crates/crypto/post_quantum/src/lean_multisig/type_2.rs`: added type_2_setup_verifier() (wraps the library's setup_verifier() / init_aggregation_bytecode()), the lightweight verify-only init called from main.rs for non-aggregators; skips the prover-only DFT twiddle precompute.
- `crates/common/fork_choice/lean/src/store.rs`: in the aggregator's recursive aggregation (state_aggregate), fold the selected child proofs into type_1_aggregate instead of passing an empty set.
- Propagates the error in some places instead of panicking. (Feel free to grill on this one, I'm not very sure if the newer one is better)

Ream, Ethlambda interop results:
Screenshot 2026-06-21 at 8 38 33 PM

### To-Do

 
### DIFF
--- a/bin/ream/src/main.rs
+++ b/bin/ream/src/main.rs
@@ -72,13 +72,13 @@ use ream_p2p::{
     network::lean::{LeanNetworkConfig, LeanNetworkService},
 };
 #[cfg(feature = "devnet4")]
-use ream_post_quantum_crypto::leansig::signature::Signature;
-#[cfg(feature = "devnet4")]
 use ream_post_quantum_crypto::lean_multisig::aggregate::{
     aggregation_setup_prover, aggregation_setup_verifier,
 };
 #[cfg(feature = "devnet5")]
 use ream_post_quantum_crypto::lean_multisig::type_2::{type_2_setup, type_2_setup_verifier};
+#[cfg(feature = "devnet4")]
+use ream_post_quantum_crypto::leansig::signature::Signature;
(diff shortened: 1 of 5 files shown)


shariqnaiyer:
This is very nice

shariqnaiyer:
I am going to test this against ethlambda shortly.

Kayden-ML (approved):
Looks good 👍

PR #1461, opened 2026-06-23
feat: implement validator custody requirement

### LINKED ISSUE #1460: Implement `get_validators_custody_requirement` from the Fulu spec
### Describe the feature

This needs to be implemented for DAS:

https://github.com/ethereum/consensus-specs/blob/master/specs/fulu/validator.md?plain=1#L106-L164

### Additional context

Currently, one Ream node stores all data columns. When we move toward full Fulu DAS compatibility, a node should only be required to custody the minimum number of custody groups (data columns) defined by the spec.

that minimum depends on the effective balance of the validators attached to the node, via `get_validators_custody_requirement`. 

Implementing this helper is needed before we can move from “store all columns” to spec-compliant validator custody.

perfogic:
### What was wrong?

Closes #1460.

Ream currently has custody group helpers, but it was missing the Fulu DAS helper for computing how many custody groups a node with attached validators must custody based on validator effective balance.

### How was it fixed?

Added `get_validators_custody_requirement` to the beacon custody group module. The helper sums the effective balances of the provided validator indices, computes the custody requirement using the Fulu network-spec values, and clamps the result between `validator_custody_requirement` and `number_of_custody_groups`.

Added unit tests for the minimum requirement, balance-based scaling, and max custody group cap.

### To-Do

### DIFF
--- a/crates/common/consensus/beacon/src/custody_group.rs
+++ b/crates/common/consensus/beacon/src/custody_group.rs
@@ -1,9 +1,45 @@
-use anyhow::{Ok, Result, ensure};
+use anyhow::{Ok, Result, anyhow, ensure};
 use discv5::enr::NodeId;
 use ream_consensus_misc::constants::beacon::NUM_CUSTODY_GROUPS;
+use ream_network_spec::networks::beacon::beacon_network_spec;
 use sha2::{Digest, Sha256};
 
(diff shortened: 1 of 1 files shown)


shariqnaiyer (approved):
Looks nice

Issue #1471, opened 2026-06-25
Embedded reth integration


Sahilgill24:
sub-issue for embedded reth integration with ream

PR #1472, opened 2026-06-26
feat: embedding reth directly into ream 

Sahilgill24:
### What was wrong?

#1471 

### Current Progress

- `RethHandle` is initialized, which contains the mpsc sender's to communicate with reth
-   The functions for communication b/w EL and CL are complete and tested. 

### To-Do

 

crates/common/reth/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+keywords.workspace = true
+license.workspace = true
+readme.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
+
+[dependencies]
+tokio.workspace = true
+reth-ethereum = { git = "https://github.com/paradigmxyz/reth.git", rev = "9384bc53d8c0c77e59cac83fdaaf3b372c6d2216", features = [
```
zemse:
```suggestion
reth-ethereum = { git = "https://github.com/paradigmxyz/reth.git", tag = "v2.3.0", features = [
```

crates/common/reth/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "reth-engine"
```
zemse: just to be consistent with other crates we can do either:
- `ream-reth-engine` crate with `reth-engine` dir
- `ream-reth` crate with `reth` dir

Sahilgill24: will do that, good suggestion !!

crates/common/reth/Cargo.toml
Sahilgill24: I kept it here to checkout a dependency check first for alloy, will be bumping it to a newer rev not related to a version

Sahilgill24:
Added the Fork Choice updated function and its related helper functions. 
Along with two tests 
- The reth node accepts tx's from the JSON-RPC method and the transaction is properly submitted in the pool 
- FCU returns the Valid status

Sahilgill24:
@zemse a few changes from my side 

- the `build()` function should only call resolve_kind, it does not need to call the `send_new_payload()`, we already have the payload id from the FCU updated. the send_new_payload() function would create a new Payload Id and cause a mismatch . 

- in the `build()` function , `built.try_into_v3()` would convert into `ExecutionPayloadV3` which has no slot_number field but as we are using slot_number field (which is in V4) it  would result in block_hash mismatch , we can just replace it with `into_execution_data()` and follow the V4 as in the main chain config. 

- Not something that will cause an error, but I don't think we need DB for even testing purposes, it's just extra import and build overhead, we can remove it ideally. What do you suggest for it or am I missing something related to the DB usage.

zemse:
> Not something that will cause an error, but I don't think we need DB for even testing purposes, it's just extra import and build overhead,

Yes we can remove the in memory db for the tests, if we can simply just remove the tests. Anyway those tests for reth are redundant since it is anyway tested on upstream, so as long as we are using their release tag we're good. I was going to remove them yesterday but I thought lets talk to you about it first.

crates/common/reth/src/payload.rs
```diff
@@ -1,81 +1,32 @@
...
+    payload_id: PayloadId,
+) -> eyre::Result<ExecutionData> {
+    // we can Ideally extract out WaitForPending to be more idiomatic
     let built = payload_builder
         .resolve_kind(payload_id, PayloadKind::WaitForPending)
         .await
         .ok_or_eyre("payload could not be resolved")??;
 
-    Ok(built.try_into_v3()?)
+    Ok(built.into_execution_data())
```
zemse: We will also need to include blob sidecars in the `build` fn for the CL to do DA

Sahilgill24: we could keep them, we just would have to include built.sidecars().clone(); in the build function and can return it along with the ExecutionData, but currently as we were doing the main ream/reth integration I thought let's remove them.

I think we should make these when we open that issue

Sahilgill24:
We can remove the tests, but I thought it was better if we could just test out the functions directly and capture the output as well to verify, I think we can remove them once we are complete with it (it's more of a personal choice, helps me better in debugging and all :)   )

.cargo/config.toml
```diff
@@ -2,4 +2,4 @@
 rustflags = ['--cfg', 'getrandom_backend="custom"']
 
 [target.x86_64-unknown-linux-gnu]
-rustflags = ['-C', 'target-cpu=x86-64-v3']
+rustflags = ['-C', 'target-cpu=x86-64-v3', '-C', 'target-feature=+pclmulqdq']
```
zemse, line 5: Just for info - this build flag is added here because rocksdb from reth requires it

KolbyML:
@Sahilgill24 @zemse please rebase this and I will look at it tomorrow

bin/ream/src/cli/lean_node.rs
```diff
@@ -102,6 +104,49 @@ pub struct LeanNodeConfig {
...
+    )]
+    pub reth_p2p_secret: Option<B256>,
+
+    #[arg(
+        long,
+        value_delimiter = ',',
+        help = "Comma-separated enode URLs of other execution layers to dial directly as a static trusted mesh (enode://<pubkey>@<ip>:<port>). Discovery stays off.",
+        requires = "reth_p2p_port"
+    )]
+    pub reth_trusted_peers: Vec<String>,
```
KolbyML: these should be behind a feature flag

bin/ream/src/main.rs
```diff
@@ -465,6 +468,49 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor, ream_
...
+    };
+
+    #[cfg(feature = "reth")]
+    chain_service.set_reth_handle(reth_handle).await;
+
+    #[cfg(feature = "reth")]
+    let reth_exit = &mut reth_node.node_exit_future;
+
+    #[cfg(not(feature = "reth"))]
+    let reth_exit = std::future::pending::<()>();
```
KolbyML: I think we  can scope this better, instead of spamming feature flags every line

crates/common/reth/src/fork_choice.rs
```diff
@@ -0,0 +1,95 @@
...
+    head_block_hash: B256,
+    safe_block_hash: B256,
+    finalized_block_hash: B256,
+) -> ForkchoiceState {
+    ForkchoiceState {
+        head_block_hash,
+        safe_block_hash,
+        finalized_block_hash,
+    }
+}
```
KolbyML: this function serves no purpose we should remove the function, and inline the logic

crates/common/reth/src/fork_choice.rs
```diff
@@ -0,0 +1,95 @@
...
+pub async fn update(
+    consensus_engine_handle: &ConsensusEngineHandle<EthEngineTypes>,
+    state: ForkchoiceState,
+    payload_attributes: Option<EthPayloadAttributes>,
+) -> eyre::Result<ForkchoiceUpdated> {
+    let updated = consensus_engine_handle
+        .fork_choice_updated(state, payload_attributes)
+        .await?;
+    Ok(updated)
+}
```
KolbyML: same here

crates/common/reth/src/fork_choice.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    create_payload_attributes(
+        genesis_time + slot * seconds_per_slot,
+        prev_randao,
+        Address::ZERO,
+        Some(parent_lean_block_root),
+        None,
+        None,
+    )
+}
```
KolbyML: same here

crates/common/reth/src/handle.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// Sends a fork choice update to the execution layer.
+    ///
+    /// Called with `payload_attributes` to start building a payload for a
+    /// proposal (the returned `payload_id` is fed to [`Self::build_payload`]),
+    /// and with `None` to canonicalize the head after a block is imported.
+    pub async fn update_forkchoice(
+        &self,
+        state: ForkchoiceState,
+        payload_attributes: Option<EthPayloadAttributes>,
+    ) -> eyre::Result<ForkchoiceUpdated> {
```
KolbyML: we use anyone in this project, not eyre

crates/common/reth/src/handle.rs
```diff
@@ -0,0 +1,229 @@
...
+    pub async fn build_payload(&self, payload_id: PayloadId) -> eyre::Result<ExecutionData> {
+        payload::build(&self.payload_builder, payload_id).await
+    }
+
+    pub async fn import_payload(
+        &self,
+        execution_data: ExecutionData,
+    ) -> eyre::Result<PayloadStatus> {
+        payload::import(&self.engine, execution_data).await
+    }
```
KolbyML, line 198: these functions don't deserve to exist

crates/common/reth/src/handle.rs
```diff
@@ -0,0 +1,229 @@
...
+        "terminalTotalDifficultyPassed": true,
+        "shanghaiTime": 0,
+        "cancunTime": 0,
+        "pragueTime": 0
+    }
+}
+"#;
+    let genesis: Genesis = serde_json::from_str(custom_genesis).expect("genesis failed");
+    Arc::new(genesis.into())
+}
```
KolbyML: ccustom chain? we should be more descriptive

crates/common/reth/src/handle.rs
```diff
@@ -0,0 +1,229 @@
...
+        "terminalTotalDifficulty": 0,
+        "terminalTotalDifficultyPassed": true,
+        "shanghaiTime": 0,
+        "cancunTime": 0,
+        "pragueTime": 0
+    }
+}
+"#;
+    let genesis: Genesis = serde_json::from_str(custom_genesis).expect("genesis failed");
+    Arc::new(genesis.into())
```
KolbyML: inline this

crates/common/reth/src/payload.rs
```diff
@@ -0,0 +1,159 @@
...
+///
+/// Returns a `PayloadStatus` reporting whether the EL accepted the payload
+/// (`Valid`, `Invalid`, or `Syncing`), so the caller must inspect it.
+pub async fn import(
+    consensus_engine_handle: &ConsensusEngineHandle<EthEngineTypes>,
+    execution_data: ExecutionData,
+) -> eyre::Result<PayloadStatus> {
+    let status = consensus_engine_handle.new_payload(execution_data).await?;
+    Ok(status)
+}
```
KolbyML: these functions don't descure to exist

crates/common/reth/src/payload.rs
```diff
@@ -0,0 +1,159 @@
...
+
+/// Convert reth's alloy `ExecutionData` into ream's `ExecutionPayload`.
+///
+/// The EL runs Prague, whose `ExecutionPayload` container is still the
+/// V3 shape — Prague's EIP-7685 execution requests are carried in the
+/// `newPayloadV4` sidecar as a separate parameter, not inside the payload body.
+/// So the payload is always an `ExecutionPayload::V3` here
+pub fn to_ream_execution_payload(data: &ExecutionData) -> eyre::Result<ReamExecutionPayload> {
```
KolbyML: anyhow

crates/common/reth/src/payload.rs
```diff
@@ -0,0 +1,159 @@
...
+    let ExecutionPayload::V3(v3) = &data.payload else {
+        return Err(eyre!(
+            "expected a V3 execution payload body (Cancun or later active in the EL genesis)"
+        ));
```
KolbyML: use bail!

KolbyML (commented):
Here is some feedback

KolbyML:
@zemse are you still working on resolving the concerns or did you forget to push?

zemse:
yes I have resolved most of them locally, couple of them are pending. I plan to do it tonight.

crates/common/reth/src/handle.rs
zemse: made some changes around here but kept these functions because I thought its worth hiding the `self.engine` which comes from reth lib. but if that does not look good I can change it

zemse:
please review @KolbyML

KolbyML:
@zemse ok fix the last conflict and I should be able to give the pr another look now

zemse:
ive fixed the conflicts @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1479, opened 2026-07-01
fix: close table transactions before return

Dsorken:
### What was wrong?

Currently there exists a race condition between backfill sync and the pruning of pending blocks where a select few functions are called and return an open transaction to a redb table instead of the passing along the data associated with the transaction. This under certain circumstances can lead to an attempt to read an old but open transaction during the sync process, which is blocked in redb causing an error in the library that results in a panic crashing Ream. 
Noted in Ream crash log:
image

### How was it fixed?

The solution is to ensure the transaction to a given redb table is closed through the use of inherent scope closure and just return the requested data.

### To-Do

 

KolbyML (approved):
:shipit: looks good

Issue #1408, opened 2026-06-05
Minimal PeerDAS Data Availability Client


shariqnaiyer:
### Describe the feature

This is the meta issue for tracking the EPF project: Minimal PeerDAS Data Availability Client

Data Availability is the guarantee that the data behind a transaction has been published somewhere any node can fetch and verify it. Currently there are only execution and consensus clients and DA is often bundled with a consensus client. We want to build a dedicated DA client that can plug into a minimal beacon client and a generic interface so it could plug into a post quantum consensus client in Rust 🦀.

This project builds a standalone ream da-node binary fully dedicated to the data layer: a Data layer client that custodies and serves the full 128 columns, and does the minimal amount of non-DAS work possible. No fork choice, no execution, no validator, just the data layer, decoupled from any one consensus runtime and runnable alongside a minimal consensus node.

tosynthegeek:
@shariqnaiyer I would like to work on this

Issue #1478, opened 2026-06-30
Full Fulu Compatibility and Cross-Client Interoperability


perfogic:
This issue is a bigger issue for creating sub issues to make `ream` fully compatible to **Fulu specification**.
It was described in this documentation: https://hackmd.io/rdPRuWyvSaaWAF6vO9r1hA.
and was topic of the cohort-seven: https://github.com/eth-protocol-fellows/cohort-seven/pull/225

## Overview

The implementation should be split into three phases so that we can migrate safely without changing too many consensus/networking paths at once.

-----
#### Tracking issues
- https://github.com/ReamLabs/ream/issues/1449
- https://github.com/ReamLabs/ream/issues/1454
- https://github.com/ReamLabs/ream/issues/1453
- https://github.com/ReamLabs/ream/issues/1460
- https://github.com/ReamLabs/ream/issues/1463

#### Status
- ✅: Done

-----
### Phase 2: Full-custody PeerDAS and DA decoupling (Kick-off)

Make Ream compatible with Fulu while every node behaves as a full-custody/supernode-style node.
This phase should focus on:
- Produce `DataColumnSidecar`s from proposed blocks.
- Broadcast data column sidecars after block production.
- Handle incoming `DataColumnSidecar` gossip.
- Validate and store data column sidecars.
- Decouple `is_data_available` from `on_block` by introducing a pending-availability flow.
- Re-check pending blocks when new data column sidecars arrive.
- Keep existing full-storage behavior for now to reduce migration risk.
- Add tests for spec-level functions and node-level propagation where needed. (First priority to make sure every issues work correctly,)

#### Definition of Done for the first phase

Phase 1 can be considered done when both local e2e coverage and devnet-level interop coverage are available.
- Add in-repo beacon e2e tests similar to the existing `ream lean_node` multi-node tests. These tests should verify beacon node startup, peer connectivity, checkpoint sync, late-joiner sync, block propagation, and eventually data-column propagation. - `Beacon E2E Test`
- Add Kurtosis/ethereum-package-based testing for interop coverage. This should verify that Ream can run in a real devnet-style environment and can interoperate with other Ethereum clients for both the beacon flow and the PeerDAS/data-column flow. - `Beacon Devnet Test`

After all implemented issues, the codebase has to pass both criteria. For early issues, it should pass only `Beacon E2E Test`, until the  `Beacon Devnet Test` is implemented - which was mentioned in https://github.com/ReamLabs/ream/issues/1445

[...]

PR #1485, opened 2026-07-01
feat(DA): Minimal PeerDAS Data Availability Client - DA client side

KimiWu123:
### What was wrong?

To address #1408, have a standalone DA client. 

This is an MVP of the minimal DA client, and I want to hear your feedback regarding the architecture before jumping into the implementation details.
**It's easier to review the PR commit by commit.**

#### Summary
The core of a minimum DA node. It runs as its own process beside the beacon node and does one thing: durably store a block's data columns and verify them, over a small local HTTP/RPC surface. The beacon keeps consensus logic; the DA node keeps the data.

**Command**: `cargo run -p ream -- -v 4 --ephemeral da_node`

#### Design philosophy
- Beacon owns consensus (gossip, fork choice, custody decisions); the DA node owns **storage** and self-contained **verification**. They talk over HTTP/RPC — the DA node has no P2P and no chain view.
- The core defines the domain and depends on neither beacon nor any commitment scheme; KZG plugs in through an adapter behind a trait, so the core stays reusable by a future scheme.
- Single-writer, load-shedding pipeline; loopback-only. One verification consumer is the sole writer, so "only verified data is stored" is structural and writes stay serialized; ingress is non-blocking and sheds load, so a slow verifier never blocks callers.
 
Crate layout
```mermaid

graph TB
    bin["bin/ream<br/>"]
    rpc["ream-rpc-da<br/>HTTP / RPC ingress"]
    node["ream-da-node<br/>(pipeline) ingest → verify → store"]
    kzg["da_adapters<br>(ream-da-verifier-kzg)<br/>KZG verifier"]
    core["ream-da<br/>core: traits + domain types"]
    ext["ream-consensus-beacon<br/>+ KZG libraries"]

    bin --> rpc
    bin --> node
    bin --> kzg
    rpc --> core
    rpc --> node
    node --> core
    kzg --> core
    kzg --> ext

    style core fill:#e8f5e9,stroke:#43a047
    style bin fill:#e3f2fd,stroke:#1e88e5
    style ext fill:#f5f5f5,stroke:#bbb
```
Arrows read "depends on". `ream-da` depends on nothing beacon- or crypto-specific; beacon types and KZG are reachable only through the adapter. The pipeline depends solely on the core's traits, so the concrete verifier is injected at bin/ream.

#### What's included
- **Store** — columns keyed by (block root, column), with availability reporting and finality-driven pruning.
- **Pipeline** — a bounded ingest queue feeding a single verify-and-store consumer, with backpressure and load-shedding.
- **KZG verification** — structure, the inclusion proof against the block header, and the cell proofs; trusted setup warmed up at startup.
- **RPC surface** — accept candidates and retention hints; serve availability and stored columns; a liveness probe. 

#### Next steps
- **File store → embedded database** — swap file-per-column for a transactional key-value DB (as Lighthouse stores DA data), for atomicity and cheaper range/prune.
- **SSZ wire format** — replace the dev-only hex-in-JSON envelope with SSZ, dropping the ~2× hex expansion and matching how columns are encoded on the wire.
- (TBD)**Verification workers / batching** — to replace current pipeline with a verification pool (and batch process) while keeping writes single-threaded. This is for performance consideration; need more investigation and data to decide.
- **Beacon integration** - need to discuss with @perfogic 

### To-Do

 

perfogic:
Thanks for putting this together and for mentioning me.
I just looked through the PR, and I think this is useful as an early DA-node scaffold for Phase 3 in this tracking issue:
https://github.com/ReamLabs/ream/issues/1478
I believe some of the types, storage/verifier abstractions, and RPC surface here can be reused later when we integrate this with the beacon-side `DataAvailabilityChecker` module, which will first decouple data availability validation from block handling.
I’ll take a closer look later this week after I finish the `DataAvailabilityChecker` issue, and then we can see which parts of this PR can help with the Phase 3 integration.

KimiWu123:
Hi @perfogic 
Quick thought on the `DataAvailabilityChecker` — might be worth deciding early where the Beacon/DA line sits.

If you just stick a thin interface in front of the availability + verification bits, anything that's really DA node's job can sit behind it — mock it for now, and later we swap the mock for the actual RPC call to the DA node. That way you don't have to touch any DA logic to keep going: it runs against the mock today, and in phase 3 we just plug in the real thing instead of reworking the checker.

Basically lets us build both sides in parallel, and makes phase 3 a wiring job instead of a rewrite.

perfogic:
Currently, the `DataAvailabilityChecker` is meant to handle the existing `is_data_available` issue described in the tracking issue, so it is definitely not something I want to mock out.

The plan is closer to the `lighthouse`-style flow: separate the DA availability logic internally inside the beacon node first, so `on_block` can keep blocks pending when data columns are not available yet and retry them when `DataColumnSidecar`s arrive.

Later, if we decide to integrate a standalone `da-node`, the backend behind the checker can be replaced with a DA-node RPC/provider interface. But the Phase 1 checker itself is still needed to fix the beacon-side pending availability flow.
You can keep working on your next steps for the DA-node side, and we can align the integration boundary later 👍. Thanks for your suggestion, btw

KimiWu123:
`is_data_available` seems like very DA-side work, and I’m worried we might be duplicating work. But, no worries, I'll keep working on the next steps.

perfogic:
Yeah, still the same as what we discussed in the common channel. Also, I just updated the Phase 3 description in the tracking issue (#1478) to add the boundary, that should answer the duplication concern.

crates/common/data_availability/src/id.rs
```diff
@@ -0,0 +1,56 @@
...
+        Ok(Self { block_root, index })
+    }
+
+    pub fn block_root(&self) -> B256 {
+        self.block_root
+    }
+
+    pub fn index(&self) -> u64 {
+        self.index
+    }
```
KolbyML, line 50: do we need these functions?

bin/ream/src/main.rs
```diff
@@ -106,6 +110,7 @@ use tracing_subscriber::EnvFilter;
 static ALLOC: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 
 pub const APP_NAME: &str = "ream";
+const DA_VERIFICATION_QUEUE_CAPACITY: usize = 256;
```
KolbyML: through the codebase can you write Data Availability instead of using the shorthand da

crates/common/da/src/column.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl DaPayload {
+    pub fn new(bytes: Vec<u8>) -> Self {
+        Self(bytes)
+    }
+
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0
+    }
+}
```
KolbyML: we can probably just use Vec<u8> directly as it doesn't look like this struct is doing anything

crates/rpc/da/src/handlers/column.rs
```diff
@@ -0,0 +1,97 @@
...
+/// format. The serving path's real consumer is the local beacon, which wants the
+/// bytes verbatim.
+/// Keep a hex branch (e.g. `?encoding=hex`) purely for `curl`/debug.
+#[derive(Serialize)]
+pub struct ColumnResponse {
+    block_root: B256,
+    index: u64,
+    slot: u64,
+    /// `0x`-prefixed hex of the opaque column payload.
+    payload: String,
```
KolbyML: We probably shouldn't be using a string here, we should be using serde derive on the real object

crates/rpc/da/src/handlers/mod.rs
```diff
@@ -0,0 +1,34 @@
+use alloy_primitives::B256;
+use ream_api_types_common::{error::ApiError, id::ID};
+
+pub mod availability;
+pub mod column;
+pub mod health;
+pub mod ingest;
+pub mod retention;
```
KolbyML: mods should go above imports

KolbyML (commented):
Can you remove any redundant functions added, and also only keep comments if they are justified, generally we prioritize readable code over comments. We generally should only have comments if the code is so complex it justifies it

crates/rpc/da/src/handlers/mod.rs
KimiWu123: addressed in [084c79b](https://github.com/ReamLabs/ream/pull/1485/commits/084c79bcd50800900fcc534bc12f94d36f3e82dc)

bin/ream/src/main.rs
KimiWu123: addressed in [28519ca](https://github.com/ReamLabs/ream/pull/1485/commits/28519cac8468110116c0ca60a7d93e4defca9d1a).

I'm wondering if any shorthand `da` is not allowed, or just for const? Many struct names also use `Da`, eg. `DaColumnId`, `DaNodeConfig`, `DaVerifier`... Do I use Data Availability instead?

crates/common/da/src/column.rs
KimiWu123: Fair point. I'll switch to Vec<u8>. If we later add payload validation (size caps) or move to refcounted bytes for cheap clones, I'd reintroduce a wrapper at that point. Addressed in [ab9f095](https://github.com/ReamLabs/ream/pull/1485/commits/ab9f09504e21141112fab1fcfdeba956bd3c7f65)

crates/rpc/da/src/handlers/column.rs
KimiWu123: Totally agree. 
This PR is a prototype of a minimal DA client, and I wanted to get feedback early. That’s why I initially implemented it using a hex string. I added a `TODO` above line 19 and also mentioned this in the PR description. This issue has already been addressed in a later PR.

crates/common/data_availability/src/id.rs
KimiWu123: I checked the range of `index` in `new()`. If we remove these two functions, then we have to make `block_root` and `index` public, and making the fields `pub` would open a path around that validation.

KimiWu123:
> Can you remove any redundant functions added, and also only keep comments if they are justified, generally we prioritize readable code over comments. We generally should only have comments if the code is so complex it justifies it

Hi @KolbyML , I have removed unnecessary comments and redundant functions. Also, all the feedback was fixed.

KolbyML:
@KimiWu123 is this PR ready for another review?

KimiWu123:
> @KimiWu123 is this PR ready for another review?

Yes, it's ready for review.

KolbyML:
@KimiWu123 can you remove all uses of the da shorthand

KimiWu123:
> the da shorthand

Hi @KolbyML fixed!

KolbyML:
ream-da = { path = "crates/common/da" }
ream-da-node = { path = "crates/common/da_node" }
ream-da-verifier-kzg = { path = "crates/common/da_adapters/kzg_verifier" }

@KimiWu123 fix this

KimiWu123:
> ream-da = { path = "crates/common/da" } ream-da-node = { path = "crates/common/da_node" } ream-da-verifier-kzg = { path = "crates/common/da_adapters/kzg_verifier" }
> 
> @KimiWu123 fix this

Sorry, I didn't realized the modules name. All the 'da' should be removed in the codebase now.

KolbyML (approved):
:shipit: looks good

Issue #1483, opened 2026-07-01
Decoupling Data Availability Validation and Handling pending DataColumnSidecar


perfogic:
## Description

Implement a `DataAvailabilityChecker`, which is described in https://hackmd.io/rdPRuWyvSaaWAF6vO9r1hA.

## Additional Context

No additional context

PR #1490, opened 2026-07-02
feat(data-availability): add pending data availability checker

perfogic:
### What was wrong?

Addresses #1483

The old beacon block import path checked data availability synchronously in `on_block`.
If required Fulu data column sidecars were not available at that moment, the block could be rejected instead of being kept pending and retried when the columns arrived.

### How was it fixed?

Added and wired the beacon-side `DataAvailabilityChecker` for the pending data availability flow.

This PR introduces:
- `PendingBlock`, `PendingAvailability`, and `DataAvailabilityChecker`.
- Beacon `Store` integration for tracking pending availability.
- `on_block` changes so blocks are executed first, then either imported if data is available or kept pending if data is still missing.
- `process_available_block`, which contains the import path after DA is satisfied.
- Data column sidecar handling that stores accepted columns and notifies the checker.
- DB backfill for columns that were already stored before a block is inserted into the checker.
- Parent-pending recovery, so child blocks wait if their parent is still pending and are retried after the parent imports.
- Sync retrieval for Fulu data column sidecars via `DataColumnSidecarsByRoot`.
- Unit tests for checker behavior.

### To-Do

- [ ] Add pruning/eviction for stale pending availability entries.
- [ ] Add focused e2e coverage for block/data-column propagation.
- [ ] Replace the current Fulu fork-epoch constant usage in sync retrieval with fork schedule/network-spec plumbing if needed.


perfogic:
Close since it will be open later on from [ream-collective](https://github.com/ream-collective/)

PR #1492, opened 2026-07-03
feat: add harness and connectivity e2e test with helpers to prevent oncelock for beacon node

### LINKED ISSUE #1480: Add beacon e2e test harness and smoke/connectivity tests
## Summary

Add an in-repo beacon e2e test harness, similar to the existing `ream lean_node` multi-node tests.
This is the first step before adding heavier beacon sync, validator, block production, and PeerDAS data-column e2e tests.

## Motivation

Phase 1 of Fulu/PeerDAS compatibility needs node-level testing. Consensus-spec tests are enough for spec/function-level changes, but they cannot verify runtime behavior such as node startup, peer connectivity, gossip wiring, checkpoint sync, or data-column propagation.

## Description

Implement beacon e2e tests in `bin/ream/src/main.rs` using `#[serial]`, so the existing CI test matrix can discover and run them.

Tests to add:

1. `test_beacon_node_runs_without_panicking`
   - start one beacon node with an isolated data directory
   - assert the node does not panic or exit early
   - assert the Beacon API is reachable
   - assert `/eth/v1/node/identity` works

2. `test_beacon_nodes_connect_with_bootnode`
   - start two beacon nodes with isolated data directories and ports
   - configure node 2 to use node 1 as a bootnode
   - assert both nodes keep running
   - assert node 2 connects to node 1
   - assert peer count becomes non-zero

## Additional Context

Kurtosis/ethereum-package testing should be added later as a separate devnet-level interop flow. 
This issue should only cover the lightweight in-repo beacon e2e foundation.

vuonghuuhung:
### What was wrong?

Fixes #1480.

Ream had no in-repo end-to-end coverage for the beacon node process itself: node startup, Beacon API reachability, and peer connectivity between two real beacon node processes.

### How was it fixed?

Added a beacon e2e tests: `test_beacon_node_runs_without_panicking` and `test_beacon_nodes_connect_with_bootnode`

`BEACON_NETWORK_SPEC`/`GENESIS_VALIDATORS_ROOT` are process-wide `OnceLock`s set once at startup, which normally blocks running more than one beacon node per test binary. `run_beacon_node` was split into a thin public wrapper (unchanged production behavior) and `run_beacon_node_inner(..., initialize_globals: bool)`; tests pass `false` and set those globals once per test binary via their own `Once` guards instead, letting multiple beacon nodes run in one process. Both tests reuse the existing Sepolia fixtures from `testing/gossip-validation`.

### To-Do


bin/ream/src/main.rs
```diff
@@ -1010,6 +1053,224 @@ mod tests {
...
+        )
+    }
+
+    fn beacon_fixture_assets_directory() -> PathBuf {
+        [
+            PathBuf::from("testing/gossip-validation/tests/assets/sepolia"),
+            PathBuf::from("../testing/gossip-validation/tests/assets/sepolia"),
+            PathBuf::from("../../testing/gossip-validation/tests/assets/sepolia"),
+        ]
+        .into_iter()
```
perfogic: This one look like a bruth force path, you can use like this:
```
PathBuf::from(env!("CARGO_MANIFEST_DIR"))
    .join("../../testing/gossip-validation/tests/assets/sepolia")
```

bin/ream/src/main.rs
```diff
@@ -497,9 +497,20 @@ pub async fn run_lean_node(config: LeanNodeConfig, executor: ReamExecutor, ream_
...
 pub async fn run_beacon_node(config: BeaconNodeConfig, executor: ReamExecutor, ream_db: ReamDB) {
+    run_beacon_node_with_global_init(config, executor, ream_db, true).await;
+}
+
+async fn run_beacon_node_with_global_init(
+    config: BeaconNodeConfig,
+    executor: ReamExecutor,
+    ream_db: ReamDB,
+    initialize_globals: bool,
+) {
```
perfogic, line 512: `run_beacon_node_with_global_init` feels like it always running with global init.
I know that you create a function with `initialize_globals` for test environment, it looks good to me.
Suggesting you change to

```rust
// add small comments above this function
async fn run_beacon_node_with_inner(
  config,
  executor,
  ream_db,
  initialize_globals: bool
)
```

perfogic (commented):
i think "what's wrong part" and "how was it fixed?" are a bit long, some duplicates with the description in issue

perfogic (approved):
A note here:
currently, we are adding e2e tests for beacon by taking advantage of `#[serial]` tag on `test_matrix.yml`, so it will be tested in both `test devnet4` and `test devnet5` jobs.

just wonder should we divide tag into `#[serial(beacon)] and #[serial(lean)]` or only `#[serial]` is enough ? - cause this will introduce some changes in the ci. @shariqnaiyer 

Image

bin/ream/src/main.rs
```diff
@@ -947,6 +981,16 @@ mod tests {
         preseed_node_3_before_checkpoint_sync: bool,
     }
 
+    // Production's `BEACON_NETWORK_SPEC`/`GENESIS_VALIDATORS_ROOT` OnceLocks only allow one
+    // set per process, so `run_beacon_node_for_test` skips them and these `Once` guards set
+    // them once per test binary instead, shared by every beacon e2e test below.
+    //
+    // Only safe because all beacon e2e tests here use Sepolia and the same Sepolia fixture.
+    // `call_once` silently no-ops after the first call, so a future test needing a different
+    // network or genesis root would silently reuse the first test's values instead of failing.
```
shariqnaiyer: we can remove these comments

bin/ream/src/main.rs
```diff
@@ -1010,6 +1054,218 @@ mod tests {
...
+            socket_port = config.socket_port,
+            discovery_port = config.discovery_port,
+        );
+        tokio::spawn(
+            async move {
+                run_beacon_node_for_test(config, executor, db).await;
+            }
+            .instrument(span),
+        )
+    }
```
shariqnaiyer, line 1072: I don't knwo if we want to have code for spawning test nodes in the main.rs

shariqnaiyer: I just remembered we have a lot of node tests in main.rs, ignore this, I shall look into this further today.

shariqnaiyer (approved):
Looks good

PR #1493, opened 2026-07-03
feat: Implements the build the block one interval early optimization 

### LINKED ISSUE #1491: Implement build the block one interval early optimization into ream
### Describe the feature

Image

### Additional context

_No response_

Kayden-ML:
### What was wrong?

Fixes: https://github.com/ReamLabs/ream/issues/1491

### How was it fixed?

Implements the build the block one interval early optimization by building blocks one interval early in a background thread so they are ready to gossip instantly at the start of a slot

I tested the pr by running a devnet using lean start on the master branch and then comparing my local branch with the change and it finalizes just as good as the master branch and reaches finalization over 180 slots

image

### DIFF
--- a/crates/common/validator/lean/src/service.rs
+++ b/crates/common/validator/lean/src/service.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::{collections::HashSet, sync::Arc};
 
 use anyhow::anyhow;
 use ream_chain_lean::{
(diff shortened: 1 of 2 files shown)


crates/common/validator/lean/src/service.rs
```diff
@@ -75,61 +85,39 @@ impl ValidatorService {
...
+                        3 => {
+                            let slot = slot + 1;
+                            // First tick (t=0): Propose a block.
+                            if slot > 0 && let Some(keystore) = self.is_proposer(slot) {
+                                info!(slot, tick = tick_count, "Building block by Validator {}", keystore.index);
+                                let chain_sender = self.chain_sender.clone();
+                                let keystore = keystore.clone();
+                                self.block = Some(tokio::spawn(async move {build_block(chain_sender, slot, keystore).await}));
+                            }
+                        }
```
shariqnaiyer, line 97: I have a feeling we want to build it on the 5th interval

shariqnaiyer (approved):

crates/common/validator/lean/src/service.rs
shariqnaiyer: We most certainly want to start building in the the 5th interval not the 4th

PR #1494, opened 2026-07-03
build: pin the host binary to x86-64-v3

shariqnaiyer:
Our current build was too slow at WHIR/XMSS signature aggregation. This optimization was shared in the PQ interop call by some of the teams I think 6 months ago. 

### How was it fixed?

 

Build the native (non-zkvm) binary for the x86-64-v3 microarchitecture level (AVX2 + FMA + BMI2). This is a large speedup for the WHIR/XMSS signature aggregation hot path.
### To-Do

 

KolbyML (approved):
:shipit: looks good

Issue #1178, opened 2026-01-23
test Ream against Qlean


Kayden-ML:
### Describe the feature

test Ream against Qlean

PR #1496, opened 2026-07-05
fix(lean): reject malformed justification state with typed errors

perfogic:
### What was wrong?

Address leanSpec [#1178](https://github.com/leanEthereum/leanSpec/pull/1178)

**TLDR**: 
- `is_justifiable_after` is now total: below-finalized slots return `false` instead of erroring, and the fixture runner supports signed deltas for those cases.
- `process_attestations` now rejects malformed justification bookkeeping before unpacking, including vote-list length mismatch and zero-hash roots.

### How was it fixed?

- Made `is_justifiable_after` total by returning `false` for slots before finalization.
- Updated the justifiability runner to support signed deltas from leanSpec 1178.
- Added malformed justification-state guards for vote-list length mismatch and zero-hash roots.
- Added direct tests for below-finalized slots and attestation senarios.

### To-Do

### DIFF
--- a/crates/common/consensus/lean/src/slot.rs
+++ b/crates/common/consensus/lean/src/slot.rs
@@ -1,14 +1,14 @@
-use anyhow::ensure;
+pub fn is_justifiable_after(candidate_slot: u64, finalized_slot: u64) -> bool {
+    if candidate_slot < finalized_slot {
+        return false;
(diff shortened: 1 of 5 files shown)


KolbyML (approved):
:shipit: looks good

PR #1499, opened 2026-07-06
fix: finalization bug for tiered attestation scoring

Kayden-ML:
### What was wrong?

fix's dropping attestation payloads early targeting the same slot

before:
image

after the fix 

image

### DIFF
--- a/crates/common/fork_choice/lean/src/store.rs
+++ b/crates/common/fork_choice/lean/src/store.rs
@@ -1275,16 +1275,6 @@ impl Store {
                         continue;
                     }
 
-                    if is_projected_slot_justified(
-                        &justified_slots,
-                        finalized_slot,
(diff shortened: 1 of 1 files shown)


crates/common/fork_choice/lean/src/store.rs
```diff
@@ -1280,7 +1280,9 @@ impl Store {
                         finalized_slot,
                         candidate_data.target.slot,
                     ) {
-                        continue;
+                        if processed_data_roots.contains(data_root) {
+                            continue;
+                        }
```
shariqnaiyer: I think we remove the entire if condition which will make clippy pass

shariqnaiyer (approved):

shariqnaiyer (approved):

PR #1501, opened 2026-07-06
fix(lean): bound block slot before the state transition in on_block

zemse:
### What was wrong?

 

Based on https://github.com/leanEthereum/leanSpec/pull/1182

Proposer for a  block is `slot % num_validators`. So a malicious proposer can sign a valid lean block at high slot number where `block.slot % num_validators == validator_index`.

The `on_block` has no upper bound on the `block.slot` and when such a malicious block is imported, the following while loop runs for a huge number of iterations.

https://github.com/ReamLabs/ream/blob/c30a9c952b378b0a46d2a339e9584f243c21b18b/crates/common/consensus/lean/src/state.rs#L118

While it rejects later in `process_block_header` however honest validator has already burned CPU on high number of iterations due to the malicious block.

### How was it fixed?

 

Two guards are added in the `on_block` so that such a malicious block is rejected before it gets into `state_transition` call.

- Parent-gap cap: reject `block.slot − parent.slot > MAX_HISTORICAL_BLOCK_HASHES` i.e. `2^18`.
- Clock horizon: reject `block.slot > current_slot + 1`, keeping far-future blocks out of the store.

### To-Do

 
### DIFF
--- a/crates/common/fork_choice/lean/src/store.rs
+++ b/crates/common/fork_choice/lean/src/store.rs
@@ -24,7 +24,7 @@ use ream_consensus_lean::{
 };
 use ream_consensus_misc::constants::lean::{
     GOSSIP_DISPARITY_INTERVALS, INTERVALS_PER_SLOT, MAX_ATTESTATIONS_DATA,
-    attestation_committee_count,
+    MAX_HISTORICAL_BLOCK_HASHES, attestation_committee_count,
 };
 use ream_metrics::{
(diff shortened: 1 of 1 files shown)


Kayden-ML:
@zemse can you rebase your pr

zemse:
Just rebased over latest master

Kayden-ML (approved):
Looks good 👍

PR #1507, opened 2026-07-07
 fix: align fork-choice attestation target selection with head checkpoint

shariqnaiyer:
### What was wrong?

Fork-choice attestation target selection and block-attestation packing judged justifiability against the **store's** `latest_finalized_provider` /
`latest_justified_provider`. Those are monotonic maxima over every processed block's parent-state checkpoints, so they can **exceed** the head chain's real finalized/justified (observed: store justified = 65 while head-state justified = 20).

The proposer's state transition (`process_attestations`) judges votes against the **head state's** `latest_finalized` / `latest_justified`. So attesters produced targets that looked valid under the higher store values but were rejected at block processing as "Target slot not justifiable" votes were silently dropped and **justification/finalization froze** (finalized plateaus while justified climbs).

### How was it fixed?

Reference the head state's checkpoints everywhere target selection and attestation packing happen, so attesters, the block builder, and block processing agree on the justifiable-slot set.

### To-Do

### DIFF
--- a/crates/common/fork_choice/lean/src/store.rs
+++ b/crates/common/fork_choice/lean/src/store.rs
@@ -568,30 +568,23 @@ impl Store {
     }
 
     pub async fn get_attestation_target(&self) -> anyhow::Result<Checkpoint> {
-        let (
-            head_provider,
-            block_provider,
-            safe_target_provider,
-            latest_justified_provider,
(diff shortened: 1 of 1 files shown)


Kayden-ML (approved):
Looks good 👍

PR #1508, opened 2026-07-07
feat: move aggregation off the hot path

shariqnaiyer:
### What was wrong?
The expensive committee-signature aggregation and block signature verification ran inline on the consensus loop while holding the store lock. Each slot the loop blocked for the full proof time, so block import, attestation handling, and safe_target updates all froze. safe_target fell behind head and finalization plateaued. 

### How was it fixed?
Move the heavy work off the loop and off the store lock, and bound it:
- Off-loop committee aggregation
- Off-lock block-signature verification
- Bound aggregation to recent slots + prove recent-first
- Evict superseded known payloads

### To-Do
- [z] I have read [[CONTRIBUTING.md](https://github.com/ReamLabs/ream/blob/master/CONTRIBUTING.md)](https://github.com/ReamLabs/ream/blob/master/CONTRIBUTING.md).
- [x] This PR title follows [[conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)](https://www.conventionalcommits.org/en/v1.0.0/).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6238,6 +6238,7 @@ dependencies = [
  "ethereum_ssz",
  "ethereum_ssz_derive",
  "itertools 0.14.0",
+ "rayon",
  "ream-consensus-misc",
  "ream-metrics",
  "ream-post-quantum-crypto",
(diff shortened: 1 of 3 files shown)


Kayden-ML (approved):
Looks good 👍 nice work

PR #1509, opened 2026-07-08
fix: calls build block in the 5th interval instead of 4th

Kayden-ML:
### How was it fixed?

calls build block in the 5th interval instead of 4th


KolbyML (approved):
:shipit: looks good

PR #1513, opened 2026-07-08
fix: backfill race condition and lean-spec-tests in hive

Sahilgill24:
### What was wrong?

All the changes mentioned in the [Hive ream tests doc](https://hackmd.io/@Ayhm2-FHQhSLkoc1OEcr9g/Bkoz3IwXzl)

### How was it fixed?

- adding the check for preventing the race condition, if the only peer available is the one in Reset rather than earlier going into an infinite loop and get stuck in syncing, try to reuse it if it is the only one available rather than hard rejection on the peer
- use the MOCK_PROOF_PREFIX from the leanSpec for mock proof aggregation in the test driver
- adding proper alias for relevant exception fields and tick related fields. 

All 117 fork choice spec tests and all the state transition tests are now passing, earlier about 39 of these in total were failing on hive. 

[Fork Choice]
image

[State Transition]
image

### To-Do

 
### DIFF
--- a/crates/common/chain/lean/src/service.rs
+++ b/crates/common/chain/lean/src/service.rs
@@ -1003,7 +1003,29 @@ impl LeanChainService {
             );
         }
 
-        self.choose_weighted_peer(&fallback_candidates)
+        if let Some(peer_id) = self.choose_weighted_peer(&fallback_candidates) {
+            return Some(peer_id);
+        }
+
+        // Last resort: reuse the avoided peer only when it is genuinely the sole peer we are
+        // connected to. A `Reset` job avoids the peer that just failed it, but if that peer
+        // is now the only connected candidate (e.g. after a pause collapses every other
+        // connection during recovery), refusing it here deadlocks backfill forever. 
+        if avoid_peer_id.is_some()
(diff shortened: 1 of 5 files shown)


shariqnaiyer:
@Sahilgill24 Could you please ensure cargo/udeps ci passes. I am reviewing the PR in the meantime.

Sahilgill24:
@shariqnaiyer , as mentioned in the CI of cargo-udeps it is a false positive, because these crates are being used under the devnet5 feature flag and the default for this crate is devnet-4, so that is what is causing the issue I believe

shariqnaiyer:
> @shariqnaiyer , as mentioned in the CI of cargo-udeps it is a false positive, because these crates are being used under the devnet5 feature flag and the default for this crate is devnet-4, so that is what is causing the issue I believe

I believe there is certainly a way to make the ci green. I would rebase the PR as the devnet4 removal is merged and give it another shot. I can take a look if it's still causing problems when I wake up.

Sahilgill24:
@shariqnaiyer , I rebased the PR and the cargo-udeps check is also fixed now, please have a look

KolbyML (approved):
:shipit: looks good

PR #1514, opened 2026-07-08
fix(lean): accept rejectionReason in lean state-transition test driver

perfogic:
### What was wrong?

image

Currently see that we have two error on two test on `lean-spec-tests-state-transition`:
image

=> `test_driver.rs` didn't handle rejectionReason (the devnet5 field for expected failures), so those fixtures were treated as successes.

### How was it fixed?

- Add `rejection_reason` to `StateTransitionRunRequest` in `test_driver.rs` and inside `run_state_transition` func.

### Notes:

- The `rejection_reason` is already added to our `StateTransitionTest` on `testing/lean-spec-tests/src/types/state_transition.rs`, that's why when we run:
```
cargo +stable test -p lean-spec-tests --no-default-features --features devnet5 \
  --test tests test_all_state_transition_fixtures -- --nocapture 
```
It passed all without above fix.

- I have re-produced on local and it worked it successfully.
   
### To-Do

 
### DIFF
--- a/crates/rpc/lean/src/handlers/test_driver.rs
+++ b/crates/rpc/lean/src/handlers/test_driver.rs
@@ -77,8 +77,12 @@ pub struct GenesisParams {
 pub struct StateTransitionRunRequest {
     pre: FixtureState,
     blocks: Vec<FixtureBlock>,
+    /// devnet4 fixtures name it `expectException`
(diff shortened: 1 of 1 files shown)


perfogic:
Closed since it's addressed in 1513

PR #1515, opened 2026-07-08
feat(beacon): add beacon e2e block production convergence test

### LINKED ISSUE #1481: Add block production and checkpoint convergence test
## Summary

Add a beacon e2e test that verifies Ream beacon nodes can produce/import blocks, propagate them to peers, and advance checkpoints.

This should build on the beacon e2e harness from the first issue.

## Motivation

After beacon node startup and peer connectivity work, the next critical milestone is proving that the beacon runtime actually works: blocks are produced, imported by peers, and fork-choice/checkpoints advance.

Unlike `lean_node`, the beacon node does not embed validator duties. This test should therefore exercise the real beacon architecture by running `beacon_node` together with `validator_node`, plus an execution backend or mock execution setup.

## Description

Implement the following test in `bin/ream/src/main.rs` using `#[serial]`:

1. `test_beacon_nodes_produce_blocks_and_converge`
   - start at least four beacon nodes with isolated data directories and ports
   - start one or more validator nodes connected to the beacon API
   - provide validator keystores and a compatible local/dev network configuration
   - provide an execution backend, mock execution engine, or documented temporary test backend
   - produce/import blocks across multiple slots
   - assert the proposer/local node head slot advances
   - assert peer nodes receive and import produced blocks
   - assert nodes converge to the same or similar head
   - assert justified/finalized checkpoints advance when enough slots/epochs have passed

## Additional Context

This test may require execution/mock-execution setup. If it is too heavy or unstable for regular PR CI at first, gate it behind an environment variable or mark it ignored until the setup is reliable.

This issue should focus on proving block production and convergence first. PeerDAS/data-column propagation should be covered separately in #1482.

perfogic:
### What was wrong?

Closes https://github.com/ReamLabs/ream/issues/1481

### How was it fixed?

- Test `test_beacon_nodes_produce_blocks_and_converge` is added to test with 2 beacon nodes, 2 validators node with 8 validators and finalized successfully at justifed_epoch=3 and finalized_epoch = 2.
**Note**: This one contains lot of changes which i didn't want to but had to. So please take time to check the link below:
- Fix all issues which are documented in https://hackmd.io/@perfogic/r1w-3oR7ze

### Result:
- Run with 2 beacon nodes and 2 validator nodes with 8 validators, every nodes is finalized:
image
image

- For running 4 beacon, validator nodes with 16 validators, i can not make it work in the PR right now, since half of peers will be fell behind the others. => This can be a fix for next PR. The result produced on local:
image

### To-Do

 
### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5984,6 +5984,7 @@ dependencies = [
  "ream-account-manager",
  "ream-api-types-beacon",
  "ream-api-types-common",
+ "ream-bls",
  "ream-chain-beacon",
  "ream-chain-lean",
  "ream-checkpoint-sync-beacon",
@@ -5999,6 +6000,7 @@ dependencies = [
  "ream-fork-choice-lean",
  "ream-keystore",
  "ream-metrics",
+ "ream-mock-execution-engine",
  "ream-network-manager",
  "ream-network-spec",
  "ream-network-state-lean",
@@ -6019,7 +6021,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "serial_test",
- "snap",
+ "sha2",
  "ssz_types",
  "tikv-jemallocator",
  "tokio",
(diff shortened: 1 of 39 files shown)


perfogic:
Closed since it is replaced by #1529

PR #1516, opened 2026-07-08
fix: prevents the prune_old_state() from deleting unfinalized history

Kayden-ML:
### What was wrong?

image

### How was it fixed?
Moves pruning behind finalization so the unfinalized active state history is not deleted.

Uses latest_finalized_slot instead of head_slot because using the latest_finalized_slot prevents the prune_old_state() function from pruning the state before the network has officially finalized it. Before, using head_slot caused the pruner to target unfinalized moving blocks which deleted active history and triggered the deeper than 128 slot reorg crash mega had mentioned

### DIFF
--- a/crates/common/chain/lean/src/service.rs
+++ b/crates/common/chain/lean/src/service.rs
@@ -2882,7 +2882,7 @@ impl LeanChainService {
     }
 
     async fn prune_old_state(&self, tick_count: u64) -> anyhow::Result<()> {
-        let (head, block_provider, slot_index_provider, state_provider) = {
+        let (slot_index_provider, state_provider, latest_finalized_slot) = {
             let fork_choice = self.store.read().await;
             let store = fork_choice.store.lock().await;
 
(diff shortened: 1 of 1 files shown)


crates/common/chain/lean/src/service.rs
```diff
@@ -2900,21 +2892,14 @@ impl LeanChainService {
...
-        let head_slot = block_provider
-            .get(head)?
-            .ok_or_else(|| anyhow!("State not found for head: {head}"))?
-            .block
-            .slot;
-
-        if head_slot > STATE_RETENTION_SLOTS {
-            let prune_target_slot = head_slot - STATE_RETENTION_SLOTS;
+        if latest_finalized_slot > STATE_RETENTION_SLOTS {
+            let prune_target_slot = latest_finalized_slot - STATE_RETENTION_SLOTS;
```
```suggestion
        if latest_finalized_slot > STATE_RETENTION_SLOTS {
            let prune_target_slot = latest_finalized_slot - 1;
```

KolbyML: we only need to keep 1 finalized state

KolbyML (approved):
:shipit: looks good

PR #1517, opened 2026-07-10
fix: common checkpoint selection

Dsorken:
### What was wrong?

In cases where more than two peers do not have identical checkpoints and there is then no common checkpoint to select Ream will stall waiting for a resolution that is not guaranteed to happen. Meaning there is a chance that ream has a chance to get stuck in a still functioning network.

### How was it fixed?

The fix is to add a selection system to break ties in this case. One common way to do so is cluster selection where Ream will look at all checkpoints available and select the highest checkpoint that is not an outlier, where an outlier would typically only pop up in the event a peer re-orgs and continues on the re-org path without the majority of peers being in agreement. This way Ream still chooses the fastest peer without choosing an incorrect peer.

### To-Do

 
### DIFF
--- a/crates/networking/network_state/lean/src/lib.rs
+++ b/crates/networking/network_state/lean/src/lib.rs
@@ -16,6 +16,9 @@ use ream_peer::{ConnectionState, Direction};
 
 use crate::cached_peer::CachedPeer;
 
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit: looks good

PR #1521, opened 2026-07-16
feat: run ef tests from root dir

KolbyML:
### What was wrong?

it is annoying to run ef tests 
### How was it fixed?

be able to run them in root folder


Kayden-ML (approved):
Looks good 👍

PR #1529, opened 2026-07-20
feat(beacon): add beacon e2e block production convergence test

### LINKED ISSUE #1481: Add block production and checkpoint convergence test

perfogic:
### What was wrong?

Closes https://github.com/ReamLabs/ream/issues/1481

### How was it fixed?

- Test `test_beacon_nodes_produce_blocks_and_converge` is added to test with 2 beacon nodes, 2 validators node with 8 validators and finalized successfully at justifed_epoch=3 and finalized_epoch = 2.
**Note**: This one contains lot of changes which i didn't want to but had to. So please take time to check the link below:
- Fix all issues which are documented in https://hackmd.io/@perfogic/r1w-3oR7ze

### Result:
- Run with 2 beacon nodes and 2 validator nodes with 8 validators, every nodes is finalized:
image
image

- For running 4 beacon, validator nodes with 16 validators, i can not make it work in the PR right now, since half of peers will be fell behind the others. => This can be a fix for next PR. The result produced on local:
image

### To-Do

 
### DIFF
--- a/crates/common/validator/beacon/src/beacon_api_client/mod.rs
+++ b/crates/common/validator/beacon/src/beacon_api_client/mod.rs
@@ -46,6 +46,21 @@ use tracing::{error, info};
 
 use crate::aggregate_and_proof::SignedAggregateAndProof;
 
+const BROADCAST_VALIDATION_QUERY_PARAM: &str = "broadcast_validation";
+const BROADCAST_VALIDATION_GOSSIP: &str = "gossip";
+const BROADCAST_VALIDATION_CONSENSUS: &str = "consensus";
+const BROADCAST_VALIDATION_CONSENSUS_AND_EQUIVOCATION: &str = "consensus_and_equivocation";
+
+fn broadcast_validation_value(broadcast_validation: BroadcastValidation) -> &'static str {
+    match broadcast_validation {
+        BroadcastValidation::Gossip => BROADCAST_VALIDATION_GOSSIP,
+        BroadcastValidation::Consensus => BROADCAST_VALIDATION_CONSENSUS,
+        BroadcastValidation::ConsensusAndEquivocation => {
+            BROADCAST_VALIDATION_CONSENSUS_AND_EQUIVOCATION
+        }
+    }
+}
+
 #[derive(Clone)]
 pub struct BeaconApiClient {
     http_client: ClientWithBaseUrl,
@@ -645,17 +660,13 @@ impl BeaconApiClient {
         broadcast_validation: BroadcastValidation,
         signed_beacon_block: SignedBeaconBlock,
     ) -> anyhow::Result<(), ValidatorError> {
-        let broadcast_validation = match broadcast_validation {
-            BroadcastValidation::Gossip => "gossip",
-            BroadcastValidation::Consensus => "consensus",
-            BroadcastValidation::ConsensusAndEquivocation => "consensus_and_equivocation",
-        };
+        let broadcast_validation = broadcast_validation_value(broadcast_validation);
         let response = self
             .http_client
             .execute(
                 self.http_client
                     .post("/eth/v2/beacon/blocks".to_string(), ContentType::Ssz)?
-                    .query(&[("broadcast_validation", broadcast_validation)])
+                    .query(&[(BROADCAST_VALIDATION_QUERY_PARAM, broadcast_validation)])
                     .header(ETH_CONSENSUS_VERSION_HEADER, VERSION)
                     .body(signed_beacon_block.as_ssz_bytes())
                     .build()?,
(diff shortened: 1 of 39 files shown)


crates/common/validator/beacon/src/beacon_api_client/mod.rs
```diff
@@ -649,15 +645,17 @@ impl BeaconApiClient {
...
         let response = self
             .http_client
             .execute(
                 self.http_client
                     .post("/eth/v2/beacon/blocks".to_string(), ContentType::Ssz)?
-                    .query(&[(
-                        "broadcast_validation",
-                        serde_json::to_string(&broadcast_validation)?,
-                    )])
+                    .query(&[("broadcast_validation", broadcast_validation)])
```
KolbyML: these strings should be consts, and reused where it makes sense, `"/eth/v2/beacon/blocks"` is fine to inline, but the rest should be const's

crates/common/validator/beacon/src/block.rs
```diff
@@ -24,17 +24,16 @@ pub fn get_block_signature(
 }
 
 pub fn sign_beacon_block(
-    slot: u64,
+    _slot: u64,
```
KolbyML: if we aren't using slot, it shouldn't be an arguement, apply this comment to other applicable places

crates/networking/manager/src/gossipsub/handle.rs
```diff
@@ -118,13 +127,62 @@ pub fn init_gossipsub_config_with_topics() -> GossipsubConfig {
...
+async fn import_gossip_attestation(
+    beacon_chain: &BeaconChain,
+    single_attestation: &SingleAttestation,
+) -> anyhow::Result<()> {
+    let (attestation, should_process_attestation) = {
+        let store = beacon_chain.store.lock().await;
+        let head_root = store.get_head()?;
+        let mut state =
+            store.db.state_provider().get(head_root)?.ok_or_else(|| {
+                anyhow::anyhow!("No beacon state found for head root: {head_root}")
```
KolbyML: import anyhow

testing/mock-execution-engine/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
+
+[dependencies]
+actix-web.workspace = true
+alloy-primitives.workspace = true
+anyhow.workspace = true
+ream-execution-rpc-types.workspace = true
+serde.workspace = true
```
KolbyML, line 17: ream dependencies shouldn't be included in the general depends, put it below with `# ream depends` or something, etc for dev-dependencies

KolbyML (commented):
Some feedback

crates/common/validator/beacon/src/beacon_api_client/mod.rs
perfogic: Thank you, i'll address them

crates/networking/manager/src/gossipsub/handle.rs
perfogic: addressed

testing/mock-execution-engine/Cargo.toml
perfogic: addressed

crates/common/validator/beacon/src/block.rs
perfogic: addressed

crates/common/validator/beacon/src/beacon_api_client/mod.rs
perfogic: addressed

perfogic:
@KolbyML I have addressed all the above comments, it's ready for your next review

KolbyML (approved):

PR #1531, opened 2026-07-22
fix(gossip): align data column sidecar validation with spec

perfogic:
### What was wrong?

`validate_data_column_sidecar_full` derived the expected proposer from the current head state instead of the sidecar block header's parent state, which is different from the consensus specification - [`validate_data_column_sidecar_gossip`](https://github.com/ethereum/consensus-specs/blob/master/specs/fulu/p2p-interface.md?plain=1#L357-L444)

It has been found in Hans' PR in our work: https://github.com/ream-collective/ream/pull/2#pullrequestreview-4744572781

### How was it fixed?

- Load the parent state from `block_header.parent_root`, advance it to `block_header.slot`, and use that state to validate the expected proposer. 
- The validation flow was also aligned with the spec for slot timing, proposer signatures, KZG proofs, and duplicate handling.

### To-Do

### DIFF
--- a/crates/networking/manager/src/gossipsub/handle.rs
+++ b/crates/networking/manager/src/gossipsub/handle.rs
@@ -1,3 +1,5 @@
+use std::time::{SystemTime, UNIX_EPOCH};
+
 use libp2p::gossipsub::Message;
 use ream_chain_beacon::beacon_chain::BeaconChain;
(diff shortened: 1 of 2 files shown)


KolbyML (approved):
:shipit: looks good

PR #1534, opened 2026-07-23
fix: bound memory growth during long non-finality

shariqnaiyer:
### What was wrong?

 

During the devnet5 run, finality stalled network-wide at slot ~8874 while heads advanced past 130k, and all ream nodes OOM-crash-looped: fork choice decoded a full SignedBlock per parent-chain hop back to the stalled justified root on every tick, and all attestation-table pruning was gated on the finalized checkpoint via a prune function that was never called.

### How was it fixed?

 - compute_lmd_ghost_head and update_head's finalized-root walk now use the compact children-index map (root → (parent_root, slot)) instead of per-hop block decodes and an extra children-table scan
 
- prune_stale_attestation_data prunes by a head-anchored window (ATTESTATION_RETENTION_SLOTS = 1024) in addition to finality, never touching data roots that still back a validator's latest vote

 

### To-Do

 

zemse (approved):

Kayden-ML (approved):
Looks good nice find 👍

PR #1543, opened 2026-07-30
fix: support checkpoint sync and P2P handshake on post-Fulu networks

### LINKED ISSUE #1538: Support syncing live post-Fulu networks (Sepolia/Mainnet)
### Why we have this
In 25/06/2026, when reviewing the PR https://github.com/ream-collective/ream/pull/6, i have found out that currently with simple syncing on beacon client with Sepolia, we faced some issues as below:
- **Checkpoint sync panics fetching blob sidecars**: currently Sepolia supports Fulu, so fetching blob sidecars using old API will fail, we need a guardcheck to make sure not syncing blob sidecars after `FULU_FORK_EPOCH`
- **Fork digest ignores EIP-7892, so no peer completes the handshake**: Peers connect and immediately drop us because our advertised fork digest matches
no known chain.
- **Attestations from gossip are all rejected**: Continuous, roughly one per second, for every attestation received.
- **DA verfied fail**: this one should be addressed after decouple DA is done, we can go with disabled da check case.
- maybe more

### Goal
`ream beacon_node --network sepolia` follows the live chain, more issues can be found during the work, so it can be listed later on the PR fixed. This can be closed once two checkboxes are done:
- [ ] sync without da: should focus on this first
- [ ] sync with da

vuonghuuhung:
### What was wrong?

Fixes https://github.com/ReamLabs/ream/issues/1538

Sepolia has passed Fulu and gone through multiple BPO (EIP-7892) blob-schedule bumps, which broke sync in three ways: checkpoint sync panics fetching blobs via the legacy `blob_sidecars` API (no longer served post-Fulu), fork digest is computed wrong (ignores per-network blob schedule, wrong hash algorithm, hardcoded mainnet epoch) so no peer ever passes the discv5/Status handshake, and forward sync bans peers that correctly return empty legacy blob responses, eventually starving itself of peers.

### How was it fixed?

- Checkpoint sync and the range-syncer now skip the legacy blob API for post-Fulu blocks.
- Fork digest now reads the network's real `blob_schedule` and uses the correct EIP-7892 hash (raw SHA256, not SSZ `hash_tree_root`).
- Added `BeaconNetworkSpec::current_epoch()` and replaced all hardcoded `FULU_FORK_EPOCH` fork-digest call sites with it.
- Verified live on Sepolia: checkpoint sync completes, peers connect and pass Status handshake, gossip flows, forward sync no longer bans peers.

Before changes, checkpoint sync panics fetching blobs for a post-Fulu block:
image

After changes, checkpoint sync completes and peers successfully connect and pass the Status handshake:
image
image

### What's to notice

Found but out of scope (tracked separately): the range-syncer only targets finalized slot (not head) with no backfill mechanism, so the chain gets stuck at the checkpoint and gossip attestations get rejected as "not a member of the committee" (stale RANDAO). Also, `req_resp` context bytes and a couple RPC handlers use a global "current epoch" instead of the epoch of the specific item/state being served wrong for historical (pre-Fulu) queries.

### DIFF
--- a/crates/common/chain/beacon/src/beacon_chain.rs
+++ b/crates/common/chain/beacon/src/beacon_chain.rs
@@ -5,7 +5,7 @@ use ream_consensus_beacon::{
     attestation::Attestation, attester_slashing::AttesterSlashing,
     electra::beacon_block::SignedBeaconBlock,
 };
-use ream_consensus_misc::constants::beacon::{FULU_FORK_EPOCH, genesis_validators_root};
+use ream_consensus_misc::constants::beacon::genesis_validators_root;
 use ream_events_beacon::{BeaconEvent, BeaconEventSender, event::chain::BlockEvent};
 use ream_execution_engine::ExecutionEngine;
 use ream_fork_choice_beacon::{
@@ -128,8 +128,10 @@ impl BeaconChain {
         };
 
         Ok(Status {
-            fork_digest: beacon_network_spec()
-                .fork_digest(FULU_FORK_EPOCH, genesis_validators_root()),
+            fork_digest: beacon_network_spec().fork_digest(
+                beacon_network_spec().current_epoch(),
+                genesis_validators_root(),
+            ),
             finalized_root: finalized_checkpoint.root,
             finalized_epoch: finalized_checkpoint.epoch,
             head_root,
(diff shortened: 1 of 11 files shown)


KolbyML (approved):
:shipit: looks good

PR #1544, opened 2026-07-30
feat: broadcast data column sidecars

### LINKED ISSUE #1482: Broadcast data column sidecars from proposer
## Description

Implement the following runtime flow.

1. Proposer data column broadcast
- After producing a full block with blobs/KZG proofs, compute the cells and KZG proofs.
- Construct `DataColumnSidecar`s from the proposed block.
- Broadcast the data column sidecars on the correct `data_column_sidecar_{subnet_id}` gossip topics.

2. Incoming data column handling
- Add or wire an `on_data_column_sidecar`-style path for received data column sidecars.
- Verify received data column sidecars.
- Store valid sidecars in beacon storage.
- Ignore/reject invalid sidecars without invalidating an otherwise valid block.
- Re-gossip accepted sidecars where appropriate.

3. Availability handling
- Keep running `is_data_available` for block availability checks.
- If data is not available yet, then just log it, since it will be handle in #1484 

4. Add e2e test: `test_beacon_nodes_propagate_data_column_sidecars`
- start at least two beacon nodes
- produce a block with blobs
- assert data column sidecars are broadcast by the proposer
- assert the peer receives, verifies, and stores them

This issue can be considered as done if flow is fully implemented and test is passed

vuonghuuhung:
### What was wrong?

Fixes #1482

Phase 1 of PeerDAS/Fulu compatibility (#1478) requires the proposer to build and broadcast `DataColumnSidecar`s after producing a block with blobs. Ream could already validate, store, and re-gossip incoming `DataColumnSidecar`s, but nothing ever produced or broadcast one, the DAS pipeline never actually ran end-to-end.

### How was it fixed?

- Cache each blob at production time (`get_blocks_v3`), keyed by block root, since the validator client currently posts back only the bare signed block after signing.
- After publishing a block, recompute cells/KZG proofs from the cached blob, assemble all 128 `DataColumnSidecar`s, store them locally, and gossip them on their subnet topics (`broadcast_data_column_sidecars`, run via `spawn_blocking` so KZG work doesn't stall the async executor).
- Share a single `DASContext` (`matrix_entry::das_context`) instead of rebuilding the trusted setup per call.
- Make `on_block` log and continue instead of rejecting a block when its data column sidecars aren't available yet, per the issue's Phase 1 scope. `is_data_available` now returns `Ok(false)` for "no sidecars yet" instead of erroring, so genuine storage/KZG errors still surface and still block import.
- Fix the `DataColumnSidecar` gossip handler to report the real `MessageAcceptance` instead of a hardcoded `Ignore`, so accepted and rejected sidecars actually propagate through the mesh; dropped the now-redundant manual re-publish. Added a regression test.
- Generate real blobs, KZG commitments, and a minimal EIP-4844 transaction in the mock execution engine so e2e tests exercise a genuine blob → cell/proof → gossip path.
- Added a beacon e2e test covering proposer broadcast and peer receive/validate/store of data column sidecars.

### What's to notice

Deliberately out of scope for this PR, tracked separately:
- Full-custody availability (`is_data_available` returns available as soon as one valid column is found, not once all 128 are present) — belongs with #1483's pending-block flow.
- Fulu/V5 proof cardinality (`get_blocks_v3` assumes one proof per commitment, which doesn't hold for a real `BlobsBundleV2`): pre-existing, not exercised by the current Electra-fixture e2e suite.
- The block-root-keyed blob cache is a compatibility bridge for the validator client currently discarding blobs after signing, not the end state , a `SignedBlockContents`-style publish payload would remove the need for it.
- Blocks proposed via an external builder never reach this path; from Fulu onward the builder is expected to publish directly.

### To-Do


testing/mock-execution-engine/src/block_generator.rs
```diff
@@ -1,15 +1,64 @@
...
+    transaction::{AccessList, BlobTransaction, ToAddress},
 };
-use ssz_types::VariableList;
+use rust_eth_kzg::{DASContext, TrustedSetup, UsePrecomp};
+use ssz_types::{FixedVector, VariableList};
+
+const BYTES_PER_BLOB: usize = 131_072;
+const BLOB_TRANSACTION_TYPE: u8 = 3;
+
+fn encode_blob_transaction(blob_versioned_hashes: Vec<B256>) -> Vec<u8> {
```
KolbyML, line 24: I think you should just implement default on BlobTransaction

KolbyML (commented):
also please rebase

testing/mock-execution-engine/src/block_generator.rs
vuonghuuhung: Thanks, I addressed it

perfogic (commented):
@vuonghuuhung the test is failed, can you give it a check ?

vuonghuuhung:
> @vuonghuuhung the test is failed, can you give it a check ?

All good, the test failed due to cicd resource

perfogic:
> > @vuonghuuhung the test is failed, can you give it a check ?
> 
> All good, the test failed due to cicd resource

https://github.com/ream-collective/ream/actions/runs/33866736551/job/101003994679?pr=22
really ? since it passed on our repository `ream-collective` and we never faced this issue before.

vuonghuuhung:
> > > @vuonghuuhung the test is failed, can you give it a check ?
> > 
> > 
> > All good, the test failed due to cicd resource
> 
> https://github.com/ream-collective/ream/actions/runs/33866736551/job/101003994679?pr=22 really ? since it passed on our repository `ream-collective` and we never faced this issue before.

I'm re-checking it, cause they're passed at that time but after several times tested, there is unstable failure

KolbyML (approved):
:shipit: looks good

PR #1554, opened 2026-08-24
feat: Implement new metrics for gossip arrival timing

Kayden-ML:
### How was it fixed?

Implements metrics for gossip arrival timing

https://github.com/leanEthereum/leanMetrics/pull/40


KolbyML (approved):
:shipit: looks good

PR #1555, opened 2026-08-25
feat: publish devnet5 LeanVM image variant

KolbyML:
### What was wrong?

The devnet's Ethlambda are running are currently using leanvm 0.9, we still want to support the last version for hive

### How was it fixed?

use feature flags


Kayden-ML (approved):
Looks good 👍 nice work

PR #1559, opened 2026-08-25
fix: sort workspace Cargo.toml files

shariqnaiyer:
`cargo sort --grouped --check --workspace` is currently failing on `master`, which fails the `cargo-sort` CI job. This restores it to passing.


Kayden-ML (approved):
Looks good 👍

PR #1562, opened 2026-08-26
ci: remove broken Lean client matrix

KolbyML:
### What was wrong?

we don't plan to fix this

### How was it fixed?

remove it


Kayden-ML (approved):
Looks good 👍

PR #1563, opened 2026-08-26
chore: raise minimum Rust version to 1.98

KolbyML:
### What was wrong?

our rust version is 1 year old

### How was it fixed?

bumping it


Kayden-ML (approved):
Looks good 👍

PR #1566, opened 2026-08-27
refactor: avoid redundant hashing and clones in attestation production

Kayden-ML:
### How was it fixed?

Compute the shared attestation root once per slot and move signed attestations into the chain service to eliminate redundant hashing, cloning


KolbyML (approved):
:shipit: looks good

PR #1568, opened 2026-08-29
ci: remove known good vs current pr ci

Kayden-ML:
### How was it fixed?

removes known good vs current pr ci


KolbyML (approved):
:shipit: looks good

PR #1569, opened 2026-08-31
feat(DA): ssz payload and batch process

KimiWu123:
### What was wrong?

1. Addressed the payload being in hex string: https://github.com/ReamLabs/ream/pull/1485#discussion_r3679994972.
2. Supported batch checks instead of checking per column. 

#### Summary
This PR makes a whole block the unit of DA ingestion, which cuts
1. KZG verification from 128 checks down to 1 batched check,
3. RPC requests from 128 down to 1, and
4. wire size roughly in half (hex JSON → raw SSZ bytes).

Admission is all-or-nothing: the batch is either fully accepted or fully shed, never split. A request is rejected (400) when SSZ decoding fails, the batch is empty, or a column index is duplicated or out of range. Verification, however, is per column: once admitted, every valid column is stored individually, so a single broken column proof doesn't break the others — the batched KZG check falls back to per-column checks to isolate the culprit.

#### Measured
End to end (mainnet slot 14722952, 17 blobs):
- one SSZ batch: 1.05 s (submission: 12.90 ms)
- 128 JSON requests: 1.46 s (submission: 31.28 ms)

**~28%** faster end to end. 
The submission phase — request round trips plus request decoding (hex JSON vs raw SSZ) — is ~2.4× faster on its own, but in absolute terms that is ~18 ms of the ~410 ms saved: the bulk of the gain is node-side, one batched KZG pairing check instead of 128.

```bash
# restart the DA node between the two runs 

# this PR: one SSZ batch
./target/release/ream-da-tool feed --beacon-url https://ethereum-beacon-api.publicnode.com 14722952

# old path: one request per column
./target/release/ream-da-tool feed --beacon-url https://ethereum-beacon-api.publicnode.com 14722952 --per-column

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7246,12 +7246,15 @@ version = "0.1.0"
 dependencies = [
  "actix-web",
  "alloy-primitives",
+ "ethereum_ssz",
+ "ethereum_ssz_derive",
  "ream-api-types-common",
  "ream-data-availability",
  "ream-data-availability-node",
  "ream-rpc-common",
  "serde",
  "serde_json",
+ "ssz_types",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6357,6 +6357,7 @@ dependencies = [
  "clap",
  "discv5",
  "ethereum_ssz",
+ "ethereum_ssz_derive",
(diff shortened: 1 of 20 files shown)


KolbyML (approved):
:shipit: looks good

PR #1570, opened 2026-09-01
feat(DA): Column Reconstruction

KimiWu123:
### What was wrong?

1. Add a retention floor. If any data is older than the retention floor, it should be deleted and should not be kept in the store.
2. Implement data column reconstruction.  

#### Summary

When a write moves a block across the recoverable threshold — at least half of the column set held (`NUMBER_OF_COLUMNS / 2 = 64`) — the verification service arms a one-shot reconstruction trigger. The trigger fires after a randomized settling delay, giving naturally-arriving columns a chance to complete the block first and desynchronizing reconstruction across nodes, per spec.

Recovered columns are re-admitted through the normal verify-then-store path (`process_candidate_block`) as candidates rather than trusted directly: the recovery math is sound, but the assembly around it is ordinary code, and re-verification catches its bug classes for free.

**Key piece**: `DaReconstructor` trait in `ream-da`, implemented by the KZG adapter — `KzgVerifier` is renamed `KzgAdapter` since one object now serves both roles. Housing both capabilities in one adapter keeps it the node's single home for the KZG machinery: the trusted setup is loaded and cached once per context, and no second KZG-aware component exists anywhere else in the node.

### Next Step
Reconstruction currently routes through the queue because the store's single-writer contract lives outside the store — the queue serializes all writes. Once the file store is replaced by a DB that owns its own concurrency, reconstruction can run self-contained and skip the queue round-trip.

### DIFF
--- a/crates/common/data_availability_node/src/service.rs
+++ b/crates/common/data_availability_node/src/service.rs
@@ -24,14 +24,10 @@ pub const DEFAULT_RECONSTRUCTION_DELAY: Duration = Duration::from_secs(3);
 pub struct DataAvailabilityVerificationService {
     receiver: mpsc::Receiver<IngestWorkItem>,
     verifier: Arc<dyn ColumnVerifier>,
-    /// Recovers a block's missing columns from the held ones
     reconstructor: Arc<dyn ColumnReconstructor>,
     store: Arc<dyn ColumnWriteStore>,
     executor: ReamExecutor,
-    /// Weak self-handle for the delayed reconstruction triggers.
     self_sender: mpsc::WeakSender<IngestWorkItem>,
-    /// Cap on how long a recoverable-but-incomplete block is given to complete
-    /// naturally before recovery is attempted
     max_reconstruction_delay: Duration,
 }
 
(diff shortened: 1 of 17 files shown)


crates/common/data_availability_node/src/service.rs
```diff
@@ -46,11 +64,136 @@ impl DataAvailabilityVerificationService {
...
+            tokio::time::sleep(delay).await;
+            // All real handles gone: shutting down, drop the trigger.
+            let Some(sender) = sender.upgrade() else {
+                return;
+            };
+            let _ = sender
+                .send(IngestWorkItem::Reconstruction(ReconstructionRequest {
+                    block_root,
+                }))
+                .await;
```
KolbyML, line 193: can we warn this

KolbyML (approved):
please fix this issue in your next PR

crates/common/data_availability_node/src/service.rs
KimiWu123: addressed in [958133c](https://github.com/ReamLabs/ream/pull/1572/commits/958133c115d5acd11eef3bb8a6ca2d95929e5869) in #1572

